PR #47, opened 2014-07-21
Battery Levels added to Monitor and Pebble

m2oore:
Added some code to be able to view Battery Levels of the CGM Uploader to the CGM Monitor and Pebble Watch.

Battery Levels are uploaded as a new element in the Mongo DB.


jasoncalabrese:
this is really old, still don't have battery on the webmonitor, but would probably be best to do that from scratch since there have been lots of updates since this

PR #60, opened 2014-07-29
Alarms

toarikaplan:
image
added authentication, user defined alarms, clock changes to red when (user defined) data is latent, resized window to fit on 1 page for iPad/iPhone


jasoncalabrese:
most of this should be covered by #305

PR #108, opened 2014-08-21
abs monitor

andrewbs:


jasoncalabrese:
this is really out of date and I'm not sure if it's needed anymore

PR #202, opened 2014-10-16
For custom High/low into WIP

raven569:


jasoncalabrese:
most of this should be covered by #305

PR #190, opened 2014-10-01
Wip/mqtt/init - tracking mqtt experiments

bewest:
The corresponding uploader changes (there are a few actually)

The protobuf model:
https://github.com/nightscout/android-uploader/blob/master/core/src/main/java/com/nightscout/core/model/Download.proto

Feel free to close/ignore if this gets in the way of things, I'll move it to my personal fork if it becomes a distraction.

PR #219, opened 2014-11-02
introduce bootevent

bewest:
Also, make modules export functions.

This also makes the app re-usable as an npm express module.


bewest:
We got all this in #190

PR #232, opened 2014-11-08
Have vagrant provisioning run setup.sh, not user

mzemel:
Most useless PR ever ;)


jasoncalabrese:
this is no longer a clean merge and should be targeting to dev

PR #258, opened 2014-11-14
Added Deploy to Azure Button

apwestgarth:
Added Deploy to Azure Button to Readme.md as per instructions here - http://azure.microsoft.com/blog/2014/11/13/deploy-to-azure-button-for-azure-websites-2/

This enables a user to easily deploy the Nightscout Web Application to Azure Websites.


jasoncalabrese:
I think it was decided that we didn't want to send users down this path since upgrades are much harder

Issue #271, opened 2014-11-22
Support basic custom alarm thresholds set via env variable


scottleibrand:

jasoncalabrese:
@gregwaehner lets open a new issue for pushover based alarms, we can work on putting that together for E

PR #305, opened 2014-12-19
Alarm / Threshold updates (and some other UI updates...)

jasoncalabrese:
Many users don't understand the current Loss/PredictAR based alarms, they also want to be able to set their own target range instead of the default 80-180.

This PR aims to make the target range and alarm types configurable, this should address most of #271.  To do this 5 new ENV variables have been added.

   * `BG_HIGH` - default: **260**, the high BG outside the target range that is considered urgent
   * `BG_TARGET_TOP` - default: **180**, the top of the target range, also used to draw the line on the chart
   * `BG_TARGET_BOTTOM` - default: **80**, the bottom of the target range, also used to draw the line on the chart
   * `BG_LOW` - default: **55**, the low BG outside the target range that is considered urgent
   * `ALARM_TYPES` - default to **simple** if any `BG_`* ENV's are set, otherwise default to **predict**.  There currently 2 alarm types are supported, and can be used independently are separately.  The `simple` alarm type only compares the current BG to `BG_` thresholds above, the `predict` alarm type works as before by using a highly tuned formula that forecasts where the BG is going based on it's trend.

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -151,7 +151,7 @@ body {
 }
 
 #bgButton {
-    background: #ff2035;
+    background: yellow;
     border: 2px solid #DDD;
     border-right: 2px solid #ccc;
     border-bottom: 2px solid #ccc;
@@ -174,6 +174,10 @@ body {
     -khtml-user-select: none;
     -moz-user-select: none;
     -ms-user-select: none;
+}
+
+#bgButton.urgent {
+  background: red;
 }
 
 .bgButton:active {
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -242,6 +242,11 @@ div.tooltip {
     font-size: 70%;
   }
 
+  html body #bgButton {
+    font-size: 80%;
+  }
+
+
   html body #chartContainer {
     top: 130px;
   }
@@ -289,13 +294,6 @@ div.tooltip {
   }
   .dropdown-menu {
     font-size: 60% !important;
-  }
-  #currentTime {
-    font-size: 65%;
-    padding: 0;
-  }
-  #lastEntry {
-    font-size: 65%;
   }
 
   #chartContainer {
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -224,37 +224,6 @@ div.tooltip {
 	padding: 0 0 10px 0;
 }
 
-@media (max-height: 480px) {
-  html body #toolbar {
-    float: right;
-    height: auto;
-  }
-
-  #toolbar h1 {
-    display: none;
-  }
-
-  html body .container {
-    top: 15px;
-    padding-bottom: 20px;
-  }
-  html body .container .status {
-    font-size: 70%;
-  }
-
-  html body #bgButton {
-    font-size: 80%;
-  }
-
-
-  html body #chartContainer {
-    top: 130px;
-  }
-  html body #chartContainer svg {
-    height: calc(100vh - 130px);
-  }
-}
-
 /* Large desktop */
 @media (min-width: 980px) {
     .content {
@@ -270,7 +239,7 @@ div.tooltip {
 
 @media (max-width: 800px) {
   #chartContainer {
-    font-size: 115%;
+    font-size: 80%;
   }
 
   .bgStatus {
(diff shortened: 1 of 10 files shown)


static/css/main.css
```diff
@@ -238,6 +242,11 @@ div.tooltip {
     font-size: 70%;
   }
 
+  html body #bgButton {
```
trhodeos: `html body` is unnecessary.

trhodeos: Also, should this be merged with the `#bgButton` rule on line 153?

static/js/client.js
```diff
@@ -436,6 +405,17 @@
         // selects all our data into data and uses date function to get current max date
         var focusCircles = focus.selectAll('circle').data(focusData, dateFn);
 
+        var focusDotRadius = function(d) {
+            if (d.type == 'mbg')
+                return 6;
+            else if ($(window).width() > WIDTH_BIG_DOTS)
```
trhodeos: cache $(window) into a local var.

static/js/client.js
```diff
@@ -452,8 +433,30 @@
...
+                tooltip.html('<strong>' + bgType + ' BG:</strong> ' + d.sgv +
+                    (d.type == 'mbg' ? '<br/><strong>Device: </strong>' + d.device : '') +
+                    '<br/><strong>Time:</strong> ' + formatTime(d.date))
+                    .style("left", (d3.event.pageX) + "px")
+                    .style("top", (d3.event.pageY - 28) + "px");
+            })
+            .on('mouseout', function (d) {
+                if (d.type != "sgv" && d.type != 'mbg') return;
+                tooltip.transition()
+                    .duration(500)
```
trhodeos: any reason why the mouseout and mouseover durations are different?

static/js/client.js
```diff
@@ -462,11 +465,12 @@
         d3.selectAll('.path').remove();
 
         // add treatment bubbles
-        //
-        //var bubbleSize = prevChartWidth < 400 ? 4 : (prevChartWidth < 600 ? 3 : 2);
-        //focus.selectAll('circle')
-        //    .data(treatments)
-        //    .each(function (d) { drawTreatment(d, bubbleSize, true) });
+
+        var bubbleSize = prevChartWidth < 400 ? 4 : (prevChartWidth < 600 ? 3 : 2);
```
trhodeos: can this use the `focusDotRadius` function (or a helper function that both use)? This logic seems similar.

static/js/client.js
```diff
@@ -505,39 +516,33 @@
...
-                  .attr('x', function (d) { return xScale(d.created_at); })
-                  .attr('y', function (d) { return yScale(scaleBg(500)); })
-                  .attr("width", 15)
-                  .attr("height", 15)
-                  .attr("rx", 6)
-                  .attr("ry", 6)
+            treatCircles.enter().append('circle')
+                  .attr('cx', function (d) { return xScale(d.created_at); })
+                  .attr('cy', function (d) { return yScale(scaleBg(d.glucose || calcBGByTime(d.created_at.getTime()))); })
+                  .attr("r", 6)
```
trhodeos: Do we want this to be 6 always, regardless of screen size?

static/js/client.js
```diff
@@ -545,11 +550,11 @@
...
                   })
-          .on("mouseout", function (d) {
-              div.transition()
-                  .duration(500)
-                  .style("opacity", 0);
-          });
+                  .on('mouseout', function (d) {
+                      tooltip.transition()
+                          .duration(500)
+                          .style("opacity", 0);
```
trhodeos: Might as well change these to be single quotes as well.

static/js/client.js
```diff
@@ -1133,33 +1004,49 @@
 
     }
 
-    ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+    function calcBGByTime(time) {
+        var closeBGs = data.filter(function(d) {
+            if (!d.y) return false;
```
trhodeos: surround with `{}`s

bewest:
Looking good.

trhodeos:
Added a few comments. Looks good otherwise.

jasoncalabrese:
@trhodeos thanks for the code review, think I took care of those issues

I merged this to my prod branch and will be using it all day tomorrow, if there aren't any new issues I plan to merge tomorrow.

PR #315, opened 2014-12-28
Process multiple profile objects with start and end timestamps and pick ...

sulkaharo:
...the last that matches the time. Also preserves old functionality of using last non-timestamped profile object, if those are present.

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -161,11 +161,26 @@ function update() {
           });
 
           profiles.list(function (err, results) {
-              // There should be only one document in the profile collection with a DIA.  If there are multiple, use the last one.
+              
+              // There should be only one document in the profile collection with a DIA.
+              // If there are multiple, use the one with latest start date.
+              // If no profile is marked with a date, use the last profile found.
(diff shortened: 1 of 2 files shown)


sulkaharo:
The last two commits are about improving the performance of this - looks like the amount of string parsing introduced by the change was a bit much for mobile browsers. The performance improvement impact of the changes is ~100ms on my laptop and took away the newly introduced lag when using Nightscout on the phone.

jasoncalabrese:
looking good and I think this will help me, I plan to test tomorrow

there's a /iob-cob.html page that is probably broken by this.  It basily a debug page, but I'm sure it won't handle the a value containers.  Would be nice to share some code with it on `websocket.js`

sulkaharo:
Yup, the entirety of DIYPS / profile data code needs to be pulled out from client.js in a manner that's usable across the codebase. I'll check what the job-cob.html looks like today.

scottleibrand:
iob-cob.html is already broken in other ways too, and needs a rewrite to
use the same code as client.js.

PR #316, opened 2014-12-30
WIP: attempt clock in mmol

bewest:
For @ELUTE and others.

### DIFF
--- a/static/clock.html
+++ b/static/clock.html
@@ -40,6 +40,7 @@
     <h1 class="bgnow"></h1>
   </div>
   <script src="/bower_components/jquery/dist/jquery.min.js"></script>
+  <script src="/api/v1/status.js"></script>
   <script type="text/javascript">
     function query ( ) {
       console.log('query');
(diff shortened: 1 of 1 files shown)


static/clock.html
```diff
@@ -51,6 +52,9 @@ <h1 class="bgnow"></h1>
       var rec = xhr[0];
       var last = new Date(rec.date);
       var now = new Date( );
+      if (window.serverSettings.units == 'mmol') {
+        rec.sgv = (Math.round((rec.sgv / 18) * 10) / 10).toFixed(1);
```
StephenBlackWasAlreadyTaken: the fact that you need to move decimals around to round in js is mindblowing. I had to look it up because I didnt believe it. Since mmol conversions just like this probably happen frequently, might be nice to just have a class/function/whatever it is in js to simply take an sgv and return its mmol value, but this looks like you were just going for a quick fix so seems fine!

bewest: Agreed.... this phrase is now copy-pasted 4 or 5 times throughout our code base... this is a recipe for badness.

bewest:
Merging this to go with @jasoncalabrese other changes soon.  For @ELUTE and other mmolers.

PR #329, opened 2015-01-09
auto ack alarms when going back into range or when the a cgm error corrects

jasoncalabrese:
This should remove some of the annoyance of alarms, no more returning home to alarms for a condition that was already corrected (but not ack'd)

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -342,7 +342,6 @@ function loadData() {
   configure( );
   listeners( );
   setInterval(kickstart(update), ONE_MINUTE);
-  setInterval(kickstart(emitData), ONE_MINUTE);
 
   return io;
 }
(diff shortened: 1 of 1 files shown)


jasoncalabrese:
I'd like to get this into the release since I think it's a big source of alarm annoyment, would like a review by @bewest and @rnpenguin

bewest:
Looks like some useful clean up here! Thanks.

Is there a new feature that should be `ENABLE`'d here or is this a bug fix?
(This is to help establish design controls, code review, etc...).

jasoncalabrese:
I see this as a bug fix.  Without this fix alarms will continue until ack'd even when the condition changes.

For example when a user is away from home they use the notifications on the dexcom receiver or the pebble to make corrections.  In that case alarms from the `cgm-remote-monitor` are not ack'd or even seen.  When they return home any devices running `cgm-remote-monitor` will still be alarming for a condition that has long since been corrected.  This leads to distrust of the system and false panic.

bewest:
:+1: please :shipit:

PR #331, opened 2015-01-11
update readme to include all the env var's that are used to configure the server

jasoncalabrese:


jasoncalabrese:
good catch @jjmason

bewest:
Here is preview link: https://github.com/nightscout/cgm-remote-monitor/blob/wip/readme-env-update/README.md

I added some links to the update-fork, mongostring, and configure tools.
I also created two tables with the things people might be interested in up front and center, clearly marked the one REQUIRED thing, and the rest of the optional things.

jasoncalabrese:
nice, some of the env's are `code` other just CAPS
Core should probably be first too

jasoncalabrese:
I'm going to update now

bewest:
Looks ok, I think the labs/features table should be first, based on how often it's likely to be used.
The only one required environment variable is covered in detail in other places, so I'm guessing people are going to want the labs/features table most often.

jasoncalabrese:
My thinking is that it would help someone get started it they saw the required/core vars first, but you have a good point too.  Either way works for me.

bewest:
Looks like the `code fences` make the thing do awful things with the scrollbars.

PR #332, opened 2015-01-12
Fix buttons

jjmason:
This PR makes a small layout change, moving the IOB/COB display to appear below the "currentDeteails" element.  This makes the buttons more easily clickable, and allows one to see the IOB/COB display on mobile devices when a custom title is in use.


scottleibrand:
LGTM, but @jasoncalabrese, wasn't there another (partial) fix for this same issue?  Just want to make sure this doesn't conflict with that before merging...

jasoncalabrese:
all of this will change as we merge dev back to iob-cob, or just port the rest of iob to the mainline

PR #334, opened 2015-01-12
adjust breakpoint to better support display on iphone6+ in portrait mode

jasoncalabrese:


bewest:
LGTM. :shipit:

jasoncalabrese:
@jimsiff found that the alarms button still doesn't display correctly

jimsiff:
I tweaked the #bgButton font size down to 80% on my test branch and can now see most of the delta and trend but it's still clipped.  The time and time ago is still missing and the button is indented from the right which might be part of the issue with the missing time.  I'm not great with CSS but I will keep trying.  Don't let this hold up 0.6.0.

jasoncalabrese:
There's also that old hack that hides the time ago when the button is displayed, that probably not kicking in.  This that happens at 500px, need to find a real fix for that.

jasoncalabrese:
I just for a nexus 6, and it had the same display issue as the iPhone 6 plus in portrait mode, will be easy for me to test and fix now

jimsiff:
Looking forward to testing what you come up with.  Thanks for looking at this again @jasoncalabrese.

jasoncalabrese:
just pushed, should be better now.  Tested for several devices using chrome's device mode

screen shot 2015-02-06 at 10 14 47 pm

jimsiff:
Looks good, thanks Jason.

screen shot 2015-02-07 at 12 01 09 am

jasoncalabrese:
Thanks for testing @jimsiff, think I'll merge this to Dev so we can get some more eyes on it before release

PR #345, opened 2015-01-15
Dreamsicle

amurray500:


jasoncalabrese:
This pull request is going in the wrong direction, looks like you were able to do the upgrade separately

amurray500:
Yes I had made some changes to formatting and following the instructions didn't work because it viewed those changes as in conflict. If this had been subversion I could have done this with my eyes closed but I am a git newbie and at this point anyway it does not seem very intuitive. Of course I know it  is just a learning curve. Thanks for checking on me.

PR #350, opened 2015-01-16
Time update fix will resolve #348

### LINKED ISSUE #348: Browsers not staying current with Dreamsicle
Several users are reporting that their websites are not keeping current with their computer's clock and are static until the next update or 2.  Some show "1 min ago" for 5 minutes, then switch to "5 mins ago" for 5 more minutes.  Back to "1 min ago".  They are not all 1 & 5 (mins ago).  Some are 4 & 12.
Seen in Chrome & Safari on Mac OSX, and on Windows with IE (v11).

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7794864-browsers-not-staying-current-with-dreamsicle?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jasoncalabrese:
With these changes the clock update is done client-side only, on exactly each minute the clock will update and the time ago will be recalculated.  Also a tooltip was add to the time ago element that will show the last update and data times.

screen shot 2015-01-15 at 11 33 14 pm

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -45,7 +45,6 @@ var dir2Char = {
   function emitData ( ) {
     console.log('running emitData', now, patientData && patientData.length);
     if (patientData.length > 0) {
-      io.sockets.emit("now", now);
       io.sockets.emit("sgv", patientData);
     }
   }
@@ -62,7 +61,6 @@ var dir2Char = {
 
   function listeners ( ) {
     io.sockets.on('connection', function (socket) {
-        io.sockets.emit("now", now);
         io.sockets.emit("sgv", patientData);
         io.sockets.emit("clients", ++watchers);
         socket.on('ack', function(alarmType, silenceTime) {
@@ -312,6 +310,7 @@ function loadData() {
         }
     }
 }
+
   function is_different (actual, predicted, mbg, treatment, errorCode) {
     if (patientData && patientData.length < 3) {
       return true;
@@ -333,6 +332,7 @@ function loadData() {
 
     // textual diff of objects
     if (JSON.stringify(old) == JSON.stringify(last)) {
+      console.info("data isn't different, will not send to clients");
       return false;
     }
     return true;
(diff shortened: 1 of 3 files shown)


bewest:
Will give this closer look tomorrow, if you can stand to wait that long. ;) Brief glimpse I like.

jasoncalabrese:
Not many lines changes, but the effect is big, don't want to rush testing it

bewest:
Will want to think about making some of this testable at some point.

bewest:
The image in the PR with "Update Time vs Data Time" that only happens on tooltip/touching the "minutes ago?"  For some reason I find the wording confusing.  What might be good is some kind of visual countdown to the next tick paired with the min ago, but not sure.

jimsiff:
I don't remember what @hackingtype1 calls it in Evan's CGM but I recall that his description of "Last Update" and "Last Data" seemed intuitive to me.

jasoncalabrese:
I checked with @hackingtype1 this is what the times look like in his system.  Maybe we should only add the last update time ago?  Then even if the data was old you'd be able to see that we did get a response from the server.

10937531_10153322424700166_104495645_n

bewest:
I don't think it's intuitive, I would have hard time explaining this in facebook or over the phone.  I think it would be more evident what is happening for their to be a countdown timer/indicator that clearly causes the display to update, paired with battery and min ago.  That said, maybe I'll implement it next rev.

hackingtype1:
Keeping tabs on the last uploader heartbeat and last receive time are hold overs from early reliability work; I still like them, but I'm likely the only viewer that takes an interest in them. More importantly, the backend service and views should indicate "out of range" or "uploader offline" errors. 
The school nurse appreciates being informed when the uploader has failed or when he's out of range, as do my wife and I :-) 

-John

jasoncalabrese:
I'm going to pull the tooltip out for now, I haven't thought to look at in the last couple days it was on my site.  We can think about extra timing info that should be available as a seperate feature

jasoncalabrese:
I'm planing to merge this and create the 0.6.1 release PR tonight

PR #353, opened 2015-01-17
Add missing else to fix error checking code

cjo20:
Missing else would have caused check to be ignored.

### DIFF
--- a/static/js/client.js
+++ b/static/js/client.js
@@ -1510,7 +1510,7 @@ var app = {}, browserSettings = {}, browserStorage = $.localStorage;
                 var iobContrib=0;
                 var activityContrib=0;
             }
-            if (minAgo < peak) {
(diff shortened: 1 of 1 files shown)


scottleibrand:
Sorry, I've been too busy with work to look at this.  @jasoncalabrese: would this be better targeted against dev?

jasoncalabrese:
The change is simple, but I'm not sure it's right, the effect is pretty big.

Without this change:
screen shot 2015-01-24 at 3 00 47 pm

With this change:
screen shot 2015-01-24 at 3 00 58 pm

scottleibrand:
Hmm, ok, I'll have to go back and walk through the algorithm again to see
how exactly it's implemented.  Since empirically the existing code is
making accurate predictions, let's not change it without further review.

PR #359, opened 2015-01-21
Added ALARM_TYPES and BG_ vars and descriptions to Heroku button.

jimsiff:
ALARM_TYPES is mandatory and set to predict to override BG_ optional vars.  BG_ vars are set to implicit defaults.  Sneaking this into wip/time-update-fix because it will likely be merged into release/0.6.1.

### DIFF
--- a/app.json
+++ b/app.json
@@ -12,6 +12,31 @@
       "value": "",
       "required": true
     },
+    "ALARM_TYPES": {
+      "description": "Controls Nightscout alarm behavior.  Default is 'predict'.  For odjustable alarm thresholds (set below), set to 'simple'.",
+      "value": "predict",
+      "required": true
+    },
+    "BG_HIGH": {
+      "description": "Optional Urgent High alarm BG value.  Default is 260.  Only used with simple alarms.",
+      "value": "260",
+      "requred": false
+    },
+    "BG_TARGET_TOP": {
+      "description": "Optional Non-urgent high alarm BG value.  This is the top of your target range.  Default is 180.  Only used with simple alarms.",
+      "value": "180",
+      "required": false
+    },
+    "BG_TARGET_BOTTOM": {
+      "description": "Optional Non urgent low alarm BG value.  This is the bottom of your target range.  Default is 80.  Only used with simple alarms.",
+      "value": "80",
+      "required": false
+    },
+    "BG_LOW": {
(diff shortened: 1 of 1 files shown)


jimsiff:
@jasoncalabrese Can we sneak this into 0.6.1 to ease Dreamsicle deployment / setup for Heroku?  

Also, we've had a few users assume the Alarm check boxes in the Settings drawer are active, even though they don't have ALARM_TYPES set to simple.  Can we add a note to the Settings drawer that will hopefully draw their attention?  I tried to tweak the text as shown in the picture below, but I don't know CSS and couldn't figure out the best way to reduce the font size.

screen shot 2015-01-21 at 2 01 40 pm

jasoncalabrese:
They should basically work using predict mode too, maybe in predict mode the BGs should be removed from the labels

gregwaehner:
Jim, Jason –

 

I thought, from the way, I read the code, that if ANY BG_ env. Variable was set, that the ALARM_TYPES was set to simple (actually thresholdset). This was written in the readme….

 

And the commit changes here has a good documentation log too…

https://github.com/nightscout/cgm-remote-monitor/commit/da1ab494dde1f678033fb9dc030b416e52628eb8

 

So can someone drive final answer on this?

 

-          G

jimsiff:
@gregwaehner Yes, that's my understanding as well.  The reason why I set ALARM_TYPES to mandatory with a value of predict is to preserve the intended default behavior even when all BG_ variables are set in the Heroku deployment.  

@jasoncalabrese does the code treat a null value for BG_* variables as the variable not set?  If not, we could add all Alarm variables as optional, with null values and preserve the default  behavior.

jimsiff:
@jasoncalabrese I agree that simply removing the BG values might be simpler and a little less confusing.  It also clarifies things for me.  I gave a user incorrect info earlier when I told her the check boxes didn't work without simple alarms.  I should have stated that they don't work the way she expected without simple alarms being set.

jasoncalabrese:
@jimsiff an empty value for any of the env var should be considered missing and get the default, but we should test it to be sure

jasoncalabrese:
Also the thresholds are used to draw the lines, so they are useful in predict mode, I'm not really happy with the current defaulting rules for `ALARM_TYPES` tempting to always have it default to predict

jasoncalabrese:
I'm going to open a new branch/PR for this and get wip/time-update-fix merged first

PR #360, opened 2015-01-22
only show the alarm threshold values in the alarm setting when simple mode is enabled

jasoncalabrese:
screen shot 2015-01-21 at 6 24 34 pm

screen shot 2015-01-21 at 6 24 54 pm

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -75,7 +75,7 @@
                         <dd><input type="radio" name="timeformat-browser" id="24-browser" value="24" /><label for="24-browser">24 hours</label></dd>
                     </dl>
                     <dl class="toggle">
-                        <dt>Enable Alarms <a class="tip" original-title="When enabled the an alarm will sound."><i class="icon-help-circled"></i></a></dt>
+                        <dt>Enable Alarms <a class="tip" original-title="When enabled the an alarm may sound."><i class="icon-help-circled"></i></a></dt>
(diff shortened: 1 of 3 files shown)


jasoncalabrese:
@jimsiff you can target your heroku deploy button changes here

jimsiff:
@jasoncalabrese @gregwaehner I created two new Heroku deploy branches. The first uses the same changes from yesterday where ALARM_TYPES was hard set to 'predict' because the BG_* values were hard set to their implied default.  That deploy behavior works, but it's not the cleanest because it hard sets variables that don't need to be set.  

The second branch leaves all alarm related variables blank with simple descriptions that highlight the implied default.  I think the second approach is cleaner and the site deploys as originally intended with implied defaults.  

Or, we can leave these variables completely out of the Heroku deployment and have users manually create and define any optional variables.  If we follow that approach to it's logical conclusion, should we then also remove the enable and pushover related variables too?  It's a balance between simplicity vs flexibility.

See the pics below:

screen shot 2015-01-22 at 1 06 34 pm

screen shot 2015-01-22 at 1 07 59 pm

jasoncalabrese:
The 2nd image with the empty values and defaults in the description make sense to me

jimsiff:
PR #364 used the empty values and defaults in the description.

PR #366, opened 2015-01-23
Fix bug that caused MBG outlines to collect at the far right of the context area

jasoncalabrese:


jasoncalabrese:
Example of bug:

screen shot 2015-01-22 at 11 25 38 pm

PR #376, opened 2015-01-27
dreamsicle

Karahbevin:


jasoncalabrese:
This is in the wrong direction I sent you a new pull request https://github.com/Karahbevin/cgm-remote-monitor/pull/3, just hit the merge button.  Note we found an issue with Night Mode, if that is an important feature to you can wait for the 0.6.2 release that should be soon.

PR #388, opened 2015-01-31

dexcom2014:
dreamsicle


jasoncalabrese:
this is going in the wrong direction, but it looks like you were able to update another way

dexcom2014:
Stressed but it worked somehow, thanks for responding. Have a good weekend!

PR #386, opened 2015-01-30
Option(s) to Enable Raw Data

jasoncalabrese:
Goal is a have a lab that will allow more advanced users to enable raw data in the mainline version

**New `ENABLE` Option**:
`ENABLE="rawbg"` or `ENABLE="rawbg-on"` (to default to on)

**Tasks:**

  - [X] Port Basic Raw Data code from `wip/iob/cob` branch
  - [x] Refactor code that filters/sends errors codes
  - [x] Add options to enable Raw Data via the `ENABLE` env var
  - [x] Based on enable option and UI settings display show/hide Raw Data
  - [x] Based on enable option include/exclude `cal` records and filtered, unfiltered, and rssi rawbg fields in `/pebble` endpoint

PR #394, opened 2015-02-02
Raw data part 2

jasoncalabrese:
Follow up to #386 that got merged before review/discussion was complete

The main issue remaining on #386 was if raw should always be display or only when there is noise.  To experiment with that I created a 3-way option

screen shot 2015-02-01 at 1 10 52 pm

After we get a chance to play with this in different situations it should be easier to come to a consensus on what we think is best for the community.

## A - Show Raw Always
a

## B - Show Raw when there is noise
b

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -167,6 +167,7 @@ function update() {
                   obj.direction = directionToChar(element.direction);
                   obj.filtered = element.filtered;
                   obj.unfiltered = element.unfiltered;
+                  obj.noise = element.noise;
                   obj.rssi = element.rssi;
                   cgmData.push(obj);
               } else if (element.slope) {
--- a/static/index.html
+++ b/static/index.html
@@ -87,7 +87,9 @@
                     </dl>
                     <dl id="show-rawbg-option" class="toggle">
                       <dt>Show Raw BG Data <a class="tip" original-title="When enabled small white dots will be disaplyed for raw BG data"><i class="icon-help-circled"></i></a></dt>
-                      <dd><input type="checkbox" name="show-rawbg" id="show-rawbg" /><label for="show-rawbg">Enable</label></dd>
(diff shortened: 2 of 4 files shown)


jasoncalabrese:
@jimsiff, @YYGIRL, @ELUTE, @hackingtype1 continuing from #386 ...

jasoncalabrese:
1 bonus of only showing raw when there is noise is the chart should paint faster since there will be much fewer dots.

I'm going to try to find a situation where raw would be useful and there and the noise level is clean.  If we can't find an example we can simplify things.

hackingtype1:
By Dexcom's methods: Clean noise = Raw based EGV (displayed on receiver). 
The only time I'd show a raw value with a clean noise classification is if I was applying an entirely separate calibration algorithm to the raw value. 
Otherwise, the "raw" value being shown, while the noise is clean, is an incomplete (but derived from Dexcom's method) EGV calculation... 

-John

jasoncalabrese:
Added Noise level to the tooltip to help understand when the raw data is displayed or not

screen shot 2015-02-04 at 5 02 15 pm

screen shot 2015-02-04 at 5 02 50 pm

jasoncalabrese:
@jimsiff I noticed last night that sensor data isn't getting uploaded when Gap Sync is used, not sure if that explains anything you're seeing now

jimsiff:
@trhodeos @jasoncalabrese Turns out the lack of noise was a layer 8 issue, solved when I upgraded the uploader feeding my raw site.  Noise is there with 0.1.12 via REST, and I like the tooltip update.  More real testing to follow now that I've debugged myself.

trhodeos:
@jimsiff ok, cool. Glad to hear you got it working.

@jasoncalabrese could you open a PR or issue for the gapsync issue?

PR #406, opened 2015-02-09
Syntax error in "Enable Alarms" help prompt

Trusties13:
Attempt number 3! @jasoncalabrese !!!

Lets hope it all works now!

Issue #402, opened 2015-02-08
Prediction cone values are different in MMOL setting


baswieb:
The prediction cone looks different in MMOL setting because the BG values are rounded to integer values. When the values are rounded to 1 decimal the prediction values look better in MMOL setting.

jasoncalabrese:
@baswieb I think I have a fix, do you know how to test with different branches?  I opened a PR with the fix #405

jasoncalabrese:
Will be part of the 0.6.3 release that will hopefully be out very soon

baswieb:
@jasoncalabrese Thanks! It's fixed

PR #397, opened 2015-02-08
fixed HIGH/LOW dispay on phablets; moved time hide/show from js to css

jasoncalabrese:
Fixes problems like this and similar with HIGH/LOW on iPhone6+ and Nexus6

PR #410, opened 2015-02-12
Dreamsicle (Release 0.6.3)

jasoncalabrese:
# Features:

### Options to enable raw data #386 & #394

To enable Raw Data update the `ENABLE` environment variable to include `rawbg`, if you want raw BG's to display by default use `rawbg-on`.  For example `ENABLE="careportal rawbg"`

screen shot 2015-02-01 at 1 10 52 pm

#### Show Raw - Always
a

#### Show Raw - When there is noise
b

# Bug Fixes and Improvements:

  * fixed HIGH/LOW display on iPhone 6+ and Nexus 6 #397
  * don't round the predicted values when mmol display mode #402/#405
  * Syntax error in "Enable Alarms" help prompt #406
  * Store display units with treatment objects, for unit conversion / checking on load. #408
  * if there isn't a glucose value, don't store the units #409


bewest:
Lovely, thank you!

jimsiff:
+1000.  IMO, the most significant release in terms of proactively managing BG during times of rapid change.  Just huge.

PR #416, opened 2015-02-15
fix coveralls badge branch

dduugg:


dduugg:
F!@#ing coveralls. I think this needs to wait for #417 to merge.

PR #424, opened 2015-02-18
create all docs before calling back

### LINKED ISSUE #423: test #2 is flaky
Travis often fails when running test #2. Example [here](https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/50449625#L369).

Interestingly, running the same test in succession can produce inconsistent results. In [this commit](https://github.com/dduugg/cgm-remote-monitor/commit/7c31f0def8dadfe5315b21d9b9559cc97bb7e2ea), I duplicated the flaky test (compare lines 30-39 with lines 41-50). Travis [failed](https://travis-ci.org/dduugg/cgm-remote-monitor/jobs/51033835#L397) the first test but [passed](
https://travis-ci.org/dduugg/cgm-remote-monitor/jobs/51033835#L457) the second.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8640215-test-2-is-flaky?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dduugg:
This resolves an issue where the `create` function in `entries.js` wasn't waiting for the updates to complete before calling back. This would frequently cause Travis failures, when tests would expect entries that had yet to be completed.

resolves #423

### DIFF
--- a/lib/entries.js
+++ b/lib/entries.js
@@ -96,21 +96,23 @@ function storage(name, storage, pushover) {
 
   // store new documents using the storage mechanism
   function create (docs, fn) {
-      with_collection(function(err, collection) {
-        if (err) { fn(err); return; }
-        // potentially a batch insert
-        var firstErr = null,
-            totalCreated = 0;
+    with_collection(function(err, collection) {
+      if (err) { fn(err); return; }
+      // potentially a batch insert
+      var firstErr = null,
+          numDocs = docs.length,
+          totalCreated = 0;
 
-        docs.forEach(function(doc) {
-            collection.update(doc, doc, {upsert: true}, function (err, created) {
-                firstErr = firstErr || err;
-                totalCreated += created;
-            });
-            sendPushover(doc);
+      docs.forEach(function(doc) {
+        collection.update(doc, doc, {upsert: true}, function (err, created) {
+          firstErr = firstErr || err;
+          if (++totalCreated === numDocs) {
+            fn(firstErr, docs);
+          }
         });
-        fn(firstErr, totalCreated, docs);
+        sendPushover(doc);
       });
+    });
   }
 
   //currently the Android upload will send the last MBG over and over, make sure we get a single notification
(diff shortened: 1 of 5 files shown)


lib/entries.js
```diff
@@ -96,77 +92,78 @@ function storage(name, storage, pushover) {
...
-      with_collection(function(err, collection) {
-        if (err) { fn(err); return; }
-        // potentially a batch insert
-        var firstErr = null,
-            totalCreated = 0;
-
-        docs.forEach(function(doc) {
-            collection.update(doc, doc, {upsert: true}, function (err, created) {
-                firstErr = firstErr || err;
-                totalCreated += created;
```
dduugg: This is especially sketchy, as `created` is an object, but the initial value of `totalCreated` is a number, turning `totalCreated` into a string of the form `0{"ok":1,"nModified":1,"n":1}{"ok":1,"nModified":1,"n":1}`...

lib/entries.js
```diff
@@ -96,77 +92,78 @@ function storage(name, storage, pushover) {
...
+      if (err) { fn(err); return; }
+      // potentially a batch insert
+      var firstErr = null,
+          totalCreated = 0;
+
+      docs.forEach(function(doc) {
+        collection.update(doc, doc, {upsert: true}, function (err, created) {
+          firstErr = firstErr || err;
+          if (++totalCreated >= docs.length) {
+            fn(firstErr, totalCreated, docs);
```
dduugg: I don't like this fix at all, but it seems to work. I'm hoping someone has a better idea...

bewest:
Such a fan of this, I created PR, is this good to go or does it need to wait for some reason?

dduugg:
It's basically good to go from my perspective. One could argue that tests 3 and 2 should be swapped (since test 3 checks for more docs, it's more likely to catch a regression). One could also nitpick that `docs.length` should not be accessed in every loop iteration (it should be stored as a variable, or the counter should count down to zero).

dduugg:
Just pushed a commit with precisely those changes.

jasoncalabrese:
Been out of town, want to a do a quick test and merge this tonight or tomorrow.

jasoncalabrese:
Tested, thanks @dduugg

PR #436, opened 2015-02-25
allow arbitrary DIAs in profile

mgranberry:
This simply linearly scales the input to match the 3h IOB curve.


jasoncalabrese:
I'm planning to test this tonight, after merging I'll update `pebble.js` to do the same

We need to have a simple way to share the code between the client and the server, need to be done before IOB goes to the mainline.

jasoncalabrese:
Just finished testing, nice work @mgranberry

PR #439, opened 2015-02-26
Page to enter and view a treatment profile

jasoncalabrese:
This PR is to make it easier to bring some of the work @MilosKozak to the mainline and to prepare for adding and option to enable IOB in the mainline

### DIFF
--- a/lib/websocket.js
+++ b/lib/websocket.js
@@ -1,5 +1,5 @@
 
-function websocket (env, server, entries, treatments) {
+function websocket (env, server, entries, treatments, profiles) {
 "use strict";
 // CONSTANTS
 var ONE_HOUR = 3600000,
@@ -29,6 +29,7 @@ var dir2Char = {
     treatmentData = [],
     mbgData = [],
     calData = [],
+    profileData = [],
     patientData = [];
 
   function start ( ) {
@@ -146,6 +147,7 @@ function update() {
     cgmData = [];
     treatmentData = [];
     mbgData = [];
+    profileData = [];
     var earliest_data = now - TWO_DAYS;
     var q = { find: {"date": {"$gte": earliest_data}} };
     entries.list(q, function (err, results) {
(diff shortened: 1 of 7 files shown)


jasoncalabrese:
Need to add IOB tests

lib/websocket.js
```diff
@@ -188,8 +190,19 @@ function update() {
...
+
+          profiles.list(function (err, results) {
+              // There should be only one document in the profile collection with a DIA.  If there are multiple, use the last one.
+              results.forEach(function(element, index, array) {
+                  if (element) {
+                    if (element.dia) {
+                        profileData[0] = element;
+                    }
+                  }
+              });
```
dduugg: Based on the comment, it looks like you're accessing the last element of results using an O(n) approach. If so, consider using an O(1) implementation, such as:

```
if (results.length > 0) {
  var element = results[results.length - 1];
  if (element && element.dia) {
    profileData[0] = element;
  }
}
```
(and you may not even need the outermost `if`)

lib/websocket.js
```diff
@@ -188,8 +190,19 @@ function update() {
...
               return treatment;
           });
-          // all done, do loadData
-          loadData( );
+
+          profiles.list(function (err, results) {
+              // There should be only one document in the profile collection with a DIA.  If there are multiple, use the last one.
+              results.forEach(function(element, index, array) {
+                  if (element) {
+                    if (element.dia) {
```
dduugg: Lines 197-198 could be simplified to `if (element && element.dia) {`, correct?

Also, consider a standard `for` loop that goes backward from the end of `results`, and breaks when it finds `element.dia`.

lib/websocket.js
```diff
@@ -237,6 +250,10 @@ function loadData() {
         });
     }
 
+    if (profileData) {
```
dduugg: Is this ever false? (an empty array is truthy).

jasoncalabrese:
@dduugg this was a quick port from the iob-cob branch, was planning on doing some cleanup, not sure if that happens here or with #452

PR #440, opened 2015-02-26
A system DEFAULTS env var

jasoncalabrese:
Will be used for system options such: display units; time format; show rawbg never,always,noise; etc


jasoncalabrese:
This work will be done in #439

PR #452, opened 2015-03-01
Add an option to enable IOB in the mainline

jasoncalabrese:
screen shot 2015-03-01 at 11 12 00 am

Like the Care Portal and Raw Data this feature will need to be enabled using the `ENABLE` env var.  For example: `ENABLE="careportal rawbg iob"`

This will use a treatment profile or default to a DIA of 3 hours.  A UI for setting the treatment profile is begin working on with #439 

This also make IOB available to the pebble endpoint


jasoncalabrese:
Need to test this on azure, and make sure the bundle can be written, etc

lib/iob.js
```diff
@@ -0,0 +1,80 @@
...
+function calcTotal(treatments, profile, time) {
+
+  console.info("trying to calc");
+
+  var iob = 0
+    , activity = 0;
+
+  if (!treatments) return {};
+
+  if (typeof profile === 'undefined') {
```
dduugg: Just curious, is this any different from `if (profile === undefined) {` (which is more readable)?

package.json
```diff
@@ -43,7 +43,8 @@
     "pushover-notifications": "0.2.0",
     "sgvdata": "0.0.2",
     "socket.io": "^0.9.17",
-    "git-rev": "git://github.com/bewest/git-rev.git"
+    "git-rev": "git://github.com/bewest/git-rev.git",
+    "browserify-express": "^0.1.4"
```
dduugg: consider using `npm install <dependency> --save` or manually slotting this alphabetically — it will simplify future diffs.

jasoncalabrese:
Trying out a new pill based display for the BG delta and IOB, these don't get the strikethrough style in retro mode.

iob-pill-noalarm

iob-pill-withalarm

iob-pill-retro

jimsiff:
I like that… looks really good.  One thing I noticed with the delta and IOB below BG and trend… the straight down arrow clips IOB in the older style.  Not sure if it happens with your new pill design.

francescaneri:
image

francescaneri:
Hi Jason, you can separate them a thread? Thanks Matteo

francescaneri:
image

francescaneri:
....with the oblique arrow are lowered.

francescaneri:
image

jasoncalabrese:
This gets even harder when you add COB and BGI.  Also want room for uploader battery, noise level, etc.  We'd have options to pick what you want to see, but still more pills.  And then to keep it responsive from desktop to tablet to iPhone4.

We really need to move to fixed sizes and get rid the % based font sizes.  We can have multiple sizes for different device targets.

jasoncalabrese:
Think we'll need multiple lines, and on a narrow screen 3

and iob cob bgi

and raw iob cob bgi

jimsiff:
Hey Jason,

I like the look of the pills.  It really cleans up the interface.  

For my eyes, the grey on black colors and small fonts don't provide a lot of contrast.  I have to look harder at the data to read it vs the current green / yellow / red on black.  

Would white on black look out of place?

Jim

ELUTE:
I was thinking the exact same thing Jim

jasoncalabrese:
Maybe a brighter grey? So not fully white?  Icons might be good too, but then we have to move to fixed font sizes.  I want to be able to plug in new types of data really easy.

jimsiff:
How does this look?  Changed the pill background color to #bdbdbd.  

screen shot 2015-03-06 at 11 28 43 am

screen shot 2015-03-06 at 11 29 13 am

francescaneri:
@jimsiff 👍

jasoncalabrese:
Looks good.  We should probably use the same grey for the value in the pill too.  Should we replace the other grey?  Clock, time ago, BG (retro and default theme), etc.

jimsiff:
Sounds good.  I do think the brighter grey pops more against the black... I'd like to see what it looks like with clock, time ago, BG, etc using that.  I hope it doesn't kill the effectiveness of Night mode.

jimsiff:
It does look like the straight down trend bleeds into the pill.  It's so hard to juggle everything with % fonts.

screen shot 2015-03-06 at 12 06 59 pm

screen shot 2015-03-06 at 12 11 59 pm

francescaneri:
@jimsiff @jasoncalabrese 
raw
...and if so for raw data

francescaneri:
Hi everyone, you can not keep a new display based pill and place as it was put by wip / IOB -COB  have a screen cleaner ? what do you say @scottleibrand

francescaneri:
image

jasoncalabrese:
@francescaneri, trying to fit the IOB/COB into the title has been a problem.  That layout is really fragile, screws up the click location of the toolbar, and runs into the title.

jasoncalabrese:
I'm going to merge this to dev and start getting it integrated with everything else

PR #458, opened 2015-03-03
profile editor. env defaults

MilosKozak:
TODO:
old env settings migration to new DEFAULTS with lower priority
change code to use new DEFAULTS

### DIFF
--- a/env.js
+++ b/env.js
@@ -46,7 +46,7 @@ function config ( ) {
   env.enable = readENV('ENABLE');
 
   // parse defaults here, expect format `key1=value1 key2=value2` split on space, ignore commas
-  var rawdefaults = readENV('DEFAULTS');
+  var rawdefaults = readENV('DEFAULTS','');
   env.defaults = { // currently supported keys must defined be here
 	  'time-format': '12'			// 12|24
 	, 'display-units': 'mg/dL'		// 'mg/dL'|'mmol'
@@ -75,7 +75,9 @@ function config ( ) {
 		  var key = a[0];
 		  var value = a[1];
 		  if (alloweddefaults[key] && (alloweddefaults[key][value] || alloweddefaults[key]['any'])) {
-			  env.defaults[key] = value.replace(/__/g,' ');
+			  value = value.replace(/__/g,' ');
+			  env.defaults[key] = value;
+			  console.log('DEFAULTS '+key+'='+value);
 		  } else {
 			console.log('DEFAULTS key: '+key+' value: '+value+' not allowed');
 		  }
--- a/env.js
+++ b/env.js
@@ -45,11 +45,49 @@ function config ( ) {
 
   env.enable = readENV('ENABLE');
 
-  //TODO: parse defaults here, expect format `key1=value1 key2=value2` split on space, ignore commas
-  // results should be an object something like env.defaults = {"DISPLAY_UNITS": "mg/dl", "TIME_FORMAT": "12"};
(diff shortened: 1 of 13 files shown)


static/js/profile.js
```diff
@@ -0,0 +1,539 @@
...
+			$('#pe_simple').hide("slow");
+			$('#pe_calculator').show("slow");
+		}
+		if (event) {
+			event.preventDefault();
+		}
+	}
+	
+	function fillFrom(event) {
+		var mmoltime = [
```
sulkaharo: Weird naming alert! Seems to assume mmols and mg/dl usage correlates to whether a country uses the 24 hour clock, or not, while it shouldn't.

static/js/profile.js
```diff
@@ -0,0 +1,539 @@
...
+			"carbratio":24,
+			"carbs_hr":30,
+			"sens":17,
+			
+			"validfrom": new Date(),
+			
+			//Bolus calculator style values, "from" are in minutes from midnight
+			"ic": [{"from":0,"val":24},
+					{"from":0,"val":0},
+					{"from":0,"val":0},
```
sulkaharo: Would be way cleaner if the number of entries was controlled programmatically and this didn't assume a set number of values.

static/js/profile.js
```diff
@@ -0,0 +1,539 @@
...
+			{ "html":"pe_if", 			"type":"int" , 			"settings":"profile.carbratio" },
+			{ "html":"pe_hr", 			"type":"int" , 			"settings":"profile.carbs_hr" },
+			{ "html":"pe_isf", 			"type":"int" , 			"settings":"profile.sens" },
+			
+			//Bolus calculator style values
+			{ "html":"pe_hr_high", 		"type":"int" , 			"settings":"profile.car_high" },
+			{ "html":"pe_hr_medium",	"type":"int" , 			"settings":"profile.car_medium" },
+			{ "html":"pe_hr_low", 		"type":"int" , 			"settings":"profile.car_low" },
+			
+			{ "html":"pe_if_from_0", 	"type":"dropdownval" ,	"settings":"profile.ic[0].from" },
```
sulkaharo: Is there a way to create these programatically? I can see how hardcoding the entries was the easiest way to get going, but this would be way cleaner if the number of entries wasn't hardcoded.

MilosKozak: well so far i did it the same way i found in our animas pump. so # of intervals is hardcoded but i worked on gui to have it easy usable. when you reach midnight the rest of intervals are disabled. I prefer to have fixed # if intervals to keep formating on page easy

static/js/profile.js
MilosKozak: currently is used timeformat based on env DEFAULTS. selected units are displayed on top of the page but not used for anything currently. all data are stored in raw format from html. we need to discuss format of profile JSON

env.js
```diff
@@ -45,6 +44,49 @@ function config ( ) {
   env.devicestatus_collection = readENV('MONGO_DEVICESTATUS_COLLECTION', 'devicestatus');
 
   env.enable = readENV('ENABLE');
+
+  // parse defaults here, expect format `key1=value1 key2=value2` split on space, ignore commas
+  var rawdefaults = readENV('DEFAULTS');
```
jasoncalabrese: @MilosKozak the tests are failing because there is no value for `DEFAULTS` try:
```diff
-  var rawdefaults = readENV('DEFAULTS');
+  var rawdefaults = readENV('DEFAULTS', '');
```

PR #476, opened 2015-03-12
Always return the last Cal entry (don't worry about how old it might be)

jasoncalabrese:
Also some refactoring by using async.parallel to make multiple ongo requests at once.

Bonus fix: also check `env.DISPLAY_UNITS` if the `units` query string param isn't set

@bewest @ELUTE @jimsiff

### DIFF
--- a/lib/pebble.js
+++ b/lib/pebble.js
@@ -12,6 +12,7 @@ var DIRECTIONS = {
 };
 
 var iob = require("./iob")();
+var async = require('async');
 
 function directionToTrend (direction) {
   var trend = 8;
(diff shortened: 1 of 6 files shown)


jasoncalabrese:
@jimsiff, @ELUTE, I talked to @YYGIRL about some users are burning through data partially due to the extra `cal` data returned from `/pebble`.  We merged in a fix (6357212222eb0a6080f71c34bd76d6e4262dd83d) that selectively chooses the fields that are needed. The plan is to have some of those users that are most concerned about data usage to start using `dev` after this PR is merged.  Then they can also help test before we release Enchilada to master.

jimsiff:
JSON payload is reduced about 40% but TCP and HTTP protocol overhead drops the net savings to 20%.  Gzip encoding may help some more.

jasoncalabrese:
thanks for testing @jimsiff did you also take a look at the main web monitor page, in addition to `/pebble`?

jasoncalabrese:
I just added gzip compression to everything, with that we're down to 255 bytes (455 with http overhead)

PR #479, opened 2015-03-14
Fix rendering of Care Portal entries that do not define BG...

sulkaharo:
... But which define insulin OR carbs


jasoncalabrese:
fixed by #481 thanks for your help @sulkaharo

PR #481, opened 2015-03-14
better treatment BG unit handling

jasoncalabrese:


sulkaharo:
:+1:

static/js/client.js
```diff
@@ -1078,28 +1078,43 @@ var app = {}, browserSettings = {}, browserStorage = $.localStorage;
...
+            return false;
+          } else {
+            return Math.abs((new Date(d.date)).getTime() - time) <= SIX_MINS_IN_MS;
+          }
         });
 
         var totalBG = 0;
         closeBGs.forEach(function(d) {
-            totalBG += d.y;
+          totalBG += d.y;
```
sulkaharo: This needs to be totalBG += Number(d.y); for it to work. The values are strings and mg/dl calculations fail without the parsing.

sulkaharo:
Ok added one comment / fix -> with that change mixed treatment record units render correctly regardless of the display units. :)

PR #495, opened 2015-03-17
Profile editor, env.defaults finished

MilosKozak:


env.js
```diff
@@ -45,6 +44,66 @@ function config ( ) {
   env.devicestatus_collection = readENV('MONGO_DEVICESTATUS_COLLECTION', 'devicestatus');
 
   env.enable = readENV('ENABLE');
+
+  env.DISPLAY_UNITS = readENV('DISPLAY_UNITS', 'mg/dl');
+
+  // parse defaults here, expect format `key1=value1 key2=value2` split on space, ignore commas
+  var rawdefaults = readENV('DEFAULTS','');
+  env.defaults = { // currently supported keys must defined be here
```
jasoncalabrese: The original idea for the defaults env was to make it a take arbitrary key=value pairs.  We've since dissed on gitter and have decided that we don't want to to that.

lib/api/profile/index.js
```diff
@@ -15,13 +15,60 @@ function configure (app, wares, profile) {
...
 
-    // List settings available
+    // List profiles available
     api.get('/profile/', function(req, res) {
         profile.list(function (err, attribute) {
             return res.json(attribute);
         });
     });
 
+	// List current active record (in current state LAST record is current active)
```
jasoncalabrese: It would be nice to get all these changes in `lib/api/profile/index.js` in a separate/simple PR

static/css/drawer.css
```diff
@@ -164,11 +165,21 @@ h1, legend,
     margin-top: 0;
     margin-left: 42px;
     padding-top: 10px;
-    padding-right: 150px;
+    padding-right: 90px;
     white-space: nowrap;
     text-overflow: ellipsis;
     overflow: hidden;
 }
+#iob-cob h1 {
```
jasoncalabrese: this is targeting `dev` there is no `#iob-cob` yet, no reason to the adjusting it's style here

static/js/profile.js
```diff
@@ -0,0 +1,645 @@
+/* Code by Milos Kozak
+
+TODO:
+	- treatment drawer must be updated to enter glycemic index
+	- update calculation to use new style of values
+	- how to handle units? store native or convert to mg/dL?
+
+*/
+	var c_profile = null;
+	var apisecret = "", storeapisecret = false;
```
jasoncalabrese: we shouldn't store the unhashed `apisecret` in memory or in localstorage, instead we should hash it right away

static/js/profile.js
```diff
@@ -0,0 +1,645 @@
...
+					{"from":0,"val":0},
+					{"from":0,"val":0},
+					{"from":0,"val":0},
+					{"from":0,"val":0},
+					{"from":0,"val":0},
+					{"from":0,"val":0}
+					]
+	};
+
+	var GuiToVal = [
```
jasoncalabrese: can we do something different here, this is UI info and should be in the markup

static/js/profile.js
```diff
@@ -0,0 +1,645 @@
...
+			$('#pe_simple').hide("slow");
+			$('#pe_calculator').show("slow");
+		}
+		if (event) {
+			event.preventDefault();
+		}
+	}
+	
+	function fillFrom(event) {
+		var mmoltime = [
```
jasoncalabrese: why do we need fixed times, pump UI's are like this, but we have way more flexibility

static/js/profile.js
MilosKozak: thx. i will change it

static/js/profile.js
MilosKozak: lot of work. but i'll try

MilosKozak:
One more thing. You're sending profile record via websocket everytime. I'm calling api to get it once at start. Is it possible for you as profile is updated rarely? We'll save traffic ..

static/css/drawer.css
MilosKozak: these changes must be from some rebase. i need to add only color to legend. legend was not displayed properly on IE.
How to solve it?

jasoncalabrese: maybe that id is getting used in an unexpected way, on the other branch it for the iob/cob values

PR #500, opened 2015-03-18
Updated profile api

MilosKozak:
trying to pull 1 by 1

### DIFF
--- a/lib/profile.js
+++ b/lib/profile.js
@@ -1,24 +1,57 @@
 'use strict';
 
-function configure (collection, storage) {
+var ObjectID = require('mongodb').ObjectID;
 
+var lastrecord = null;
+
+function storage (collection, storage) {
+	
   function create (obj, fn) {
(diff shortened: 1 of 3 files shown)


jasoncalabrese:
Looks good, thanks @MilosKozak I'll test and merge this tonight

lib/profile.js
```diff
@@ -1,24 +1,55 @@
...
+	obj._id = new ObjectID(obj._id);
+    obj.created_at = (new Date( )).toISOString( );
+    api().save(obj, function (err, doc) {
+      fn(err, doc);
+    });
+  }
+
   function list (fn) {
-    return api( ).find({ }).sort({created_at: -1}).toArray(fn);
+    return api().find({ }).sort({validfrom: -1}).toArray(fn);
```
jasoncalabrese: I don't think we should be introducing `validfrom` here, there is no data in that field yet to the sort will be undefined

PR #504, opened 2015-03-19
allow API to search things better

bewest:
This allows a query such as this to search for events of hypoglycemia for
example:
```
curl -g localhost:3434 \
     /api/v1/entries   \
     '?find[sgv][$lte]=70&find[sgv][$gte]=20&count=1000
```
It's possible to construct most mongo queries by url encoding the query string.
In this instance, mongo performs poorly when searching for lte/gte for strings.
In order for ranged queries to perform properly, the query parameters must be
set to integer type.  There is a quick and ugly helper to ensure that some sgv
queries will be respected as integer searches.

### DIFF
--- a/lib/api/entries/index.js
+++ b/lib/api/entries/index.js
@@ -37,7 +37,7 @@ function configure (app, wares, entries) {
           es.pipeline(output, sgvdata.format( ), res);
         },
         json: function ( ) {
-          es.pipeline(output, sgvdata.lint({strict: false}), es.writeArray(function (err, out) {
+          es.pipeline(output, entries.map( ), es.writeArray(function (err, out) {
             res.json(out);
           }));
         }
--- a/lib/api/treatments/index.js
+++ b/lib/api/treatments/index.js
@@ -17,7 +17,7 @@ function configure (app, wares, treatments) {
 
     // List settings available
     api.get('/treatments/', function(req, res) {
-        treatments.list({}, function (err, profiles) {
+        treatments.list({find: req.params}, function (err, profiles) {
             return res.json(profiles);
         });
     });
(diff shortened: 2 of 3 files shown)


bewest:
```bash

$ curl -g localhost:3434/api/v1/entries'?find[dateString][$lte]=2015-01-32&find[dateString][$gte]=2015-01-00&count=10000'  | wc -l && echo
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  559k    0  559k    0     0   187k      0 --:--:--  0:00:02 --:--:--  187k
9813

```

jasoncalabrese:
Cool, very useful

jasoncalabrese:
Looks like lots of data is getting filtered out, missing `cal` and `sensor` entries and filed like `unfiltered`, `filtered`, `slope`, `mbg`, `type`, etc, etc

jasoncalabrese:
and would be nice if `http://localhost:1337/api/v1/entries.json?find[mbg]&count=100` or `http://localhost:1337/api/v1/entries.json?find[type][mbg]&count=100` would work

bewest:
`?find[type]=mbg` will now search exclusively for `mbg` records.  Also json responses will include `mbg` and other data elements, so long as they have a `type` property.

I strongly encourage using iso8601 date encoding to make the date filtering features work with ease.

jasoncalabrese:
I agree on date format, I don't think the api really used yet so it's a good fine to change.  Maybe add a param to return UTC timestamps?

jasoncalabrese:
Works good should we do something with the date format here (export only) or wait till we do something bigger?

jasoncalabrese:
Lets look at date formatting outside of this, going to merge now

Issue #499, opened 2015-03-18
Ignore sensor noise level when Dex sends special value SGV


jimsiff:
If raw is displayed only when noise is present, raw can disappear during sensor warmup after 45-50 minutes when noise level switches from 5 to 1.  Per @hackingtype1 "It can and will switch from 5 to 1, etc - which is why some people report ??? during startup." and "it takes 45 minutes to get enough data to calculate noise level, after that, it uses the denoising algorithm even if it is masking EGV as part of the start up process."

jimsiff:
This still seems to be an issue.  After a sensor restart, raw was displayed while noise was 5, but then went away when noise switched to 1.

PR #506, opened 2015-03-20
display of noise level and rawbg value when rawbg is enabled

jasoncalabrese:
**Light**
screen shot 2015-03-20 at 12 08 44 am

**Medium**
screen shot 2015-03-20 at 12 08 34 am

**Heavy/Error**
screen shot 2015-03-20 at 12 08 22 am


jasoncalabrese:
This also resolves #499

PR #511, opened 2015-03-22
Wip/enable is optional

bewest:
Pebble crashes if there's no default value set.

* `ENABLE` should not be required.
* also `env.head` is used by mqqt and also needs a value: in `dokku` it needs to be set manually because the deployed asset is actually the output for `git archive`, not an actual git repo.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: node server.js
+web: ./node_modules/.bin/forever --minUptime 100 -c node server.js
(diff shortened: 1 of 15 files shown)


bewest:
Also add `forever.js` for heroku, dokku and anyone else using `Procfile`.

jasoncalabrese:
Tested with a fresh heroku deploy

PR #516, opened 2015-03-24
Dev

ccoliver:


jimsiff:
Going the wrong direction here... try this

https://github.com/nightscout/cgm-remote-monitor/compare/ccoliver:dev...nightscout:dev

PR #524, opened 2015-03-24
Allow BG thresholds in mmol 

MilosKozak:


jasoncalabrese:
Think we should still do this, but probably easier to start over now that the code has changed so much

PR #527, opened 2015-03-25
update API to support searching for all types

bewest:
Should enable searching for all data types.


bewest:
Not sure we want this in this release per se, but these changes help searching for data.

It would be trivial to add some `textKeys` property to format into csv/tsv.  With a json-schema, the api could also handle text inputs as well.

jasoncalabrese:
Just noticed that the test coverage drops almost 5% with this, we should add some basic test of the new features

PR #528, opened 2015-03-25
update doc title with BG and delta vals

justingshreve:
Small PR to display current BG and delta values in HTML title. Makes for quick reference while working in other tabs or windows, if you're using Nightscout in browser.

screen shot 2015-03-24 at 10 19 31 pm


jasoncalabrese:
We don't merge PR's directly to master, but something like this targeting dev could be a nice feature.  We  should probably include the custom label after the BG and delta since some people use multiple sites for a variety of reasons

justingshreve:
Okay, I'll close and reopen on dev, as well as append the custom title.

PR #532, opened 2015-03-26
Update doc title with BG and delta vals

justingshreve:
Small PR to display current BG and delta values in HTML title. Makes for quick reference while working in other tabs or windows, if you're using Nightscout in browser.

This PR was copied over from PR against master branch.

screen shot 2015-03-24 at 10 19 31 pm


justingshreve:
Ah man, need to add these features to dev, closing again!

PR #533, opened 2015-03-26
Bg and Delta to HTML title

justingshreve:
Okay, think I finally got the flow down right. Refactored my initial bg to title and created new function to do so. Appends custom page title after bg and delta.

### DIFF
--- a/static/js/client.js
+++ b/static/js/client.js
@@ -295,6 +295,18 @@ var app = {}, browserSettings = {}, browserStorage = $.localStorage;
         return errorDisplay;
     }
 
+    function updateTitle(prev, current) {
+        var delta = scaleBg(current) - scaleBg(prev);
+        if (browserSettings.units == 'mmol') {
+            delta = delta.toFixed(1);
+        }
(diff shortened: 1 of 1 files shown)


jasoncalabrese:
There are some changes in dev that I think will make us want to call `updateTitle()` from a different location.  Previously we did a full update when ever we got a new sgv event from socket.io, that caused a lot of drawing to get queued if the document wasn't visible.  I prevent that on https://github.com/nightscout/cgm-remote-monitor/blob/dev/static/js/client.js#L700.  I think we want to update the title before returning early from `updateChart()`  Otherwise the tab title will only update when the document is visible.

justingshreve:
Yes, I noticed values were not updating in title when doc was not visible. Moved call to updateTitle inside socket.io callback. Seems to be working now.

jasoncalabrese:
Nice addition thanks @justingshreve, I'm want to do a little refactoring now and make common delta function, then we can add interpolation, and use raw data when we don't have an svg

PR #540, opened 2015-04-01

LRyan6922:
Dreamsicle


jimsiff:
Looks like this is going the wrong direction.  Try this link instead:

https://github.com/nightscout/cgm-remote-monitor/compare/LRyan6922:master...nightscout:master

PR #543, opened 2015-04-03
improvements for the title, current bg status, and delta

jasoncalabrese:
screen shot 2015-04-02 at 11 23 59 pm

screen shot 2015-04-02 at 11 24 34 pm

screen shot 2015-04-02 at 11 26 57 pm

screen shot 2015-04-03 at 1 22 05 am

### DIFF
--- a/app.json
+++ b/app.json
@@ -12,6 +12,11 @@
       "value": "",
       "required": true
     },
+    "DISPLAY_UNITS": {
+      "description": "Server display units for BG values.  Default null value implies 'mg/dl'.  Set to 'mmol' to change the server to mmol mode.",
+      "value": "",
+      "required": false
+    },
     "ENABLE": {
       "description": "Space delimited list of optional features to enable, such as 'careportal'.",
       "value": "",
@@ -54,6 +59,7 @@
     }
   },
   "addons": [
-    "mongolab:sandbox"
+    "mongolab:sandbox",
+    "papertrail"
   ]
 }
--- a/static/css/drawer.css
+++ b/static/css/drawer.css
@@ -66,7 +66,7 @@
 }
 
 #eventTime {
-
+    padding-bottom: 5px;
 }
 
 #eventTime > span {
@@ -258,3 +258,27 @@ h1, legend,
     color: #fff;
     font-size: 1.5em;
 }
+
+button {
+    background-color: #bbb;
+    font-size: 16px;
+    padding: 5px;
+    border-radius: 5px;
+    border: 2px solid #aaa;
+    margin: 15px 0;
+}
+
+button.default {
+    font-weight: bolder;
+    background-color: #ccc;
+    box-shadow: 2px 2px 0 #eee;
+}
+
+.pull-right {
+    float: right;
+}
+
+#drawer button {
+    margin: 15px;
+}
+
--- a/static/css/drawer.css
+++ b/static/css/drawer.css
@@ -33,6 +33,10 @@
     color: white;
 }
 
+input[type=number]:invalid {
+    background-color: #FFCCCC;
+}
+
 #drawer .actions {
     margin: 10px 2px;
 }
--- a/static/css/drawer.css
+++ b/static/css/drawer.css
@@ -1,3 +1,13 @@
+#drawer button, #treatmentDrawer button{
+    font-size: 18px;
+    font-weight: bolder;
+    background-color: #ccc;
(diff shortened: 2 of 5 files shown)


jasoncalabrese:
@jimsiff I'm hoping this takes care of most of the little bugs with the title and bg status

jasoncalabrese:
Basic validation 
screen shot 2015-04-03 at 10 45 16 pm

jasoncalabrese:
Easy way to reset to default setting from server 
screen shot 2015-04-03 at 10 55 47 pm

jimsiff:
:+1: It all tests fine for me.

oamyoamy:
1) This may already be documented... seeing issue (previously when testing dev and now when testing wip/focus-info) with refresh of graph for the new multi-hr views when switching to landscape on mobile (both Safari and Puffin on iOS). Refresh solves it, but graph doesn't redraw properly upon rotation without a manual refresh. 

2) Request... Any way to have careportal entry time either not default to "now" or else dynamically shift to "other" when a change is made in the time box?

jasoncalabrese:
@oamyoamy I haven't been able to reproduce the redraw, tested it on an iphone5 in landscape, but I've never used puffin, thought it used the ios webkit renderer though

Care portal entries should work exactly as you suggest, default is now, click the time should select other.  This pr was merged to dev so update your dev branch and try again.  If your still seeing issues I can try your site or see if anyone else can reproduce.

oamyoamy:
On a fresh load of the site, that seems true for Careportal. 
But try this:
1) load site, open Careportal, select a type, and click in the time box --> time switches to other
2) click the + to close Careportal 
3) open Careportal again, select a type, and click in the time box --> time does not switch to other

Same behavior seems true after you make a Careportal entry the first time (rather than backing out of one). For subsequent entry, the now/other does not seem to toggle when you make a time change. You have to manually flip the radio button. (If you need to see my site, let me know.) 

On reload of page it works fine. It is for consecutive use of the Careportal window (site left open) that I am not seeing the toggle.

Updated Dev just now. Seeing this both on my wip/focus-info site and my dev site.

ELUTE:
I have the same issue with the landscape view. 

IPhone 5s -safari

oamyoamy:
I have photos for the landscape issue @jasoncalabrese. I can post them here. Or would you rather I send them to you via email? I took a series this morning when I checked to see if this was still an issue on mobile.

oamyoamy:
This is one (limited) sample from Safari on iPhone 5. 

dev040415-landscape

oamyoamy:
Limited sample. Safari on iPhone 5.
dev040415-landscape

oamyoamy:
Limited sample. Safari on iPhone 5.
dev040415-landscape

hackingtype1:
Jason, 
I believe this is an issue where the clipping path in the main chart area is not updated when rotated. It appears to stay at the same x location when rotation occurs, causing all circles to the right x to be masked.

-John

jasoncalabrese:
That could be it John, I see that clip is only set in init, going to try moving that, new branch comming up...

jasoncalabrese:
Moving the clip out of init creates a similar problem as the screenshoots above for me, maybe the width/height isn't being calculated right?  Also the clip does get translated if not in init. There was an existing issue with this before so I don't think anything new was introduced by this PR.

oamyoamy:
I noticed it when I started testing dev that introduced the new "hour" views (3/6/12/24). We just need to be prepared to explain to users to reload the page, if necessary.

jasoncalabrese:
I have seen the clipping with iOS in landscape before, we should be able to find a fix.  I think the width we get using the code below is wrong on iOS/safari sometimes.  Maybe the rotate animation hasn't completed?

@hackingtype1 do use something different in your version?

```
var chartWidth = (document.getElementById('chartContainer')
            .getBoundingClientRect().width) - padding.left - padding.right;
```

jimsiff:
I've seen periodic refresh issues similar to @oamyoamy since Brownie or so.  It usually happens when going into landscape mode.  I'm not sure it's necessarily a new issue, but might be worth tracking.

There is also another IOS related issue that would be food to address.  The tool tips don't go away unless you refresh the window.  You can tap another tooltip and it will switch but it would be nice if taps into open space could clear the active tool tip.  

Maybe we can start tracking all of these in that new branch for a point release for E if it's not fixed quickly.

PR #554, opened 2015-04-10
Client side alarm for stale data

jasoncalabrese:
If we find issues with this we shouldn't try squeezing it into the first Enchilada release, but I trust this more that my current network connection...


jasoncalabrese:
BG has the line-through text decoration, time ago pill border and background change while the alarm is active
screen shot 2015-04-15 at 11 31 13 pm

After alarm is snoozed, only the # of mins will be yellow/red
screen shot 2015-04-15 at 11 31 44 pm

Client side overrides for time ago warn/urgent mins
screen shot 2015-04-15 at 11 19 11 pm

PR #556, opened 2015-04-12
Wip/garmin watch

beached:
This small change will allow easy support from the Garmin watches that use the Connect IQ SDK.  Their date manipulation is almost non-existent so it pre-calculates the number of minutes since the last sensor data.  Also, it only sends what will be displayed to reduce network and battery usage.

### DIFF
--- a/lib/garmin.js
+++ b/lib/garmin.js
@@ -1,0 +1,109 @@
+var DIRECTIONS = {
+  NONE: 0
+, DoubleUp: 1
+, SingleUp: 2
+, FortyFiveUp: 3
+, Flat: 4
+, FortyFiveDown: 5
+, SingleDown: 6
+, DoubleDown: 7
+, 'NOT COMPUTABLE': 8
+, 'RATE OUT OF RANGE': 9
+};
+
+function directionToTrend (direction) {
+  var trend = 8;
+  if (direction in DIRECTIONS) {
+    trend = DIRECTIONS[direction];
+  }
+  return trend;
+}
+
+function garmin (req, res) {
+  var ONE_DAY = 24 * 60 * 60 * 1000;
+  var uploaderBattery;
+
+  function requestMetric() {
+      var units = req.query.units;
+      if (units == "mmol") {
+        return true;
+      }
+      return false;
+  }
+  var useMetricBg = requestMetric();
+
+  function scaleBg(bg) {
+      if (useMetricBg) {
+          return (Math.round((bg / 18) * 10) / 10).toFixed(1);
+      } else
+          return bg;
+  }
+
+  function get_latest (err, results) {
+    var now = Date.now();
+    var sgvData = [ ];
+    var calData = [ ];
+
+    results.forEach(function(element, index, array) {
+        if (element) {
+            var obj = {};
+            if (element.sgv) {
+                var next = null;
+                var sgvs = results.filter(function(d) {
+                    return !!d.sgv;
+                });
+                if (index + 1 < sgvs.length) {
+                    next = sgvs[index + 1];
+                }
+                obj.sgv = scaleBg(element.sgv);
+                obj.bgdelta = (next ? (scaleBg(element.sgv) - scaleBg(next.sgv) ) : 0);
+		obj.timedelta = Number((Date.now( ) - element.date)/60000.0).toFixed( 0 );
+                if (useMetricBg) {
+                    obj.bgdelta = obj.bgdelta.toFixed(1);
+                }
+                if ('direction' in element) {
+                    obj.trend = directionToTrend(element.direction);
+                }
+                sgvData.push(obj);
+            } else if (req.rawbg && element.type == 'cal') {
+                calData.push(element);
+            }
+        }
+    });
+
+    var count = parseInt(req.query.count) || 1;
+
+    var bgs = sgvData.slice(0, count);
+    //for compatibility we're keeping battery here, but it would be better somewhere else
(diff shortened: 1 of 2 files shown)


jasoncalabrese:
I don't think we should duplicate everything in pebble.js

beached:
I'm looking at the cgm-pebble and this and I think the output can be minimized and retain the same functionality.  I need the timedelta as building time stamps from components in connect IQ isn't obvious to me.  What if I cleaned up the pebble.js.

I am looking at cgm-pebble and it is recalculating the trend value that has been passed from the server.  I will double check but what if the output was like 
{"now":1429024463346,"sgv":"92","bgdelta":0,"trend":4,"datetime":1429014776000,"battery":"23", "timedelta", 3},"cals":[]}

In the pebble js it is hard wired to bgs[0] so the array may not be necessary.

YYGIRL:
There are some things in pebble.js that are there for backwards
compatibility and historical reasons, and you should only take what you
need for data savings / pulling purposes. That being said, if you need
certain data, I would duplicate the format of pebble.js if at all possible
so that some kind of consistency can be maintained across wearables.

On Tue, Apr 14, 2015 at 11:20 AM, Darrell Wright <notifications@github.com>
wrote:

> I'm looking at the cgm-pebble and this and I think the output can be
> minimized and retain the same functionality. I need the timedelta as
> building time stamps from components in connect IQ isn't obvious to me.
> What if I cleaned up the pebble.js.
>
> I am looking at cgm-pebble and it is recalculating the trend value that
> has been passed from the server. I will double check but what if the output
> was like
> {"now":1429024463346,"sgv":"92","bgdelta":0,"trend":4,"datetime":1429014776000,"battery":"23",
> "timedelta", 3},"cals":[]}
>
> In the pebble js it is hard wired to bgs[0] so the array may not be
> necessary.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nightscout/cgm-remote-monitor/pull/556#issuecomment-92904042>
> .
>



-- 
*Join the Jack Attack!*
*Fight Type 1 Diabetes*
*Give to the JDRF - **www.jdrf.org* <http://www.jdrf.org>

beached:
I am willing to do some work here, but I don't want to step on toes.  How is the best way to approach this then?  I realize my patch is pebble.js copied to garmin.js and modified to reduce and produce what I needed.  I can see a few approaches, but I am ignorant of much of the history/backwards comparability.

YYGIRL:
My point being - you don't need the backwards compatibility. Just take what
you need, so you don't need everything in pebble.js.

On Tue, Apr 14, 2015 at 1:18 PM, Darrell Wright <notifications@github.com>
wrote:

> I am willing to do some work here, but I don't want to step on toes. How
> is the best way to approach this then? I realize my patch is pebble.js
> copied to garmin.js and modified to reduce and produce what I needed. I can
> see a few approaches, but I am ignorant of much of the history/backwards
> comparability.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nightscout/cgm-remote-monitor/pull/556#issuecomment-92985886>
> .
>



-- 
*Join the Jack Attack!*
*Fight Type 1 Diabetes*
*Give to the JDRF - **www.jdrf.org* <http://www.jdrf.org>

jasoncalabrese:
Ideally we'd create a whole new endpoint that bundles data like `pebble.js`, but based on what we've learned since the original pebble endpoint was created.  We could call it something like `\summary`.

Then have the next version of the pebble watchface/app use the new summary endpoint. It would do most (if not all) the processing and calculating on the server so it's easy to just display the data.  Then add more options to allow each device to ask for what it wants.  Could even make /pebble work as an alias for the new endpoint passing any options to preserve backwards compatibility.

Maybe thats too big?

beached:
That sounds good.  This will be utilized more and more as wearable devices come to market and being too specific will make for more work in the long run.  Tonight I'll update my patch

jasoncalabrese:
Sounds great, I'd much rather maintain something that can be used to support more devices.  Was worried about having dedicated code for every new device.  It might be worth trying to align with the protobuff download schema for areas that overlap. See: https://github.com/nightscout/android-uploader/blob/dev/core/src/main/java/com/nightscout/core/model/Download.proto

YYGIRL:
There are pros and cons to both. I like having the flexibility of
displaying what is needed on each device, and one size does not fit (and
will never) fit all. I would rather get straight values than processing and
calculations done on the backend, because the needs are so varied and
different. Plus it's much easier to change and deploy code on a wearable
than it is on cgm-remote-monitor. One final note - we don't always get data
from cgm-remote monitor - and can get data from many other sources -
camping mode in a mobile application, possibly straight from a receiver,
etc. Just getting straight values reduces the need for wearables to always
depend on cgm-remote-monitor to always do processing and calculations.

On Tue, Apr 14, 2015 at 1:55 PM, Jason Calabrese <notifications@github.com>
wrote:

> Sounds great, I'd much rather maintain something that can be used to
> support more devices. Was worried about having dedicated code for every new
> device. It might be worth trying to align with the protobuff download
> schema for areas that overlap. See:
> https://github.com/nightscout/android-uploader/blob/dev/core/src/main/java/com/nightscout/core/model/Download.proto
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nightscout/cgm-remote-monitor/pull/556#issuecomment-92997450>
> .
>



-- 
*Join the Jack Attack!*
*Fight Type 1 Diabetes*
*Give to the JDRF - **www.jdrf.org* <http://www.jdrf.org>

PR #607, opened 2015-05-24
Settings to show plugin visualisations in client settings

jasoncalabrese:
Also lots of refactoring


jasoncalabrese:
Client settings not functional yet

jasoncalabrese:
@bewest if you get a chance take a look, trying to get the base plugins figured out.  Starting with adding extra pills, but eventually want everything to be a plugin.  Even allow external plugins to be required.

PR #609, opened 2015-05-25
Wip/plugins

sulkaharo:
IOB formatting to closer match a pump display, mmol value display with expected number of digits

### DIFF
--- a/lib/plugins/cannulaage.js
+++ b/lib/plugins/cannulaage.js
@@ -16,7 +16,7 @@ function init() {
     var age = 0;
     var found = false;
     var treatmentDate = null;
-    var message = 'no notes';
+    var message = '';
 
     for (var t in this.env.treatments) {
       if (this.env.treatments.hasOwnProperty(t)) {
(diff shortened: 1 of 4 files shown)


lib/plugins/cannulaage.js
```diff
@@ -16,6 +16,7 @@ function init() {
     var age = 0;
     var found = false;
     var treatmentDate = null;
+    var message = 'no notes';
```
jasoncalabrese: maybe instead of showing "no notes" we can just push 1 more label/value to the info array if there is a note

lib/plugins/pluginbase.js
```diff
@@ -50,6 +50,30 @@ function updatePillText(updatedText, label, info, major) {
   }
 }
 
+function roundInsulinForDisplayFormat(iob) {
```
jasoncalabrese: since this is pump specific we should make it an option, been thinking of how we can pass options to the plugins.  Maybe something like `IOB_ROUNDING="medtronic"`?

PR #612, opened 2015-05-29
initial tests for cob

jasoncalabrese:
After seeing some strange results for COB I created these simple tests, I think the source of the issue is the way delayed carbs are added back.

@scottleibrand, hoping can help make sense of this, maybe we should simply the COB logic?

See:
https://github.com/nightscout/cgm-remote-monitor/blob/wip/test-cob/lib/plugins/cob.js#L113

Ported from:
https://github.com/nightscout/cgm-remote-monitor/blob/wip/iob-cob/static/js/client.js#L1525

To test check out the branch, `npm install`; `./node_modules/mocha/bin/_mocha tests/cob.test.js`

### DIFF
--- a/tests/cob.test.js
+++ b/tests/cob.test.js
@@ -1,0 +1,69 @@
+'use strict';
+
+var should = require('should');
+
+describe('COB', function ( ) {
+  var cob = require('../lib/plugins/cob')();
+
+  cob.profile = {
+    sens: 95
+    , carbratio: 18
+    , carbs_hr: 30
+  };
+
+  it('should calculate IOB, multiple treatments', function() {
+
+    var treatments = [
+      {
+        "carbs": "100",
+        "created_at": new Date("2015-05-29T02:03:48.827Z")
+      },
+      {
+        "carbs": "10",
+        "created_at": new Date("2015-05-29T03:45:10.670Z")
+      }
+    ];
+
+    var after100 = cob.cobTotal(treatments, new Date("2015-05-29T02:03:49.827Z"));
+    var before10 = cob.cobTotal(treatments, new Date("2015-05-29T03:45:10.670Z"));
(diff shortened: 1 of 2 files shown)


tests/cob.test.js
```diff
@@ -0,0 +1,69 @@
...
+    var before10 = cob.cobTotal(treatments, new Date("2015-05-29T03:45:10.670Z"));
+    var after10 = cob.cobTotal(treatments, new Date("2015-05-29T03:45:11.670Z"));
+
+    console.info('>>>>after100:', after100);
+    console.info('>>>>before10:', before10);
+    console.info('>>>>after2nd:', after10);
+
+    after100.cob.should.equal(100);
+    Math.round(before10.cob).should.equal(59);
+    Math.round(after10.cob).should.equal(69); //WTF == 128
```
jasoncalabrese: How can 110 carbs create a COB of 128?

jasoncalabrese:
@sulkaharo have you seen anything like this?

MilosKozak:
I found a bug in cob too. treatments in the future (while calculation to the past) are added to cob too. i fixed it in my branch boluscalc

jasoncalabrese:
that might be a different bug that I saw with this. Can you point me to a commit with your fix so I can port it over.  Hoping to wrap up most of my refactoring/clean up in a week or 2.

MilosKozak:
https://github.com/MilosKozak/cgm-remote-monitor/blob/boluscalc/lib/cob.js#L5 focus on skipnewer param. i was not sure if it was intention so i better added a param

MilosKozak:
btw to anylyze cob try my reports with "cob on". there is very nice visible result of calculation. i'm sure it's calculating wrong when you have more carbs treatments in a short period of time (<15min)

PR #615, opened 2015-05-31
improve client perf

jasoncalabrese:
cache displayBG for treatments
use _.debouce to avoid calling expensive functions too often

@sulkaharo this seems to have improved things for me, will test with my old nexus7 later


sulkaharo:
Pulled, seems good. Made a further PR against this branch.

PR #617, opened 2015-05-31
querySelectorAll optimization + don't render invisible treatments

sulkaharo:
Optimization for querySelectorAll, which seems to be non-performant and used by D3 a _lot_.

One-liner to disable drawing of treatments not visible on screen.


jasoncalabrese:
Nice finds, made a big difference

PR #631, opened 2015-06-12
Wip/docker

Fokko:
I have created scripts which enable the host Nightscout onto a Docker-ecosystem. Also a Docker-Compose script has been provided which enables to deploy Nightscout+MongoDB+Mosquitto onto Docker.


jasoncalabrese:
Since this is based on `wip/push-notify` it will need to wait on that to get to dev. That will also make it a lot easier to see what the real diff is.  It would be better to have just the docker commit in a PR, looks like there was some other minor changes here and there, but don't think they were related.

Fokko:
Yes, my mistake. I will make a new PR with only the Docker part.

PR #632, opened 2015-06-12
Wip/docker compose

Fokko:
I have created scripts which enable the host Nightscout onto a Docker-ecosystem. Also a Docker-Compose script has been provided which enables to deploy Nightscout+MongoDB+Mosquitto onto Docker.

docker


docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+nightscout:
+  build: .
+  links:
+   - database
+   - broker
+  ports:
+   - "1337:1337"
+  environment:
+   - MONGO_CONNECTION=mongodb://database/nightscout
+   - API_SECRET=mylittlesecret
```
jasoncalabrese: How could we make this configurable?

jasoncalabrese:
Thanks @Fokko, I haven't done anything with docker yet, so I'll use this as my excuse.  Added 1 line comment.

docker-compose.yml
stavlor: Very Carefully? heh there are probablt a few ways i suppose we could make something to make a new docker-compose.yml based on input from a form or something not sure of the best way to do that.

Fokko: Yes, that should not be a problem. The user should adjust it's environment variables such as the secret and the Pebble keys. The only visible ports from the outside are the exposed ports, in the current case the HTTP port of Node and the MQTT port.

Docker uses containers to isolate the application from the big-bad outside world. For example the MongoDB container is referred by `database` and is linked to the Node container. It does not expose any ports to the outside world, so it cannot be reached from outside of the docker ecosystem.

The environment variables shown in the .yml are passed on to the container as environment variables. I will dive in to see if there is any way to make these easy configurable.

Fokko:
I have just created a new VM at Digital-ocean and booted Nightscout, MongoDB and Mosquitto using Docker. http://ns.driesprong.frl:1337/

Fokko:
I have added some extra tests for the database driver and it is working now properly. The tests have been extended and the database will be initialized synchronous which removes the problems with the double initialization as described in the other [pull request](https://github.com/nightscout/cgm-remote-monitor/pull/635). Also, a Docker container can easily be build from the source.

Next, I will provide documentation and next would be to nice add the hooks to Github so a new version of the container will be deployed as a new commit has been done. But this will be described in the upcoming documentation.

docker-compose.yml
jasoncalabrese: I don't think we should have a default api secret, would make it too easy for people to use it

env.js
```diff
@@ -169,14 +169,6 @@ function config ( ) {
   env.pushover_api_token = readENV('PUSHOVER_API_TOKEN');
   env.pushover_user_key = readENV('PUSHOVER_USER_KEY') || readENV('PUSHOVER_GROUP_KEY');
 
-  // TODO: clean up a bit
```
jasoncalabrese: We'd like to keep everything that reads from env vars in `env.js`, then for testing it's easy to create a mocked env

lib/api/entries/index.js
```diff
@@ -223,10 +223,12 @@ function configure (app, wares, core) {
     }, insert_entries, format_entries);
 
     if (app.enabled('api')) {
+
       // Create and store new sgv entries
       api.post('/entries/', wares.verifyAuthorization, function (req, res, next) {
-        req.persist_entries = true;
-        next( );
+          console.log(req)
```
jasoncalabrese: remove logging of every request

lib/bootevent.js
```diff
@@ -1,38 +0,0 @@
-
```
jasoncalabrese: why did this get removed?  bootevent is the core of the app now

lib/entries.js
```diff
@@ -101,14 +101,21 @@ function storage(name, storage, pushover) {
     // TODO: implement
   }
 
-  function remove (fn) {
-    // TODO: implement
+  function remove (query, next) {
```
jasoncalabrese: need auth for remove

testing/populate_rest.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+///////////////////////////////////////////////////
+// This script is intended to be run as a cron job
+// every n-minutes or whatever the equiv is on windows
+//
+// Author: John A. [euclidjda](https://github.com/euclidjda)
```
jasoncalabrese: we don't have author name on any files, shouldn't start now, maybe port and mention source in commit message?

testing/util.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+///////////////////////////////////////////////////
+// This script is intended to be run as a cron job
+// every n-minutes or whatever the equiv is on windows
+//
+// Author: John A. [euclidjda](https://github.com/euclidjda)
```
jasoncalabrese: no author names

server.js
```diff
@@ -1,67 +1,81 @@
...
+    return transport.createServer(app);
 }
 
-var bootevent = require('./lib/bootevent');
-bootevent(env).boot(function booted (ctx) {
-    env.store = ctx.store;
-    var app = require('./app')(env, ctx);
-    var server = create(app).listen(PORT);
-    console.log('listening', PORT);
+// We need the database
```
jasoncalabrese: what happened here?  merge gone bad? this is undoing some recent refactoring

docker-compose.yml
Fokko: Good point Jason, I have commented the line out, so people have to explicitly set the secret to enable the API.

lib/entries.js
Fokko: This is not implemented in the API but just for the tests to clean up afterwards.

lib/bootevent.js
Fokko: Because it is not necessary anymore at this moment. As we see in the code, we observe two acquire blocks. Which are processed asynchronously as we read from the (documentation)[https://www.npmjs.com/package/bootevent]. But it is obvious that the second block is dependent on the first block as we use the `ctx.store` in the second block, which is initialized in the first block. So we have to wait for second block to execute as the first block is finished (and the `ctx.store` variable is initialized.

The problem is observable in the build logs as the initialization method of the [Singleton-like](http://stackoverflow.com/questions/137975/what-is-so-bad-about-singletons) pattern, as we see four times the `Connecting to MongoDB` [line](https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/67610160#L437). This will run just right as the variable within the store class just gets overwritten a couple of times, but it is smelly in terms of code quality.
We do not observe this in the [build-log](https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/67378958#L426) of pull-request.

When we wait for the store-callback, we reduce the bootevent to a single acquire block, which makes the boot-event a bit redundant. Therefore I decided to remove it.

lib/api/entries/index.js
Fokko: Good point.

env.js
Fokko: I agree, but I removed this line as it overwrites the environment variables with custom variables as defined in the `database_configuration.json`. This is confusing when you assume that you are connecting to the Mongo-database by the credentials in the environment, when you in fact are using the credentials from the file.

testing/populate_rest.js
Fokko: It was already in the `populate.js` file. I spit some functions to the `util.js`, and therefore I copied the author names along. I didn't want to be the person who removes the names of the original author, but I agree that in an open-source project author names are not appropriate in the source itself.

testing/util.js
Fokko: It was already in the `populate.js` file. I spit some functions to the `util.js`, and therefore I copied the author names along. I didn't want to be the person who removes the names of the original author, but I agree that in an open-source project author names are not appropriate in the source itself.

Fokko:
The [Docker container is building](https://registry.hub.docker.com/u/fokkodriesprong/cgm-remote-monitor/builds_history/229380/). If this becomes accepted it would be nice to provide proven Docker images through the name of nightscout (instead of fokkodriesprong). I am [working on the wiki](https://github.com/nightscout/cgm-remote-monitor/wiki/Docker), but this is still in bèta stage.

Fokko:
I will revise my PR and provide a new one.

PR #635, opened 2015-06-14
Feature/mongodb double init

Fokko:
Simplification of the MongoDB connection-pool. Please take a look at:
http://stackoverflow.com/questions/10656574/how-to-manage-mongodb-connections-in-a-nodejs-webapp

### DIFF
--- a/database_configuration.json
+++ b/database_configuration.json
@@ -1,4 +1,5 @@
 {
     "url": "mongodb://root:root@localhost/nightscout",
-    "collection": "entries"
+    "collection": "entries",
+    "API_SECRET": "thisismysercret"
 }
--- a/database_configuration.json
+++ b/database_configuration.json
@@ -1,5 +1,0 @@
-{
-    "url": "mongodb://root:root@localhost/nightscout",
-    "collection": "entries",
-    "API_SECRET": "thisismysercret"
-}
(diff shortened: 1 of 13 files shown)


jasoncalabrese:
So far we've been careful to only read from env var in env.js, I don't think we want to move that to `storage.js`

Fokko:
I will provide a new commit which removes the bug. At this moment I have isolated the problem. The problem is in the [process of booting](https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/bootevent.js).

I am fairly new to Node, but what I understand from the [plugin](https://github.com/bewest/bootevent) is that the acquire function are executed asynchronously. As not in parallel as NodeJS only runs on [one thread](https://nodejs.org/about/). But this will invoke the `maybe_connect` method multiple times, and multiple connections will be made as the callback which sets the `mongo.db` variable is executed asynchronously.

Therefore I would suggest to wait until the MongoDB connection is initialized and then apply the indexes and create the collections.

Fokko:
@jasoncalabrese I agree, reading environment variables must be limited to the `env.js`.

Fokko:
Already fixed in [#632](https://github.com/nightscout/cgm-remote-monitor/pull/632).

PR #628, opened 2015-06-10
Notification plugins, unify pushover/websocket alarm, everything is becoming a plugin

jasoncalabrese:

PR #642, opened 2015-06-20
Wip/duplicate mongo connection callbacks

Fokko:
Yes, I have reduced the number of changes of Mongo. It would be nice to have something which reconnects when MongoDB is not available or so as provided in [Mongoose](http://bites.goodeggs.com/posts/reconnecting-to-mongodb-when-mongoose-connect-fails-at-startup/). And it will throw an exception after a few unsuccessful retries. This would be a very nice feature from the perspective from Docker as the different containers are created simultaneous.

The `Store` class which handles the connection is at the moment quite bloated from my perspective. But I understand that this is delicate as it is the foundation of the application. Therefore I reduced the proposed changes to a minimum.

I removed the `Bootevent`, but I changed the structure just a little so it works quite similar as with the Bootevent. As we see in the Bootevent-code, we observe two acquire blocks. Which are processed asynchronously as we read from the documentation. But it is obvious that the second block is dependent on the first block as we use the `ctx.store` in the second block, which is initialized in the first block. So we have to wait for second block to execute as the first block is finished, and the `ctx.store` variable is initialized. 

I am fairly new to Node, but what I understand from the [plugin](https://github.com/bewest/bootevent) is that the acquire function are executed asynchronously. As not in parallel as NodeJS only runs on [one thread](https://nodejs.org/about/). But this will invoke the `maybe_connect` method multiple times, until the callback has been invoked which sets the `mongo.db` variable.

We see multiple times the `Connecting to MongoDB` [line](https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/67610160#L437). This means that the asynchronous call is invoked multiple times before the local variable is set. This will run just right as the variable within the store class just gets overwritten a couple of times. Therefore I would suggest to wait until the MongoDB connection is initialized (As I changed in bootevent) and then apply the indexes and create the collections etc.

When we wait for the store-callback, we reduce the bootevent to a single acquire block, which makes the boot-event a bit redundant. Therefore I decided to remove it. Leaving the Bootevent in would only provide some syntactic sugar.

Furthermore, I have changed some of the `entries` tests which first checks if it is an array. Otherwise it will try to assert the `.length` property which is not available on an object or null value. Now the errors will be nicer if the tests fail.


bewest:
Nice work!  In bootevent, startup, I prefer the `boot.acquire(phase)` used here:

https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/bootevent.js#L7-L35

The thought here was:
```javascript

function setupMongoPhase ( ) {
// ...
// next( )
}

function setupInternals ( ) {
// ...
// next( )
}

function booter (env) {
  var boot = bootevent( )
    .acquire(setupMongo)
    .acquire(setupInternals)
    ;
}
```

This makes it clear that the phases must run in order, and it's easy to add/remove different phases (eg as we add plugins).

I like a lot of your changes, but noticed that all the logic is now embedded inside the "mongo phase."  What do you think about restoring the `.acquire(phase)` idiom, so that it's clear the internal stuff happens in a phase after mongo is set up?  We think this will scale better as we consider adding more phases and services.

Fokko:
Hi @bewest,

Thank you for the comments. Based on @jasoncalabrese the Bootevent is back into the code. Looks much better now! Personally I would also integrate the starting-event of the MQTT listner. But thats not a priority right now.

Cheers, Fokko

lib/bootevent.js
```diff
@@ -1,13 +1,11 @@
...
 
-var bootevent = require('bootevent');
-
 function boot (env) {
 
   function setupMongo (ctx, next) {
     var store = require('./storage')(env);
     // initialize db connections
-    store( function ready ( ) {
+    store( function ready ( err, store ) {
```
jasoncalabrese: is this a different `store` instance?

jasoncalabrese:
Just gave this a try, still seeing the double connection

>
  Setting up new connection to MongoDB
  Setting up new connection to MongoDB
  GIT HEAD 03a98c1
  Successfully established a connected to mongo
  storage system ready
  Ensuring indexes

Fokko:
[Works now!](https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/67797964#L435-L443) The callback method is now not yet used. I would like to do some more cleaning on the store class. But I want to keep the PR's small.

jasoncalabrese:
Wonder why the git commit id is printing multiple times, when I tested it only printed once.

Fokko:
This is also (the case)[https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/67772175#L440-L446] in #645, which is essential the current `dev`. Locally, when I run it a couple of times, I sometimes get a single line, sometimes multiple a couple of times. It is obvious that in the current `dev` this issue exists, but it does not explicitly involve this PR as far as I can see.

Fokko:
@jasoncalabrese I looked at the multiple git commit it. This only occurs when running `npm test`. There are many `before` hooks within all the test which contain the `var env = require('../env')( );`. When all the test-classes are initialized the git commit id is printed multiple times. This is only in the `dev` because in the #628 PR introduced a lot of tests. It does not introduce any bugs or so.

jasoncalabrese:
oh, that makes sense, and explains why I didn't see it when testing locally

bewest:
Nice. :-)

Fokko:
Thanks!

PR #645, opened 2015-06-21
Initial style guide and dev tips in CONTRIBUTING.md

jasoncalabrese:
Some initial ideas, open to any other ideas.  Just want to get something basic in place.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,23 @@ design.
 We develop on the `dev` branch.
 You can get the dev branch checked out using `git checkout dev`.
 
+## Style Guide
+
+Some simple rules, that will make it easier to maintain our codebase:
+
+* All indenting should use 2 space where possible (js, css, html, etc)
+* A space before function parameters, such as: `function boom (name, callback) { }`, this makes searching for calls easier
+* Name your callback functions, such as `boom('the name', function afterBoom ( result ) { }`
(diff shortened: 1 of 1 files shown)


bewest:
:+1:

Fokko:
Perfect, I prefer the tab's instead of two spaces. But this is a matter of taste, I will configure my IDE. 

Any idea's on the single line statements:
``` js
if(err)
  throw new Error("MongoDB connection string is missing");
// or
if(err) {
  throw new Error("MongoDB connection string is missing");
}
```
I don't mind, the second is more verbose than the first one.

jasoncalabrese:
I prefer `if (` vs `if(`, but wrapping seems dependent on the length of the line.  With languages like Scala I avoid the extra `{}`s, but trend to add them with js.  Don't think it's a big deal either way.

sulkaharo:
However we do this, I'd love simple instructions how to set IntelliJ to auto-format using the conventions. Ideally an importable settings file... ;)

jasoncalabrese:
I looked at JSCS a littlehttps://github.com/jscs-dev/node-jscs 

I don't want to have to worry about every detail, so if there a button or script that can make everything look nice that would be nice.


More about jscs here:

https://medium.com/@addyosmani/auto-formatting-javascript-code-style-fe0f98a923b8

jasoncalabrese:
Opened #648 to get JSCS configured, there's a plugin for intellij and many other editors that could be used

PR #647, opened 2015-06-22
Wip/profileobject

sulkaharo:
New profile object, which support the old data format and new time of day based values for all contained data structures, including basal. Converted existing profile code to use the profile object. Added basal plugin, which shows current basal rate and other profile information. Added temporary basal adjustment calculation to BWP, if basal profile is present. Fixed a significant bug in how COB called IOB (passed a date to IOB instead of profile). Test for profile code.

### DIFF
--- a/lib/plugins/basalprofile.js
+++ b/lib/plugins/basalprofile.js
@@ -1,0 +1,52 @@
+'use strict';
+
+var _ = require('lodash');
+
+function init() {
+
+  function basal() {
+    return basal;
+  }
+
+  basal.label = 'Basal Profile';
+  basal.pluginType = 'pill-minor';
+
+  function hasRequiredInfo (sbx) {
+
+    if (!sbx.data.profile) return false;
+    
+    if (!sbx.data.profile.hasData()) {
+      console.warn('For the Basal plugin to function you need a treatment profile');
+      return false;
+    }
+    
+    if (!sbx.data.profile.getBasal()) {
+      console.warn('For the Basal plugin to function you need a basal profile');
+      return false;    
+    }
+    
+    return true;
+ }
+
+  basal.updateVisualisation = function updateVisualisation (sbx) {
+  
+    if (!hasRequiredInfo(sbx)) {
+      return;
+    }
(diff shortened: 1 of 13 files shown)


sulkaharo:
TODO: enable the code to load multiple profiles from the database and use the relevant profile based on when it's been taken into use. This can be implemented without a change in the API for the profile code access.

sulkaharo:
FIXING TESTS!

jasoncalabrese:
really cool, agree that the pebble endpoint needs work, it should feed off the same data and not do new mongo queries

lib/plugins/basalprofile.js
```diff
@@ -0,0 +1,52 @@
...
+
+  basal.updateVisualisation = function updateVisualisation (sbx) {
+  
+    if (!hasRequiredInfo(sbx)) {
+      return;
+    }
+    
+    var basalValue = sbx.data.profile.getBasal(sbx.time);
+    
+    var info = [{label: 'Current basal:', value: basalValue + ' IU'}
```
jasoncalabrese: IU or U?  we use U everywhere else

lib/plugins/boluswizardpreview.js
```diff
@@ -100,6 +103,19 @@ function init() {
       , {label: 'Expected effect', value: '-' + results.effectDisplay + ' ' + sbx.units}
       , {label: 'Expected outcome', value: results.outcomeDisplay + ' ' + sbx.units}
     ];
+    
+    if (results.tempBasalAdjustment) {
+      if (results.tempBasalAdjustment.thirtymin > 0) {
+        info.push( {label: '30m temp basal', value: results.tempBasalAdjustment.thirtymin + '%'});
```
jasoncalabrese: for some pumps like our animas ping, temp basals are set with +/-%, so 50% = -50%, 110% = +10% and can only be set in 10% increments, so I'd need to add some animas specific display.  Similar to the medtronic specific rounding.  Wonder where that should go.  Guess we'll figure it out later.

Issue #657, opened 2015-06-26
Breakage due to automatic MongoLab mongo version upgrades


dduugg:
My nightscout server broke abruptly when MongoLab upgraded the mongo version of my db to 3.0.x. The response I received from their support indicates they are planning to automatically, irreversibly upgrade all sandbox accounts in the coming weeks:

>Hello,

>Sorry for the delayed reply. Unfortunately, there is not a way to roll back the version of an Experimental Sandbox database.

>For a guide on migrating data between the two databases, please refer to the "Migrating Data into MongoLab" section of our documentation, linked below: http://docs.mongolab.com/migrating/#dump-and-restore

>Note that you must dump the data using 3.0.X version of mongodump and restore using a 2.6.x version of mongorestore.

>Also, please keep in mind that our normal Sandboxes will be upgraded to MongoDB 3.0 in the coming weeks, but that we will be informing users to this change prior to the update.

>Regards,

>Samson

jasoncalabrese:
What was the actual error, was data lost?  Any ideas for a way we could prevent problems for others?

dduugg:
No data lost, just can't connect to MongoLab. There are others on SO having what I believe is the same problem: 

http://stackoverflow.com/q/30659213/2193416
http://stackoverflow.com/q/30924859/2193416

To fix, it may be sufficient to upgrade the mongodb package to v1.4.29: http://docs.mongodb.org/manual/release-notes/3.0-scram/#upgrade-drivers

jasoncalabrese:
the version on dev is current, if that works we should be able to to a hot fix

dduugg:
Just opened https://github.com/nightscout/cgm-remote-monitor/pull/658. Haven't tested on dev, but I committed directly to master on my fork, and it seems to be up and running with the mongo 3 collection. (The page errored out for the first minute or so, something about a type error in calling `.extent` on undefined, but it should be unrelated.)

jasoncalabrese:
Guess we can close this one too, as mentioned on #658 the solution is to redeploy so the newer mongo module is picked up

PR #658, opened 2015-06-26
Upgrade mongodb to 1.4.29

dduugg:
resolves https://github.com/nightscout/cgm-remote-monitor/issues/657

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "forever": "~0.13.0",
     "git-rev": "git://github.com/bewest/git-rev.git",
     "long": "~2.2.3",
-    "mongodb": "^1.4.7",
+    "mongodb": "^1.4.29",
     "moment": "2.8.1",
     "mqtt": "~0.3.11",
     "pushover-notifications": "0.2.0",


jasoncalabrese:
Dev is at 2.0.34, but this would be a smaller change for a hot fix

dduugg:
gah, meant to target dev. i leave it to your discretion if a hotfix is merited.

jasoncalabrese:
wouldn't make sense to target dev with this, since we should be at a much more current version.  I'll ask around and see if there are other reports of this, but sounds like there will be more and that we should take care of it now.

dduugg:
right. i meant your discretion in merging or closing this PR. i'm not advocating downgrading :smile:

stavlor:
My guess is this isn't affecting dev will try a master deployment though but my dev has been running without issues on MongoLab will test in master

jasoncalabrese:
This is currently only an issue with the experimental sandboxes that were upgraded to 3.0.  We need to get a patch release out before the normal sandbox upgrade cycle starts.

You'd need to select experimental when creating the sandbox

screen shot 2015-06-26 at 2 04 30 pm

screen shot 2015-06-26 at 2 04 36 pm

blob

jasoncalabrese:
Based on some testing done by @stavlor it looks like by doing a re-deploy or a fresh deploy of master will pull in newer versions of the mongo module so we probably don't need to do a hotfix for this

dduugg:
sg! closing out.

PR #659, opened 2015-06-27
Adding Docker to the repository.

Fokko:
The `Dockerfile`-file defines how the docker-service should build the repository. The second file `.dockerignore` defines which files should be left out of the image to reduce the footprint of the image. Currently the image is build on top of Node 0.12.5 and NPM 2.11.3.

As discussed at Gitter, the rest of the files will be put in the separate repository to keep everything nice and tidy.


Fokko:
In progress. Found out that everything works fine and deploys, but the bower build is not triggered and therefore the web-interface isn't working. :+1:

Fokko:
The current Dockerfile is proper and works for me locally and on my Digital-ocean cloud provider.

Fokko:
```
    NAMES
9d83d2a3a7ea        nightscout/cgm-remote-monitor-development:latest   "node server.js"       4 hours ago         Up 3 hours          0.0.0.0:1337->1337/tcp                           nightscout-1.Nightscout.4d2a1621   
a8ed8543da0a        prologic/mosquitto:latest                          "/usr/sbin/mosquitto   4 hours ago         Up 4 hours          8883/tcp, 0.0.0.0:1883->1883/tcp                 broker-1.Nightscout.b55086d0       
0baaee1043fc        tutum/mongodb:latest                               "/run.sh"              4 hours ago         Up 4 hours          28017/tcp, 0.0.0.0:27017->27017/tcp              database-1.Nightscout.a1134afb     
```

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM node:latest
+
+MAINTAINER fokko@driesprong.frl
```
jasoncalabrese: Can we use a generic name here?  Does it need to be an email?

jasoncalabrese:
@bewest does this look ok to you? The docker install is doing `npm install` as we discussed

Dockerfile
Fokko: Sure, the maintainer field can be arbitrary. It will fill in the author field at the Docker registry. Currently I have the whole-set up running using Docker as in https://github.com/nightscout/nightscout-docker

Fokko:
@jasoncalabrese, I changed the maintainer to Nightscout. As you already pointed out, this script uses `npm` to build the image, therefore it is not likely to change the `Dockerfile` in the future.

jareware:
What's the rationale for keeping the Dockerfile in https://github.com/nightscout/nightscout-docker and not in this repository? As to the images at https://hub.docker.com/r/nightscout/cgm-remote-monitor-development/builds/, it looks like they're not really updated, and even if they were, they're building directly off of `master`. It would be superbly cool to have all **release tags** from this repository automatically built and tagged into the Docker Hub.

Fokko:
Hi Jereware, This is because it it not possible to push to Heroku with the Dockerfile present. At pushing the code to Heroku, it will fail due an error. This is obviously an error of Heroku, therefore a different repo has been created.

It would be possible to wire up everything in the different repo using hooks and stuff, but this is not an easy job to do. I would also prefer to place the Dockerfile in the main repo and wire it up to the Docker-hub.

I've updated the Docker-version, please take into account that is the latest development version. Unfortunately I don't have that much time at this moment. I just left university and started to work, which takes a lot of time/energy.

jareware:
@Fokko thanks for the answer, and that's totally OK, people have other obligations beside their OSS activities! :) Was just curious.

I'll see if I can take a look at automating this at some point.

PR #679, opened 2015-07-04
Added the option to use a linear scale on the y-axis.

Fokko:
Added the possibility to set the y-axis linear. I believe that a linear scale might be more appropriate as the glucose-level approximate reacts linear to the intake of the carbohydrates or injected insulin. 

I agree with the argument that in the 18-24 range it does not really makes a difference as it is in both cases too high. The log-scale feels a bit more scientific, as the lin-scale might feel more intuitive. (for example, if you have a linear decrease after injecting a lot of insulin, the drop will be visually at first quite slow, but the visual rate will increase when the glucose decreases).

### DIFF
--- a/static/js/client.js
+++ b/static/js/client.js
@@ -135,7 +135,7 @@ var app = {}, browserSettings = {}, browserStorage = $.localStorage;
     xScale = d3.time.scale()
       .domain(d3.extent(data, function (d) { return d.date; }));
 
-    if(app.defaults.bgScale === 'lin'){
+    if(browserSettings.bgScale === 'lin'){
       yScale = d3.scale.linear();
       yScale2 = d3.scale.linear();
     }
--- a/static/js/client.js
+++ b/static/js/client.js
@@ -135,14 +135,21 @@ var app = {}, browserSettings = {}, browserStorage = $.localStorage;
     xScale = d3.time.scale()
       .domain(d3.extent(data, function (d) { return d.date; }));
 
-    yScale = d3.scale.log()
-      .domain([scaleBg(30), scaleBg(510)]);
+    if(app.defaults.bgScale == 'lin'){
+      yScale = d3.scale.linear();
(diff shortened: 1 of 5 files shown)


static/js/client.js
```diff
@@ -135,14 +135,21 @@ var app = {}, browserSettings = {}, browserStorage = $.localStorage;
     xScale = d3.time.scale()
       .domain(d3.extent(data, function (d) { return d.date; }));
 
-    yScale = d3.scale.log()
-      .domain([scaleBg(30), scaleBg(510)]);
+    if(app.defaults.bgScale == 'lin'){
```
jasoncalabrese: Codacy Issue found: [Expected '===' and instead saw '=='.](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/768573671/issues/source?bid=2452379&fileBranchId=2457568#l138)

Fokko:
logarithmic:
scalelog

Linear:
scalelin

jasoncalabrese:
We should add an option to the settings panel to override the default, then you can toggle without messing with env vars.

Fokko:
Sure, I would suggest both, as with the time setting.

jasoncalabrese:
Yeah, right now I think all the defaults can be set with env vars and the client can override from the setting.  There's a lot of duplication the the defaults code between `env.js` and `ui-utils.js` and some quirks that we should consider cleaning up.

Fokko:
I have added an option in the menu. I did not want to do this directly, because the menu might get bloated with features which aren't toggled very often. I did not much refactoring as this is the first front-end pull-request and I am not yet familiar with the front-end code.

jasoncalabrese:
@Fokko this wouldn't be hard to do with the refactored settings and client code if you want to give it a try.  Since this is based on such an old version if would probably be easier to start over, so I'm going to close this.  Others have asked for this, so it would be a nice feature to add.

PR #684, opened 2015-07-06
don't let future data block the current

jasoncalabrese:
Noticed a problem where a treatment entered in the future prevented a notification for a current treatment.

Also updated the `sbx.lastSGV*` functions to filter future data also, we wouldn't want BWP or some other plugin doing calculation based on future data that's probably wrong.

@sulkaharo

### DIFF
--- a/lib/sandbox.js
+++ b/lib/sandbox.js
@@ -113,13 +113,19 @@ function init ( ) {
     }
   };
 
+  sbx.lastSGVEntry = function lastSGVEntry ( ) {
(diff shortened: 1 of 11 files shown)


lib/sandbox.js
```diff
@@ -113,14 +113,28 @@ function init ( ) {
     }
   };
 
+  sbx.lastEntry = function lastEntry (entries) {
```
sulkaharo: Really like this type of API calls on the sandbox, which guarantee bug free access to the data

PR #704, opened 2015-07-13
Treatment drawer update

MilosKozak:
- ability to enter date of treatment
- by holding up/down key time rolls over hours and day correctly
- treatment js code moved from utils.js to a separate file to make it plugin easier in the future
- confirm dialog shows only entered values
- fixed bug when sometimes time radio buttons don't work as expected

PR #708, opened 2015-07-16
treatmentdrawer moment

jasoncalabrese:
An extension of @MilosKozak's #704 that makes heavy use of moment to do all date parsing and manipulation.

Needs testing in other timezones


sulkaharo:
LGTM

PR #713, opened 2015-07-17
alias subscription to /downloads/$username/protobuf

bewest:
/downloads/$username/protobuf is alias for /downloads/protobuf


jasoncalabrese:
fixed test and a bug too

bewest:
Wow, very nice.

Fokko:
Looks great! Love the ISO time format.

PR #734, opened 2015-07-23
fixed width of labels+inputs in drawer

MilosKozak:


static/index.html
```diff
@@ -165,25 +165,26 @@
...
-                            <option value="Exercise" class="translate">Exercise</option>
-                            <option value="Site Change" class="translate">Pump Site Change</option>
-                            <option value="Sensor Start" class="translate">Dexcom Sensor Start</option>
-                            <option value="Sensor Change" class="translate">Dexcom Sensor Change</option>
-                            <option value="Insulin Change" class="translate">Insulin Cartridge Change</option>
-                            <option value="D.A.D. Alert" class="translate">D.A.D. Alert</option>
-                        </select>
+                    <label for="eventType" class="w260">
+                      <span class="translate">Event Type</span>:
+                      <select id="eventType" class="">
```
jasoncalabrese: 2 elements with the same `eventType` id.

static/css/drawer.css
```diff
@@ -68,18 +68,11 @@ input[type=number]:invalid {
...
-}
-
-#treatmentDrawer .left-column span {
-  width: 110px;
-  display: inline-block;
-}
-
-#treatmentDrawer .left-column input, #treatmentDrawer .left-column select {
-  width: 140px;
+#treatmentDrawer .w260 {
```
jasoncalabrese: the name of the class should be semantic not the name of the width, something like `wide` would be better

jasoncalabrese:
I don't like the new alignments, would like to make it look like before

screen shot 2015-07-23 at 5 49 30 pm

jasoncalabrese:
The approach used in #735 is a little simpler and keeps the same compact/aligned layout as before

PR #735, opened 2015-07-24
adjust treatment form css so labels have a fixed width

jasoncalabrese:


jasoncalabrese:
French translations need some work, but at least they don't break the layout.

@MilosKozak, main issue was the `: `'s, since they were outside the spans they were creating new lines and really making a mess.

screen shot 2015-07-23 at 10 25 36 pm

PR #752, opened 2015-07-27
Fixes === comparison between a string and a number on CAGE alerts

sulkaharo:


sulkaharo:
Heh, the first commit is actually unneeded with the change to the env variable processing...

jasoncalabrese:
Don't know what Codacy doesn't like about this, looks good to me

PR #756, opened 2015-07-30
Bulgarian translation part 1

xpucuto:
I have translated till  You secret API, i will do my best to finish the rest in the next few days. As i see the translation will be int he new version of Nightspot where yo can edit you Events.  
Please let me know if have to change anything in the language in order to be viewed correctly . 
Hristina

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -24,6 +24,8 @@ function init() {
       ,fr: 'Lu'
       ,pt: 'Seg'
       ,ro: 'Lu'
+      ,бг: 'Пон'
+     
       }
     ,'Tu' : {
       cs: 'Út'
@@ -32,6 +34,7 @@ function init() {
       ,fr: 'Ma'
       ,pt: 'Ter'
       ,ro: 'Ma'
+      ,bg: 'Вт'
       },
     ',We' : {
       cs: 'St'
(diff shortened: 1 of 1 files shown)


lib/language.js
```diff
@@ -422,6 +473,7 @@ function init() {
       ,fr: 'Percentiles'
       ,pt: 'Percentis'
       ,ro: 'Grafic percentile'
+      ,бг: "Процентна графика"
```
jasoncalabrese: Codacy Issue found: [Mixed double and single quotes.](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/819607010/issues/source?bid=2452379&fileBranchId=2748601#l476)

xpucuto: Oh sorry I must have done it by mistake . 
I hope there not many like that . I will also check it . 
Hristina

jasoncalabrese:
@xpucuto, looks like you used a mix of `бг` and `bg` for the language code.  I think we should stick to ascii for the language code to prevent any issues with env vars in different environments.

Also take a look at the quote issue above, just need to change the quotes on that 1 line I think.

xpucuto:
Ok i will fix that and send a new request . Thank you

jasoncalabrese:
This was replaced with #759

PR #759, opened 2015-07-31
Bulgarian language - test version  - 1

xpucuto:
Hi , i have changed the бг то bg  and used only ' .  Can you just check if i have to make any other changes or its fine that way :) 
Hristina


jasoncalabrese:
Looks good, thanks

PR #767, opened 2015-08-03
Bulgarian language - full - edit 03.08.2015

xpucuto:
I have tested the person and saw that at some places the words need to be shorted and tried to fix that. 
Still there is not option to translate announcement in the event type list .

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -845,7 +845,7 @@ function init() {
       ,fr: 'IG'
       ,pt: 'IG'
       ,ro: 'CI'
-      ,bg: 'Гликемичен Индекс'
+      ,bg: 'ГИ'
       }
     ,'Edit record' : {   
       cs: 'Upravit záznam'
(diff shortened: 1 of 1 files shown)


jasoncalabrese:
Thanks for reviewing the translations, for announcement I think you just need to add the key and translation, the other languages will catch up.  We'll need a tool to show what's missing for each language.

xpucuto:
I will do my best to make as many correction as i can and not send you many edits. I am aware that you have much important things to do :) 
The tool is a good idea but there are also many words from the program who are not in the language file. Step by step the final result will be more than magnificent. 
P.s. - is it possible at some future point of update the basal rate to be displayed as a graph below the glucose readings ?So if you want to test your basal to be able to see where to make the correction . Just curious about it.
Can i activate the food database or it is not ready yet?
I am not a developer and sometimes i ask stupid questions - sorry about that. But if there is some tests that can be done i can help with it :)

Hristina

PR #779, opened 2015-08-10
adjust the xScale.domain so we don't need the placeholder point kludge

jasoncalabrese:


jasoncalabrese:
going to run this on my prod site today

jasoncalabrese:
Fixed a some bugs last night, but has been working good since then

PR #782, opened 2015-08-11
profile editor, hash authentication

MilosKozak:


sulkaharo:
I get multiple errors to my Javascript console when running this:

[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (jquery-ui.min.css, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (jquery-ui.min.js, line 0)
[Error] Failed to load resource: the server responded with a status of 401 (Unauthorized) (verifyauth, line 0)
[Error] TypeError: undefined is not an object (evaluating '$('#chartContainer')[0].getBoundingClientRect')
	debouncedUpdateChart (bundle.js, line 41765)
	complete (bundle.js, line 7698)
	delayed (bundle.js, line 7708)

Also having trouble understanding how the authentication works? Pressing Authenticate in the editor doesn't seem to do anything in Chrome or Safari.

MilosKozak:
try npm install 1st

jasoncalabrese:
the `getBoundingClientRect` thing is from using the full client on a different page, we need to adjust the client init to prevent trying to update the chart in cases like this.  I like how it sets the browser title though.

If we get that right and we can build new pages that use a subset of the features easily

static/profile/js/profile.js
```diff
@@ -0,0 +1,486 @@
...
+  }
+  
+  function fillTimeRanges(event) {
+    var mmoltime = [
+      '0:00','0:30','1:00','1:30','2:00','2:30','3:00','3:30','4:00','4:30','5:00','5:30',
+      '6:00','6:30','7:00','7:30','8:00','8:30','9:00','9:30','10:00','10:30','11:00','11:30',
+      '12:00','12:30','13:00','13:30','14:00','14:30','15:00','15:30','16:00','16:30','17:00','17:30',
+      '18:00','18:30','19:00','19:30','20:00','20:30','21:00','21:30','22:00','22:30','23:00','23:30'
+    ];
+    var mgtime = [
```
sulkaharo: mmoltime and mgtime are pretty misleading as variable names, as the units being used doesn't have anything to do with whether the user prefers to be on a 24 or 12 hour clock and/or if the units being used is mmol or mgdl. Not a biggie but potentially confusing.

MilosKozak: yeah. name from some prehistoric version

jasoncalabrese: think 24 time is good for this anyway, my pump shows basal in 24 hour time even though we have it set for 12 hour time in other cases

static/profile/js/profile.js
```diff
@@ -0,0 +1,486 @@
...
+        c_profile[e.array][i].time = toTimeString(lowesttime);
+      }
+      return tr[0].outerHTML;
+    }
+    
+    // Fill dropdown boxes
+    [{prefix:'pe_basal', array:'basal', label:'Rate: '},
+     {prefix:'pe_ic', array:'carbratio', label:'IC: '},
+     {prefix:'pe_isf', array:'sens', label:'ISF: '}
+    ].forEach(function (e) {
```
jasoncalabrese: should use `_.forEach` some old browsers don't like `forEach`

static/profile/js/profile.js
```diff
@@ -0,0 +1,486 @@
...
+  $('#pe_status').hide().text('Loading profile records ...').fadeIn('slow');
+  $.ajax('/api/v1/profile.json', {
+    success: function (records) {
+      c_profile = {};
+      mongoprofiles = records;
+      // create new profile to be edited from last record
+      if (records[0] && records[0].dia) {
+        // Use only values(keys) defined in defaultprofile, drop the rest. Preparation for future changes.
+        c_profile = _.cloneDeep(defaultprofile);
+        for (var key in records[0]) {
```
jasoncalabrese: use `_.forEach` and we can skip the `hasOwnProperty` thing

static/profile/js/profile.js
```diff
@@ -0,0 +1,486 @@
...
+      alert('Time ranges of target_low and target_high don\'t  match. Values are restored to defaults.');
+      profile.target_low = _.cloneDeep(defaultprofile.target_low);
+      profile.target_high = _.cloneDeep(defaultprofile.target_high);
+    }
+  }
+  
+  function initeditor() {
+    // Load timezones
+    $('#pe_timezone').empty();
+    moment.tz.names().forEach(function addTz(tz) {
```
jasoncalabrese: Nice :)

jasoncalabrese:
I feel asleep putting the kids to bed last night, but I'm planing to play with this tonight

static/profile/js/profile.js
MilosKozak: ```hasOwnProperty ``` wants codacy there

MilosKozak: you can do this small changes on merge. i'm unable to check it with codacy as it takes always 5h+

PR #789, opened 2015-08-14
food editor

MilosKozak:


MilosKozak:
it's failing because careportal.test.js is not updated from profile editor

PR #802, opened 2015-08-16
Wip/pebblebwp

sulkaharo:
Send BWP and calculated outcome to Pebble, just in case the watch face would magically show it. ;) Tested to be backward compatible with the latest watchface (which just ignores the extra fields).


jasoncalabrese:
Instead of extending the `/pebble` endpoint I'd like to create a new summary endpoint that designed to support multiple devices and uses

PR #815, opened 2015-08-17
update2 it-laguage.js

francescaneri:
Hi Jason, adapted words. Missing translation of announcement. (menu food). no charge translation of amount in gramns and the explanation of shows raw data BG. Thanks Matteo
image

image

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -223,7 +223,7 @@ function init() {
       ,ro: 'Categorie'
       ,bg: 'Категория'
       ,hr: 'Kategorija'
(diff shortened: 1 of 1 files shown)


lib/language.js
```diff
@@ -2199,7 +2199,7 @@ function init() {
       ,ro: '3h'
       ,bg: '3часа'
       ,hr: '3h'
-      ,it: '3h'
+      ,it: '3ORE
```
jasoncalabrese: Missing `'` here, I can fix when I merge if you don't beat me to it

PR #816, opened 2015-08-17
Bulgarian language  edit - 17.08.2015

xpucuto:
Hi , i changed some of the translation and added announcement to the list .  
Is there another language file with all the words from the treatment profile?
Is it going to be possible to have more than one treatment profiles and change to the stored setting when that occur.  For example when you are sick. 
If there is anything else i am ready to help


lib/language.js
```diff
@@ -2500,6 +2500,18 @@ function init() {
       ,hr: 'Vrijednost GUK-a'
       ,it: 'Lettura glicemie'
       }
+      ,'Announcement' : {
+      cs: ''
```
jasoncalabrese: Shouldn't set the other language to `''`, that will cause a empty values instead of falling back to English

jasoncalabrese:
@xpucuto I added 1 comment about the empty values for Annoucement

jasoncalabrese:
This was replaced by #821

PR #828, opened 2015-08-20
if there isn't a battery value, don't set the field

jasoncalabrese:
trying to fix an issue with share and pebble with @oamyoamy

### DIFF
--- a/lib/pebble.js
+++ b/lib/pebble.js
@@ -78,7 +78,10 @@ function prepareSGVs (req, sbx) {
       bgs[0].bgdelta = bgs[0].bgdelta.toFixed(1);
     }
 
-    bgs[0].battery = data.devicestatus && data.devicestatus.uploaderBattery && data.devicestatus.uploaderBattery.toString();
+    if (data.devicestatus && data.devicestatus.uploaderBattery) {
+      bgs[0].battery = data.devicestatus.uploaderBattery.toString();
+    }
+
     if (req.iob) {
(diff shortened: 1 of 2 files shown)


jasoncalabrese:
If this works need to add a test before merging

oamyoamy:
Tested with site using Share variables and Pebble watchfaces: Nightscout 7.3 and CGM in the Cloud 7.2. Both loaded successfully. The battery variable is now gone from the pebble string:

{"status":[{"now":1440120370989}],"bgs":[{"sgv":"173","trend":5,"direction":"FortyFiveDown","datetime":1440119678000,"bgdelta":0,"iob":"0"}],"cals":[]}

jasoncalabrese:
Great, I'll get this merged tonight

oamyoamy:
May need to ignore my prior test. My site with Share bridge variables is
failing... xDrip may have grabbed my Share and pushed data to the Share
server rather than the Share app. The above test I ran may, in fact, not
have been with Share app upload.

jasoncalabrese:
@oamyoamy, if this doesn't fix your issue I'll try something different, but there wasn't any reason to send `battery: false` so I wanted to get this in even if it's not that full fix.

oamyoamy:
This change (on dev) seems to have fixed it. The user who reported the issue was running the 0.8-beta1 branch. She deployed a copy of dev with the update, and it is working.

jasoncalabrese:
Great

Wish the interface between the server and pebble want so sloppy.  So many assumptions.  This is why I want a new endpoint.

PR #850, opened 2015-08-26
language translations

MilosKozak:


lib/language.js
```diff
@@ -7,6 +7,21 @@ function init() {
...
   }
 
+  var languages = [
+    { code: 'bg', language: 'Български език' }
+    , { code: 'hr', language: 'Hrvatski' }
+    , { code: 'cz', language: 'Čeština' }
+    , { code: 'dk', language: 'Dansk' }
+    , { code: 'en', language: 'English' }
+    , { code: 'fr', language: 'Français' }
+    , { code: 'de', language: 'German' }
```
jasoncalabrese: Should this be Deutsch instead of German?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Nightscout",
-  "version": "0.8.0-beta1",
+  "version": "0.8.0-beta2",
```
jasoncalabrese: ugh, beta2 bump didn't make it to dev

lib/language.js
MilosKozak: once you start copy/paste you're in trouble :(

PR #865, opened 2015-08-29
Wip/walk prop - dev experiment

bewest:
Experiment to handle type of incoming parameters.

### DIFF
--- a/lib/entries.js
+++ b/lib/entries.js
@@ -10,6 +10,8 @@ var ObjectID = require('mongodb').ObjectID;
 \**********/
 
 function find_sgv_query (opts) {
+  ['date', 'sgv'].forEach(function iter_prop (key) {
+  });
   if (opts && opts.find && opts.find.sgv) {
     Object.keys(opts.find.sgv).forEach(function (key) {
       var is_keyword = /^\$/g;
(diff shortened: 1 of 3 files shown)


lib/entries.js
```diff
@@ -10,17 +11,32 @@ var ObjectID = require('mongodb').ObjectID;
...
-  if (opts && opts.find && opts.find.sgv) {
-    Object.keys(opts.find.sgv).forEach(function (key) {
-      var is_keyword = /^\$/g;
-      if (is_keyword.test(key)) {
-        opts.find.sgv[key] = parseInt(opts.find.sgv[key]);
-      }
-    });
-  }
+
+  opts = walk_prop('date', parseInt)(opts);
```
jasoncalabrese: an array to list these int fields fields would be nice, maybe a map?

```
var typedProps = [
  date: parseInt
  , sgv: parseInt
];

bewest:
All done, this should probably move to it's own module, any opinion there?

jasoncalabrese:
yeah, would be nice to be able to reuse for the treatments api, easier to test then too

bewest:
Where to put it? In `storage.js`?  In `lib/foobar`?  In `lib/utils.js`?  `lib/` feels like big directory.

jasoncalabrese:
maybe in something like `/lib/api/walker.js`

jasoncalabrese:
guess it's more generic than that, maybe just adding it to utils?

PR #868, opened 2015-08-30
Wip/walk prop

bewest:
Add some comments


jasoncalabrese:
now we just need to add docs like this to the rest of the code base

PR #872, opened 2015-08-30
0.8 beta2

LynneRB:


jasoncalabrese:
This is going in the wrong direction, but it looks like your beta2 branch is current and ready to deploy

PR #887, opened 2015-09-01
Finnish localization

sulkaharo:
First pass Finnish localization. I assume the reporting side needs work. ;) Seems to fit the UI after some tweaking.

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -21,6 +21,7 @@ function init() {
     , { code: 'pt', language: 'Português (Brasil)' }
     , { code: 'ro', language: 'Română' }
     , { code: 'sv', language: 'Svenska' }
+    , { code: 'fi', language: 'Suomi' }
   ];
 
   var translations = {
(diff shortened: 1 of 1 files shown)


MilosKozak:
reporting should be already covered by current language.js

sulkaharo:
Yeah I meant the actual content I put in will probably need polish.

jasoncalabrese:
Wonder if we should add country names name next to the language, maybe for some of the others too?

MilosKozak:
we can do it if people will ask for it. i think everyone know his country code. or not?

jasoncalabrese:
was thinking more for everyone else that doesn't know the languages, just to be able to look at the list

MilosKozak:
hmmm don't know. never seen it . either english list or only national

MilosKozak:
@jasoncalabrese maybe we could add a flag. this would be pretty clear. but there could be an issue language vs country
or something like this 
image

Issue #888, opened 2015-09-01
Announcements


oamyoamy:
Checking intended functionality (compared to screenshots I've seen). When sending an "announcement":

1) pushover notification is working (message text comes through)
2) IFTTT notification is working (message text comes through)
3) site status bar (top) changes to "Urgent Announcement" (red) or "Announcement" (yellow) (depending on level of BG entered) -- but text of announcement is NOT shown in browser. (I saw testing images from Jason with a "juicebox" announcement... which seemed to suggest announcement text would go to the status bar area.)

SOLVED:
If you do NOT enter a BG, the announcement status goes to the status bar properly. (Orange text.)

FUNCTIONALITY:
IF entering a BG and selecting TYPE of CP entry as announcement, shouldn't you see the level (Urgent Announcement or Announcement) with the "message" of the announcement appended to the end of it: Announcement: Juice Box Needed!

jasoncalabrese:
Think this is a bug, I'd like the message to be displayed always.  With the Urgent/Warning prefix if a BG is entered.

oamyoamy:
Yes. That is what I would expect, too, from the feature.

PR #891, opened 2015-09-02
when an alarm is triggered by an announcement, show the message

jasoncalabrese:
screen shot 2015-09-01 at 6 43 37 pm

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -8,6 +8,7 @@ var language = require('../language')();
 var sandbox = require('../sandbox')();
 var profile = require('../profilefunctions')();
 var units = require('../units')();
+var levels = require('../levels');
 var times = require('../times');
 
 var client = { };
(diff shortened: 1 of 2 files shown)


jasoncalabrese:
resolves #888

oamyoamy:
Testing wip/announce-alarms -- alarm triggered by announcement is now showing both alert level (e.g., urgent) and announcement text. 

New issue: Announcement-triggered alarm is reappearing every 1 minute, even though snoozed via browser. (Tested at both red and yellow thresholds; same behavior.)

jasoncalabrese:
So the pushover isn't getting cleared by the browser?

oamyoamy:
Correct!

oamyoamy:
The browser alarm itself keeps replaying. Every 1 minute. Despite snoozing it for 120 mins in browser. (I tried a 30 min snooze as well.)

jasoncalabrese:
@oamyoamy, I looked over things that could cause that and the only possibility I see is that server lost the pushover receipts.  So when the web alarms was cleared the server wasn't able to tell pushover to cancel the alarms.  Before the receipts were only stored for 15mins, I extended that to an hour.  In any case I don't think it was related to this fix.

Issue #979, opened 2015-09-08
Question: Mongo Node.js Driver 2.0.34 with MongoDB 3.x for C-R-M usage


gregwaehner:
Question - With Funnel Cake (0.8.0), the package.json dependency was upgraded to mongodb 2.0.34. I'm assuming this is the 2.0.34 node.js mongodb driver. From the release notes, it's not clear that it is compatible with MongoDB 3.x in relation to how C-R-M is using it. From the "roadmap", there is no release date for the 3.0 node.js driver. However, it's possible that the specific binary incompatibilities that have been broken are not used by C-R-M, in which case, it wouldn't be a problem. As noted in the MongoDB compatibility release notes, the minimum compatibility with 3.x for node is 1.4.29. Just want to confirm that users who upgrade to FC will not have any C-R-M related mongodb 3.x issues 

(I think the answer is that there is no problem... but it depends on how the driver is used).

Minimum Compatibility for Node Driver and Node. 3.x
---------------------------------
http://docs.mongodb.org/master/release-notes/3.0-compatibility/

All MongoDB Node Driver Versions
------------------------
https://jira.mongodb.org/browse/NODE?selectedTab=com.atlassian.jira.jira-projects-plugin:versions-panel&subset=-1

MongoDB Node.js Road Map
---------------------------
https://jira.mongodb.org/browse/NODE/?selectedTab=com.atlassian.jira.jira-projects-plugin:roadmap-panel&subset=-1

jasoncalabrese:
With Nightscout 0.4.0 we used `"mongodb": "^1.4.7"` which basically translates to 1.4.x.  The `1.4.29` version that's compatible with MongoDB 3.0 was released on Jan 21, 2015.  So any site deployed or redeployed after that would pickup a version of the driver compatible with MongoDB 3.0.

Will 0.8.0 we updated the mongo driver to `"mongodb": "2.0.34"` that's > than the 2.0.14 version needed for MongoDB 3.0.  For 0.8.1, I'm setting it to `"mongodb": "^2.0.42"`, that will select the most current 2.0.x version.

PR #995, opened 2015-09-12
Select mongodb 2.0.x instead of a specific version

jasoncalabrese:
`"mongodb": "^2.0.42"`


jasoncalabrese:
After double checking for #979 I noticed that we were using a specific version here instead of using `~`

PR #1045, opened 2015-09-17
When loading profiles from Mongo, sort by the actual field that tells…

sulkaharo:
… when the profile was created. Load only one profile to speed up the process and consume less resources. Removed old check for dia field existence from iob-cob days.

### DIFF
--- a/lib/data.js
+++ b/lib/data.js
@@ -295,15 +295,12 @@ function loadTreatments (data, ctx, callback) {
 }
 
 function loadProfile (data, ctx, callback) {
-  ctx.profile.list(function (err, results) {
+  ctx.profile.last(function (err, results) {
     if (!err && results) {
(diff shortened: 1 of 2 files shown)


sulkaharo:
@MilosKozak or @jasoncalabrese can you LGTM if this works for you. Tested on our deploy.

MilosKozak:
for me correct profile is used after server restart. new/updated  profile is not detected by server.
positive change it use correct record

sulkaharo:
Do you mean the server doesn't reload the new profile? I'm seeing changes happen on the fly on the client side.

MilosKozak:
retested and it works. :+1: 
good job. you found it

jasoncalabrese:
Should update the `api.indexedFields` too current indexing `validfrom` instead of `startDate`

PR #1064, opened 2015-09-19
Fix delayed carbs

scottleibrand:
@cjo20 found a problem with my units that turned out to be a bug in the formula for delayedCarbs, where I was multiplying by ISF and dividing by IC ratio, but should've been doing the reverse.  This may account for the problems @jasoncalabrese and others were seeing on children trying to use wip/iob-cob, as that error results in a much bigger effect (6x as large) in children as it does in adults.  For now I have adjusted liverSensRatio from 1 to 8 to compensate for half the 16x difference seen with adult ratios like 40 and 10.

This will mean that the effect of delayedCarb on displayed COB will be less for everyone, but especially for kids.  COB will decay much closer to the straight carb absorption rate, which should be much more intuitive for people using it, and should also result in less likelihood of bolusing too much for any carbs that COB says are delayed, but may not actually be.

### DIFF
--- a/lib/plugins/cob.js
+++ b/lib/plugins/cob.js
@@ -30,7 +30,8 @@ function init() {
       return {};
     }
 
-    var liverSensRatio = 1;
+    // Adjusted from 1 to 8 to compensate for half of the effect (in adults; much less in kids) of fixing units on the delayedCarbs term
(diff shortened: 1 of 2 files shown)


jasoncalabrese:
Thanks, good catch

PR #1116, opened 2015-09-27
update devicestatus to be more consistent with entries

jasoncalabrese:
Now that the old direct mongo upload is gone we can use a query to prevent the sorting of all data, to find the last uploader battery status.

Need to test with xdrip and mmcommander rest uploads

@sulkaharo @stavlor

### DIFF
--- a/lib/devicestatus.js
+++ b/lib/devicestatus.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var find_options = require('./query');
+
 function storage (collection, ctx) {
   var ObjectID = require('mongodb').ObjectID;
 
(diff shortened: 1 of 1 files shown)


bewest:
Looks good to me.

sulkaharo:
Is the plan that we'll tell users that we'll start requiring REST going forward?

jasoncalabrese:
Not that we're requiring rest, just that we require the `created_at` field for device status to be a ISO string.  Before there were broken direct mongo uploads, that used the wrong format, but those all broke.  All update uploaders should use the expected data format.

In the past this caused us to sort the whole device status collection and take the last doc., now we can filter just like the other collections.

PR #1183, opened 2015-10-09
specify shorter connecti timeout

bewest:
Hat tip to @oamyoamy
See also:
* http://blog.mongolab.com/2013/10/do-you-want-a-timeout/


jasoncalabrese:
Maybe we can use https://github.com/mongolab/mongodb-uri-node or do something similar so the options passed on the uri override these defaults?  Or does the `mongodb` module already do that?

bewest:
I think `mongodb` already does that; looks like that library specifically makes the password and trailing slash optional.

Here's relevant mongodb sources:
* https://github.com/mongodb/node-mongodb-native/blob/4950fd81d5b3505721298d91572f0278b2171cce/lib/mongo_client.js#L393-L414
* https://github.com/mongodb/node-mongodb-native/blob/4950fd81d5b3505721298d91572f0278b2171cce/lib/url_parser.js#L186-L193

So putting it in mongo string is easy way to test.

bewest:
Setting timeout to 30 seconds now.  Initial scattered reports indicate some success.  No ill-effects seen, possibly helps make the Azure-MongoLab connection more reliable.

jasoncalabrese:
I can test this tonight if you want, but looks good if you just want to merge

PR #1205, opened 2015-10-18
RO update Oct, 18 2015

bogdangorescu:
Added Romanian translation for all text in beta1


jasoncalabrese:
we don't need to update the beta1 branch, everything that makes it to dev will be in the next release beta2, final, or whatever we call it

PR #1247, opened 2015-10-23
Wip/bewest/increase upload limit

bewest:
Bunch of failed attempts, the last one, installing a "default" in a "global" namespace seemed to work, although I tried at higher level and lower levels and tried.  This seems to be the goldilocks spot in the last commit, but I have no idea if the other areas are also needed.


jasoncalabrese:
We should make sure this is tested on Azure and Heroku

jasoncalabrese:
Thanks for testing on Azure @StuArnold

PR #1201, opened 2015-10-17
A time ago plugin (part of #1375, shouldn't be merged on it's own)

jasoncalabrese:
Goals

  1. have system wide stale data alarms when there is an issue with an uploader, and still maintain client side stale data alarms when there is a network issue between client and server
  1. preserve existing behavior for defaults
  1. Introduce independent alarm channels, for example, BG based alarms shouldn't be snoozed acking a stale data alarm
  1. Put `cage` in it's own alarm channel

PR #1253, opened 2015-10-24
Plugin to show pump IOB from MiniMed Connect

mddub:
Mostly copied the format of the `iob` plugin. Confirmed working with beta testers.

I'd like to get this in first, but afterwards I'll file an issue so we can agree on the format of the `pump_status` entries (like normalizing MM's `conduitBatteryLevel` to `uploaderBatteryLevel`) so that this also works with RileyLink. I also haven't checked if MMCommander forwards this data too, but assuming it's reading MySentry packets, it should be able to show this data, too.


lib/client/index.js
```diff
@@ -28,6 +28,7 @@ client.init = function init(serverSettings, plugins) {
     , isInitialData = false
     , SGVdata = []
     , MBGdata = []
+    , pumpStatuses = []
```
jasoncalabrese: Codacy Issue found: ['pumpStatuses' is defined but never used.](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/1181182976/issues/source?bid=2452379&fileBranchId=2921449#l31)

lib/pebble.js
```diff
@@ -104,9 +105,23 @@ function addExtraData (first, req, sbx) {
...
+      if (result) {
+        first.iob = result.pebbleDisplay;
+      }
+    }
+  }
+
   addDelta();
   addBattery();
   addIOB();
+  addPumpIOB();
```
jasoncalabrese: not sure we should be adding something new to `/pebble`, this thing is so fragile with all the random pebble watchface versions and other apps pulling from it

lib/plugins/pumpiob.js
```diff
@@ -0,0 +1,71 @@
...
+function init() {
+
+  var pumpiob = {
+    name: 'pumpiob'
+    , label: 'Pump IOB'
+    , pluginType: 'pill-major'
+  };
+
+  pumpiob.setProperties = function setProperties(sbx) {
+    sbx.offerProperty('pumpiob', function setPumpIOB () {
```
jasoncalabrese: I was hoping we'd be able to use the same `iob` property here, then any other plugins like BWP could get the info they need.

lib/plugins/pumpiob.js
```diff
@@ -0,0 +1,71 @@
...
+
+  pumpiob.updateVisualisation = function updateVisualisation(sbx) {
+    var prop = sbx.properties.pumpiob;
+
+    var info;
+    if (prop && prop.uploaderBattery) {
+      info = [{label: 'Uploader Battery', value: prop.uploaderBattery}];
+    }
+
+    sbx.pluginBase.updatePillText(pumpiob, {
```
jasoncalabrese: Would someone want have both IOB pills display or only 1?  If they had 1, and this plug could just set the property and let the normal iob plugin display it.

mddub: I was thinking maybe if someone is in the habit of manually entering treatments already, they can enable pump IOB as a sanity check. But that someone is definitely not me; I don't use care portal. What do you think?

* If that's not a plausible use case, yeah, pumpiob could set the same property and the README can warn against enabling both.
* Alternatively, instead of adding a separate plugin, these could be extended settings of the iob plugin to make it read from pump_status entries instead of calculating treatments.

Either way, a few other places (cob, reports) call `iob.calcTotal` directly instead of reading the property, so they need to be updated anyway.

lib/plugins/pumpiob.js
mddub: Does the `iob` plugin take basal into account? Pump IOB considers bolus only (@scottleibrand actually suggested calling this something like bolus-on-board to distinguish them), so it would result in different calculations from other plugins if so.

lib/pebble.js
mddub: It sets the same key as the existing Pebble IOB code, and I think this is a big enough win that it's worth it. (A lot of people already using NS with Dexcom say they're getting a Connect just for this kind of data.)

jasoncalabrese:
The main thing we need to figure out with this is should Pump IOB be different from Careportal/Treatment API IOB.  Also what about the bIOB  (Bolus only IOB from OpenAPS, since the other IOB calc includes temps).

Maybe we can get a group together to talk about the options and how we think things will fit together.  Gitter, Hangout, etc.

@bewest @scottleibrand @ps2 @sulkaharo @MilosKozak

scottleibrand:
Probably don't want to report bIOB directly, as that's calculated off an
accelerated DIA and won't make sense to most users.

Otherwise, we should be calculating the same IOB given the same inputs,
either from treatments or from pump history.

-Scott

jasoncalabrese:
One partial solution to this is to have a single iob plugin and
merge/dedupe entries and treatments.

But the iob being reported here is something different, it's the value
calculated on the pump.

scottleibrand:
I would suggest we report net IOB if we can calculate it from pumphistory
data, otherwise pump IOB if we have it from Connect, otherwise net IOB from
human-entered treatment data.  I trust data coming from the device over
human-entered data, but net IOB is better than pump IOB.

-Scott

mddub:
Is there a standard format which distinguishes pumphistory treatment data from manual treatment data in NS? I thought @bewest just recently started working on that. That would need to be standardized before we can add a heuristic like that.

If we want a single IOB plugin with different modes of deriving the number, I think it's better to toggle the mode via an explicit setting rather than have it magically choose the right data source. I suggest we leave the default behavior of calculating net IOB based on treatments (regardless of their source), and add an extended setting `IOB_USE_PUMP_IOB` which causes it to read pump IOB from `pump_status` entries instead. What do you think?

bewest:
FWIW, I'm reporting [treatments to NS](https://gist.github.com/bewest/85ed5f624a81eba25a9e).
Using this [branch](https://gist.github.com/bewest/85ed5f624a81eba25a9e).

Also, I've been pulling out functions/calculators from both NS and putting them in other repos.  Right now, https://github.com/nightscout/dcalc/tree/master/lib.

My suggestion is to put the NS calculators in this dcalc repo, and then require them in, but not sure how much bandwidth there is for that on top of implementing the features.

jasoncalabrese:
There are some benefits to treatments instead of entries, it's only using an iso date, and we send all treatments (within a range) to the client instead of splitting them up like we do for svgs, mbgs, cals, etc.  So treatments are more like the generic event stream that I think we want.

Also @MilosKozak added special handling for `Sensor*` `eventType`s to extend the range of what's sent to the the client (beyond the default 8 days), we could do that for other types.

Only issue I see with using treatments is that we'll be sending tons more treatments, and that could become a performance issue.  Think we're already seeing that with the temp basals.

jasoncalabrese:
I'm wondering it we should go with this for now, and look at combining the iob plugins later.

lib/pebble.js
jasoncalabrese: Looks like pump iob will override treatment iob, not sure if that would be expected.  Maybe only do that if treatment iob isn't defined/enabled?

mddub:
@jasoncalabrese: addressed your comment about prioritizing iob over pumpiob in /pebble. I'm also fine with changing this to be part of iob + an extended setting `IOB_USE_PUMP_IOB` as I suggested, if you think that's a better way to go.

jasoncalabrese:
Maybe something like `IOB_SOURCE="pump treatments bolusonly"` and let the order define the priority.  Then have a way to fallback if one isn't available and an other is. Could expose that on the client too.

mddub:
But for now there would only be `pump` and `treatments`, right? AFAICT the iob plugin has only one mode of calculation, which is bolus-only.

jasoncalabrese:
We're planning to include temp basals in the iob calculation just like OpenAPS.  Hopefully sharing the same calculator.  Via the new nightscout/dcalc repo.

mddub:
I agree with using `IOB_SOURCE`, but I'm not sure I agree with falling back. Imagine your source list is `"pump treatments bolusonly"` because you're using a Connect/RileyLink. If the Connect/RL goes out of range or dies, there are no recent `pump_status` entries with IOB, so it falls back to treatments, but since you haven't been entering any treatments, IOB is reported as 0. Won't it always be indeterminate whether the absence of treatments means no boluses or no data? It seems to me it would be better to not report IOB in that case. Less surprising, better for troubleshooting.

I think(?) there would also be no clear way of falling back between bolus-only iob and temp basal-aware iob, since the presence of treatments means either both or neither can be calculated.

MilosKozak:
@mddub i think in your case you set `IOB_SOURCE=pump` only

jasoncalabrese:
Fallback does complicate things, only supporting a single source at first makes sense, then if we find a good use for fallbacks we can add it then

tests/iob.test.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
-require('should');
+var should = require('should');
```
jasoncalabrese: Codacy Issue found: ['should' is defined but never used.](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/1243400235/issues/source?bid=2452379&fileBranchId=2921449#l3)

mddub:
Updated the iob plugin to use `IOB_SOURCE` to toggle between pump IOB and treatments. Either mode sets the same iob property read by other plugins. The way I cleaned it up makes it straightforward to remove `calcTotal` and `calcTreatments` and drop in dcalc instead, after @bewest or I expose the functions in that package.

There were still a handful of places calling `calcTotal` directly instead of using the property. My solution was to modify the call signatures for the iob and cob plugins:
* `iob.calcTotal(treatments, profile, time)` --> `iob.calc(sbx, time)`
* `cob.cobTotal(treatments, profile, time)` --> `iob.cobTotal(sbx, time)`

Less functional style and more complex to test, but now able to use any source or calculation method based on env settings without callers needing to know about what data to pass. I did this based on what I saw in the bwp plugin.

I also made the day to day report able to plot IOB from the Connect, @MilosKozak:

screenshot 2015-11-02 22 55 54

Incidentally, this graph shows how continuous sampling of pump IOB from a Connect can be a decent proxy for bolus timing and amount.

MilosKozak:
looks nice
there are some changes on passing params in 0.9
it will be hard to merge

MilosKozak:
watching your diff i changed my mind to HARD to merge

MilosKozak:
changing iob, cob to use only sbx is a step wrong way i think. we limit the usage of cob, iob only to one purpose because sbx is designed to hold most recent data

merging all data, filling sbx and reusing for iob, cob calculation is just a hack which goes against philosophy of reports and make it less clear what it does

i agree it's going to be great function but i'd wait with merge until all developers send their comments and modifications to 0.9 (dcalc too) and the way of handling different IOBs will be defined because it's not only about connect

my personal opinion is every IOB is going to have own plugin because this
https://github.com/nightscout/cgm-remote-monitor/pull/1253/files#diff-99d1e261bda3eceee5c0acc56f9659f1R38
is not sustainable for more IOBs
@jasoncalabrese  @bewest

mddub:
My original proposal was to add pump IOB as its own plugin. The issue was raised that other plugins would then not be able to use those IOB values.

I was considering a mixin pattern to keep each IOB mode in its own plugin, but it felt like overkill/YAGNI to do that right now since that `updateVisualisation` you pointed out is the only part where they differ.
https://github.com/mddub/cgm-remote-monitor/commit/22e42d92e46fb645e0042cbc04ec8c45554b0380#diff-99d1e261bda3eceee5c0acc56f9659f1R25

If we want other plugins to use a consistent IOB, different IOB modes need to not only set the same property, but also be able to be called the same way. That wasn't simple for the places `calcTotal` is called, particularly reports. Using pump IOB requires extended settings for source and ideally recency threshold. I guess those could be part of `profile`?

jasoncalabrese:
I don't mind the `usePumpIOB` stuff too much for now, we can always do some refactoring later when we have other IOB's and a better idea of how they will play together.

I'm not sure about mixing pump battery/timeago into IOB though, since that might be harder to undo later when we have something better.  I have a branch that I've been slowly working on that turns the timeago pill into a plugin, and was expecting that we'd have multiple instances of it for cases like this.  (https://github.com/nightscout/cgm-remote-monitor/pull/1201)

Maybe we can reduce the scope of this a little and get some of it in now and more later?

jasoncalabrese:
I merged in the huge PR, do there are some conflicts here, but hopefully won't be too bad.

I'm ok passing sbx to the iob/cob functions, think a lot of that will be changing soon as we pull in dcalc, so having to while sbx around will be most flexible.

mddub:
Merged in dev, only a few small conflicts needed resolving. iob/cob are working.

jasoncalabrese:
Would be nice to get this PR going again, I think the device status refactoring is all ready for it now

mddub:
Yep, I'm excited to get back to this. I'll be back on the grid after New Year's... will update this to take advantage of your devicestatus changes then.

mddub:
Closing this in favor of #1560

PR #1277, opened 2015-10-26
Wip/basalprofiles

MilosKozak:


francescaneri:
image
you can fix it ? ( page food) thank you

MilosKozak:
what should i fix?

jasoncalabrese:
think @francescaneri was saying the that the Foods page is using an old favicon, need to find a way to share the common html, wonder if using jade is too much

PR #1283, opened 2015-10-28
only expect an error, not a specific message for travis…

jasoncalabrese:


jasoncalabrese:
@MilosKozak it passes!

PR #1344, opened 2015-11-09
make profile history and multiple profiles optional

MilosKozak:
@jasoncalabrese ok to use fake plugin again?


jasoncalabrese:
Yeah, and maybe it doesn't need to be fake, maybe it would make sense to have a pill for it

PR #1389, opened 2015-12-06
Include Node 4.2 and 5.1 in Travis tests

jasoncalabrese:


jareware:
Interestingly, I can `npm install` without issue on node 5.1 on OS X, with only

    npm WARN EPEERINVALID mongodb-core@1.2.24 requires a peer of kerberos@~0.0 but none was installed.

which is apparently A-OK unless you actually want to use Kerberos authentication.

jareware:
...and with the [Docker 5.1 image](https://hub.docker.com/_/node/) as well.

jareware:
So it looks like Travis can't build `contextify`, and might just be a matter of [adding `gcc` to the image](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements).

jasoncalabrese:
That's a dependency of `benv` and `jsdom` that we use for testing, but there is a new version that at the time only worked with io.js, but with node 4.x + it should be fine.

PR #1399, opened 2015-12-11
A few little ui tweaks

jasoncalabrese:
Main goal is to fix loading so that old data isn't flashed before the new data is displayed.


lib/client/chart.js
```diff
@@ -2,11 +2,8 @@
 
 var _ = require('lodash');
```
jasoncalabrese: Codacy Issue found: ['_' is defined but never used.](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/1548085120/issues/source?bid=2452379&fileBranchId=3016446#l3)

PR #1400, opened 2015-12-11
Date should not be a string

jmatheson:
Removed quotes around the date. Added some comments for clarity.


jmatheson:
cc// @jasoncalabrese

PR #1426, opened 2015-12-26
Adjust node version to fix Azure deploy issues

jasoncalabrese:
Hopefully this simple change will be a real solution for #1412

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "nightscout",
-  "version": "0.8.4-hotfixtest1",
+  "version": "0.8.4",
   "dependencies": {
     "jquery": "2.1.0",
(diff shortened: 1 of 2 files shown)


jasoncalabrese:
@amazaheri @stavlor, can you confirm that @davidebbo's fix solves the issue and that this branch now deploys cleanly?

amazaheri:
:+1:

amazaheri:
Now that this is over, Per @davidebbo suggestion it would be great to implement Deploy To feature as it will ease up deployment for non techie folks. @jeffhollan added it to LogicApp repos and it works great as an example: https://github.com/logicappsio/EventHubAPI

davidebbo:
I opened #1433 to track Deploy To Azure.

PR #1432, opened 2015-12-28
Fix logic that determines npm version to use

davidebbo:
This was caused by a bug in an older version of kuduscript (see https://github.com/projectkudu/KuduScript/issues/35). The bug has been fixed for a while, but once you generate your custom deploy.sh with the bug in it, you have to hand fix it, which is what this patch does.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -78,7 +78,7 @@ selectNodeVersion () {
       exitWithMessageOnError "getting node version failed"
     fi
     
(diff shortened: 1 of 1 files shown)


jasoncalabrese:
Nice catch, going to merge to the hotfix branch now, and we can verify that works then get the hotfix merged to master

amazaheri:
Awesome!

PR #1437, opened 2015-12-29
added basic cors configuration

GarthDB:
I'd like to add more complex configuration - like a whitelist of domains.

There is plenty of documentation on the cors library https://github.com/expressjs/cors

What would be the best way to add a whitelist configuration?


jasoncalabrese:
Except for hot fixes we target PRs to dev.  Would a single env var with a whitespace separated list of domains work?

GarthDB:
Yeah it could. I close this one and open a pr to dev.
Any recommendations on naming the whitelist env var?

jasoncalabrese:
Saw you were doing it as a plugin, if we can make that work, it's really easy to add additional extended settings.

But so far we haven't used plugins for this type of thing, if a single env var would work that night be easier. Maybe `CORS_DOMAINS` and allow multiple domains whitespace separated.

PR #1460, opened 2016-01-07
Make MM Connect plugin use devicestatus collection

mddub:
This brings in the changes from https://github.com/mddub/minimed-connect-to-nightscout/pull/1 and inserts to devicestatus instead of entries.


jasoncalabrese:
looks going, going to merge

PR #1510, opened 2016-02-09
replace a couple _.cloneDeeps with shallow _.clones

jasoncalabrese:
testing to see if this will reduce server cpu and memory usage


jasoncalabrese:
Big reduction of cpu and memory use when using lots of AP data

PR #1525, opened 2016-02-18
Fixed typos. Changed grammar. Adjusted texts to fit better.

turbofart:


jasoncalabrese:
This was merged to merged, we should never merge PRs to master, only to dev.  Wish there was a way to prevent that.

Trying to undo it would make a mess so probably better to leave it.

PR #1527, opened 2016-02-21
fix issue with drawer closing when android keyboard opens

jasoncalabrese:
draw is now full screen instead of only the chart area, also a little fix for the context growing over time.


MilosKozak:
i noticed this issue on iphone too. but not everytime

jasoncalabrese:
This is working good for me, way better than before anyway so going to merge now

PR #1560, opened 2016-03-25
Show IOB from devicestatus when available (OpenAPS or MM Connect)

mddub:
* Pass `devicestatus` to `iob.calcTotal` and `cob.cobTotal`
* Let IOB be undefined sometimes (No treatments and no devicestatus in sandbox == haven't heard from device in a while == undefined IOB, displayed as "---U". Assuming/showing "0U" in this case would be misleading.)
* ...but always return a number on /pebble for backwards compatibility with the clients in the wild

### DIFF
--- a/lib/client/boluscalc.js
+++ b/lib/client/boluscalc.js
@@ -390,14 +390,14 @@ function init(client, $, plugins) {
     // Load IOB
     record.iob = 0;
     if ($('#bc_useiob').is(':checked')) {
-      record.iob = roundTo(iob.calcTotal(client.sbx.data.treatments, client.sbx.data.profile, record.eventTime, record.profile).iob, 0.01);
+      record.iob = roundTo(iob.calcTotal(client.sbx.data.treatments, client.sbx.data.devicestatus, client.sbx.data.profile, record.eventTime, record.profile).iob, 0.01);
     }
 
     // Load COB
     record.cob = 0;
     record.insulincob = 0;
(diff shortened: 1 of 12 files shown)


bewest:
Looking pretty good.

jasoncalabrese:
Wish we merged this a little sooner, there's a conflict in `report.js`

@MilosKozak

MilosKozak:
hmm i missed this. like your changes in loading ....

MilosKozak:
it should be easy to put it together. the only conflict to be solved is i filter devicestatuses with openaps part ... you need all records

mddub:
Merged dev. The conflict was pretty small.

The only difference is that for the reports page, I leave data.devicestatus sorted by mills descending. This is required for the IOB calculation, and if I sort or assert an ordering before that calculation, the rendering takes significantly longer. The only place outside of my changes where it's used is to map each SGV to a devicestatus within 5 minutes of it, and that doesn't need optimization.

jasoncalabrese:
Was ready to merge this, but I think there's an issue when there are multiple rigs, since the latest devicestatus might not be the best, there could be another rig sending old info.

I think the fix for that isn't so bad and probably makes sense for me to do it, having OpenAPS IOB in the pill will be really nice, so I'm hoping to get to it tonight.

Also I'm wondering if we should have an option to pick the IOB source instead of doing the auto detect, maybe that's for later.

PR #1568, opened 2016-03-28
Update language.js

apanasef:
Russian translation contd
please note there seems to be a mistake in polish translation in line 2556 (?) - should be BAZAL inst of BOLUS (don't know where to report properly)


MilosKozak:
polish translation is wrong

PR #1581, opened 2016-04-02
Added node 4 and 5 to travis

jontelm:


jontelm:
My branch is passing the tests https://travis-ci.org/jontelm/cgm-remote-monitor so it's something that get Travis build to fail sometimes.

jasoncalabrese:
I tried this with #1389, but ran into issues with `benv` and `jsdom` but looks like you found a way around it.  Would still be nice to upgrade those dependencies.

jontelm:
This was the problem: https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements

PR #1601, opened 2016-04-14
Request temp basal treatments only once for reports

mddub:
Before: every time reports are generated, make a separate request for the last 32 days of temp basals, regardless of report time period.

After: every time reports are generated, pull temp basals from the treatments-by-day requests which are already being made, and cache them by day in the same way as SGV, treatments, and device status.

Only difference is that now a temp basal begun shortly before midnight on the first report day won't be graphed. `¯\_(ツ)_/¯`

### DIFF
--- a/static/report/js/report.js
+++ b/static/report/js/report.js
@@ -427,11 +427,9 @@
         if (options.order === report_plugins.consts.ORDER_NEWESTONTOP) {
           sorteddaystoshow.reverse();
         }
-        loadTempBasals(from, function loadTempBasalsCallback() { 
-          loadProfileSwitch(from, function loadProfileSwitchCallback() { 
-            $('#info > b').html('<b>'+translate('Rendering')+' ...</b>');
-            window.setTimeout(function () {showreports(options); }, 0);
-            });
+        loadProfileSwitch(from, function loadProfileSwitchCallback() {
+          $('#info > b').html('<b>'+translate('Rendering')+' ...</b>');
+          window.setTimeout(function () {showreports(options); }, 0);
           });
       }
     }
@@ -457,10 +455,12 @@
     datastorage.treatments = [];
     datastorage.devicestatus = [];
     datastorage.combobolusTreatments = [];
+    datastorage.tempbasalTreatments = [];
     Object.keys(daystoshow).forEach( function eachDay(day) {
       datastorage.treatments = datastorage.treatments.concat(datastorage[day].treatments);
       datastorage.devicestatus = datastorage.devicestatus.concat(datastorage[day].devicestatus);
       datastorage.combobolusTreatments = datastorage.combobolusTreatments.concat(datastorage[day].combobolusTreatments);
+      datastorage.tempbasalTreatments = datastorage.tempbasalTreatments.concat(datastorage[day].tempbasalTreatments);
     });
     
     report_plugins.eachPlugin(function (plugin) {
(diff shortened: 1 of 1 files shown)


MilosKozak:
huge temp basal load is there because of you can select only some days of week and over midnight basals would not appear. not an issue for openaps users but people can set long term temp basal over midnight. teoreticaly it would be enough to load -1 day only. not remember why i added 32 days :)
or load day-1 for all days too

mddub:
Hmmm, how about we decide on a max amount of time to look backwards (start of day minus 2 hours?), and use that as the `from` in `loadTreatmentData`?

MilosKozak:
well, on animas pump you can set temp up to 24h. not easy to decide how much time back load to cover most common situations. 
most "correct" and optimized solution could be:
check if previous day is loaded too and if not, load temps from this day

MilosKozak:
@sulkaharo

sulkaharo:
I bet OpenAPS users will be pretty confused if the midnight basals are not included? Can / should we change the daily data load to pull in data starting 23:30 the previous day?

sulkaharo:
For performance purposes, just dropping the 32 to 1 on line 621 does the trick. I have another PR that implements binary search and caching for the temp basal events, that changes the performance from o(n) to o(log n) so some excess data isn't a huge deal anymore.

MilosKozak:
well it's wrong anyway ..... if you don't specify time range ... temps are not loaded properly

mddub:
Sounds like there are two options upon removing the 32-day temp basal fetch:

* Fetch daily data for the days selected to be graphed, but consider "daily" to start a little farther backwards for the `treatments` fetch. (There would need to be a small extra filtering step before graphing, since some plotting code assumes every entry in the treatments array falls in the 24-hour period being graphed.) I suggested fetching 2 hours before midnight above, Sulka suggested half an hour before midnight; either is probably fine for strictly OpenAPS purposes.

* After fetching daily data the way we currently do, do an extra series of fetches from `treatments` for each of `[(day - 1) for day in days_to_graph if (day - 1) not in days_to_graph]`. This is so that someone can select to graph only Monday, Wednesday, Friday and still see a very long temp set at 13:00 on Tuesday. To me personally, this use case seems so unusual that it's not worth adding to this file's complexity to get it 100% correct. (As another example, graphed IOB in the reports is wrong for the first `DIA` hours of the day in such cases, since we don't also fetch the last 32 days of boluses.) It feels like the kind of thing to revisit if a user reports they actually need the feature.

Thoughts?

MilosKozak:
option 2 is more "correct". i think it could be done easy. and if we fetch data from previous day we can fix iob too

jasoncalabrese:
One other issue here is mixed timezone data, the queries consider `created_at` as a string.  OpenAPS uploads "zoned" timestamps `2016-04-20T22:59:10` vs `2016-04-20T22:59:10-07:00`.

So to be correct I think we need to select a larger window, and then convert to dates and filter again.

mddub:
The reports page has always queried against treatments' `created_at` by comparing strings with `new Date(x).toISOString()`; I was just extending that.

To be 100% correct, it sounds like we should remove the treatments-by-day queries completely, and instead fetch treatments in batches of say ~500, until the timestamp of the oldest treatment fetched is at least 24 hours before the start of the first date being graphed (in the profile timezone)? Then split them along day boundaries (using profile timezone) for the graphs?

PR #1604, opened 2016-04-18
Use 24-hour time format for minute marks in chart

ruuddotorg:
The tick marks in the bottom chart were displayed using 12-hour format for the non-whole-hour ticks, even if 24-hour format was selected in the client settings. (This only applies in the first few hours of using Nightscout, after that the bottom chart has enough time in it so there is no need to display non-whole-hour ticks anymore)

This PR fixes that by using the client settings for time format for those tick marks as well. I've verified by running Nightscout locally that they look correct now and that they are unchanged with 12-hour format. Please let me know if there is any additional testing you would like me to do.

Thank you and everyone else working on the Nightscout project for this awesome tool. #WeAreNotWaiting!


MilosKozak:
:+1:

PR #1614, opened 2016-04-22
Add colorblind-friendly color theme

ruuddotorg:
The color theme uses light green and yellow which to some colorblind
people may be too close to distinguish easily. This adds a
"colorblind-friendly colors" theme which uses light blue instead of
light green for the "in-range" color.

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -387,13 +387,13 @@ client.init = function init(serverSettings, plugins) {
   function sgvToColor(sgv) {
     var color = 'grey';
 
-    if (client.settings.theme === 'colors' || client.settings.theme === 'colorblindfriendly') {
+    if (client.settings.theme !== 'default') {
       if (sgv > client.settings.thresholds.bgHigh) {
         color = 'red';
       } else if (sgv > client.settings.thresholds.bgTargetTop) {
         color = 'yellow';
-      } else if (sgv >= client.settings.thresholds.bgTargetBottom && sgv <= client.settings.thresholds.bgTargetTop) {
-        color = client.settings.theme === 'colorblindfriendly' ? 'deepskyblue' : '#4cff00';
+      } else if (sgv >= client.settings.thresholds.bgTargetBottom && sgv <= client.settings.thresholds.bgTargetTop && client.settings.theme === 'colors') {
+        color = '#4cff00';
       } else if (sgv < client.settings.thresholds.bgLow) {
         color = 'red';
       } else if (sgv < client.settings.thresholds.bgTargetBottom) {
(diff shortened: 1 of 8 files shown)


ruuddotorg:
Note: I am partially colorblind and although I like the "color" theme, I had a hard time distinguishing between the light green "in range" color and the yellow color. Rather than changing the existing color scheme and forcing an ugly set of colors on everyone, I feel it's reasonable to add an additional option for colorblind people.

A screenshot of this theme in use: https://www.ruud.org/nightscout/nightscout-screenshot-colorblindfriendly.png

I am open to other suggestions for the color, as long as it is not a yellowish green or a greenish yellow :)

I would also respect your decision if you think that adding this option adds too much clutter or too little value to Nightscout.

lib/client/index.js
```diff
@@ -387,13 +387,13 @@ client.init = function init(serverSettings, plugins) {
   function sgvToColor(sgv) {
```
jasoncalabrese: Codacy Issue found: [This function's cyclomatic complexity is too high. (9)](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/2447185760/issues/source?bid=2452379&fileBranchId=3246919#l387)

lib/client/index.js
```diff
@@ -407,13 +407,13 @@ client.init = function init(serverSettings, plugins) {
   function sgvToColoredRange(sgv) {
```
jasoncalabrese: Codacy Issue found: [This function's cyclomatic complexity is too high. (9)](https://www.codacy.com/app/Nightscout/cgm-remote-monitor/file/2447185760/issues/source?bid=2452379&fileBranchId=3246919#l407)

jasoncalabrese:
This seems like a good way to do it, but I'm wondering if blue is a problem since we're already using some blues for the insulin hemispheres and temp basals.

For example: https://twitter.com/jasoncalabrese/status/715578211984224256

ruuddotorg:
Good point, as a colorblind person it should not be a surprise that I'm bad at picking colors.

Happy to pick something else, maybe just a blue-er shade of green, like 20B2AA or 40E0D0.

Open to other suggestions as well.

jasoncalabrese:
You know more about what colors would work than me, I just wasn't sure if you knew of some of the optional features that add more colors.  Maybe you can ask around in the FB group, gitter, etc and figure out what works best.

ruudderooij:
Good suggestion, I will play with some of the features to see what they look like color-wise and will ask for some opinions as well.

Thanks very much for the feedback!

ruuddotorg:
Good suggestion, I will play with some of the features and will see what they look like color-wise and will also ask for some opinions on the colors.

Thanks very much for your feedback!

ruuddotorg:
So after asking on the FB group and playing with the various UI features, I agree blue did not work well, and I could not find a good shade of green either. Then I thought of a trivial solution: use the same grey as in the default theme. The distinction between grey, yellow and red is quite obvious to me (and should be to most colorblind people), it does not look *too* ugly, and it should not conflict with other UI elements.

Example at https://www.ruud.org/nightscout/nightscout-screenshot-colorblindfriendly-grey.png

ruuddotorg:
@jasoncalabrese Does my last suggestion (leave the in-range color as default grey, but use yellow/red as-is) make sense? I've been running my personal site like that for a while and am quite happy with it. In case you did not notice, I did add a commit to this PR to implement that suggestion.

Please let me know if you have any further questions or concerns, and thanks again for taking the time to review this PR.

jasoncalabrese:
I didn't see you previous comment, using the default gray of good idea, nice not to be adding a new color

jasoncalabrese:
I'll give this a quick test tonight before merging

PR #1608, opened 2016-04-20
Wip/reportsperformance

sulkaharo:
Rendering two weeks of graphs on my laptop drops from over a minute to around 9 seconds with these. Using binary search for temps, deduping data, caching latest temp result for huge gain in reporting speed & loading less temp basal data. Note the new loadash version.

PR #1625, opened 2016-04-24
fix merge conflict with #1601

jasoncalabrese:
@mddub @MilosKozak @sulkaharo with #1608 the perf issue was solved, but this is still getting rid of an extra request.  Should this be in beta1?

### DIFF
--- a/static/report/js/report.js
+++ b/static/report/js/report.js
@@ -420,26 +420,24 @@
      function addPreviousDayTreatments() {
       for (var d in daystoshow) {
         if (daystoshow.hasOwnProperty(d)) {
-            if (dayscount < maxdays) {
-              var day = moment.tz(d,zone);
-              var previous = day.subtract(1,'days');
-              var formated = previous.format('YYYY-MM-DD');
-              if (!daystoshow[formated]) {
-                daystoshow[formated] = { treatmentsonly: true};
-                console.log('Adding ' + formated + ' for loading treatments');
-              }
-            }
+          var day = moment.tz(d,zone);
+          var previous = day.subtract(1,'days');
+          var formated = previous.format('YYYY-MM-DD');
+          if (!daystoshow[formated]) {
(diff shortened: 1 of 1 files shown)


sulkaharo:
Did I understand correctly that the PR breaks midnight temp basal reporting? With OpenAPS we have a temp near 100% of the time during midnight (growing kid).

jasoncalabrese:
I think @MilosKozak was going to work on a fix for that by looking up to a day back, issue would only be on the oldest midnight on the report.  But with mixed date zoning and mixed profile timezone/local timezone there are bigger issues I think.

mddub:
The issue with my change is that the first day reported doesn't show midnight temp basals, since treatments are only fetched as far back as that day. It affects midnights on other days only if a range such as "only Mondays and Thursdays for the past 4 weeks" is being reported, for the same reason.

If we want to ignore the timezone issue for now, a change like this would bring in the missing days of treatments:
https://gist.github.com/mddub/e1e09e3e04329958bfc824727a1cf669

If we want to fix the bigger issue that the client can't ever reliably query for ranges of treatments because `created_at` contains zoned ISO 8601 date strings: as I suggested in #1601, I think we should just turn the 32-day temp basal fetch into fetching all treatments in batches (of a reasonable size so that cases like reporting the last week don't hang or take unnecessarily long) until the last batch fetched contains a treatment >= 24 hours before midnight of the first day reported. This would also replace the daily treatment fetches, so it shouldn't slow things down much. Does that plan make sense?

mddub:
Also, this doesn't need to be part of the first beta, it's a nice to have because as an OpenAPS user I want to be able to generate or regenerate a week of reports without waiting 10 seconds every time I click Show.

MilosKozak:
@mddub @sulkaharo @jasoncalabrese could you test please? no extra temp basals loading now

PR #1704, opened 2016-05-01
WIP: Other Keys

jasoncalabrese:
@MilosKozak an idea for a different way to handle multiple keys without the magic single letter permission you were looking at.

For example:
`OTHER_KEYS="school family"`
`OTHER_KEYS_SCHOOL="careportal"`
`OTHER_KEYS_FAMILY="careportal edit"`

Will create a mapping like this:

```javascript
{
  careportal: [
    '655f83be7512e5b5b3ba4c9976c043ece4b3ce51'
    , 'c33f059b0ca7725fbfd6c9ea4f2f012cc7ac5a74' 
  ]
  , edit: [
    'c33f059b0ca7725fbfd6c9ea4f2f012cc7ac5a74'
  ] 
}
```

### DIFF
--- a/env.js
+++ b/env.js
@@ -35,7 +35,7 @@ function config ( ) {
   return env;
 }
 
-function setSSL() {
+function setSSL ( ) {
   env.SSL_KEY = readENV('SSL_KEY');
   env.SSL_CERT = readENV('SSL_CERT');
   env.SSL_CA = readENV('SSL_CA');
@@ -51,7 +51,7 @@ function setSSL() {
 }
 
 // A little ugly, but we don't want to read the secret into a var
-function setAPISecret() {
+function setAPISecret ( ) {
   var useSecret = (readENV('API_SECRET') && readENV('API_SECRET').length > 0);
   //TODO: should we clear API_SECRET from process env?
   env.api_secret = null;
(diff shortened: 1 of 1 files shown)


MilosKozak:
will think about impact and implementation ...
does it mean we keep the config in env? setting of rights could be clickable in some config editor ...

jasoncalabrese:
I was thinking about an editor, if we do that we'd need a new collection, etc, it would be nice.  Also with this we'd never send unhashed over the network.  With an editor we'd need to keep the unhashed keys.

I'm thinking that at some point we'll want special permissions for a specific plugins features.

MilosKozak:
well if you think the rights could became even more complicated it wouldn't continue on env ... it's even harder to maintain anyway.
storing unhashed secret in mongo is not necessary. it would not be only possible to see secret. just reset to another

jasoncalabrese:
Maybe a display label and key, and not allow them to match

MilosKozak:
yeah with editor it can be easy username + password

jasoncalabrese:
The computer in the health office at my son's school clears local storage and cookies every night, think I'll need to be able to have a token as a url param.

Issue #1713, opened 2016-05-03
v0.9.0-beta2: Store "Show AR2 Forecast" selection


chadmmartin:
This is just a suggestion/ask to store the 'Show AR2 Forecast' selection so it's not always reset to selected when refreshing or closing/opening Nightscout forcing the user to unselect it each and every time.

PR #1728, opened 2016-05-08
save focus hours and render selections to local storage

jasoncalabrese:


jasoncalabrese:
will resolve #1713 and also save the focus hours selection

PR #1730, opened 2016-05-10
Padding on hamburger button

rkresha:
On hover, scroll bars appear. Adjust from 10px to 15px on padding-right


jasoncalabrese:
Did you still get the scrollbars on dev without this?  If so what browser.  I do remember seeing that bug but thought it was fixed along with some other changes.

rkresha:
Yes with chrome (multiple versions). Forked from dev two days ago so I was working off a current version of the branch. Didn't try any other browsers. Still got it with 13px, so decided to round up to 15 and seems fine now.

rkresha:
The treatment bubbles causes the scrollbars to come up as well, so my adjustment is only half of the battle. Fix will need to address bubble pop ups

MilosKozak:
send screenshot what it does

rkresha:
image

rkresha:
Hovering on another comment a little older shows as expected, no scroll bars
image

rkresha:
Hovering over the hamburger button yields similar results. Perhaps this PR closes and we open it as an issue because this is a work around for the settings button not freaking out not for the treatment bubble issue

rkresha:
Closing this PR in favor of adding it as an issue. #1788

PR #1753, opened 2016-05-22
Show treatments IOB along with devicestatus IOB

MilosKozak:
image
@jasoncalabrese if you agree ...


jasoncalabrese:
I think that's ok, I was wondering about calling it Bolus IOB, but that could be confusing in other ways

PR #1765, opened 2016-05-26
Https/GitHub.com/nightscout/cgm remote monitor/tree/dev

LynneRB:


LynneRB:
NP  just can’t figure why my pump is not showing, and OpenAps says unknown.  I have 9.0.2 with the pill boxes, etc., but maybe I need the tree/dev branch Thanks

PR #1781, opened 2016-05-29
Change defaults for azure deployments

MilosKozak:
@jasoncalabrese agree?


jasoncalabrese:
Might still be enabling too much by default, but better than turning pump and openaps on by default before

PR #1807, opened 2016-06-09
Korean translation update

jungsomyeong:
Korean translation update


MilosKozak:
stop doing another PRs an contact me first

PR #1809, opened 2016-06-09
Changed the data loader to pull less data to reduce load on Mongo

sulkaharo:
@jasoncalabrese see; reduces page load size HUGELY for OpenAPS users

### DIFF
--- a/lib/api/treatments/index.js
+++ b/lib/api/treatments/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var _ = require('lodash');
 var consts = require('../../constants');
 
 function configure (app, wares, ctx) {
@@ -20,6 +21,7 @@ function configure (app, wares, ctx) {
   // List treatments available
   api.get('/treatments/', function(req, res) {
     ctx.treatments.list(req.query, function (err, results) {
+      _.forEach(results, function clean(t) { t.carbs = Number(t.carbs); t.insulin = Number(t.insulin); });
       return res.json(results);
     });
   });
--- a/lib/client/receiveddata.js
+++ b/lib/client/receiveddata.js
@@ -2,7 +2,11 @@
 
 var _ = require('lodash');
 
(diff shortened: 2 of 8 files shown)


MilosKozak:
profile switch could be much older than 2 months

MilosKozak:
don't understand this change https://github.com/nightscout/cgm-remote-monitor/pull/1809/files#diff-5735b457d742cb24034018dd8a1ad4e8L183

sulkaharo:
Is there a better way to load the profiles than loading everything from last 12+ months?

The change from "just load a few events from a longer time period" to "load everything except temp basals" is based on two things: 1) the change above where we don't load everything from last 8 days anymore and 2) the only datapoint in treatments that's potentially got a _lot_ of data being OpenAPS issued temp treatments. So loading everything from last 2 days and everything except the potentially large data spam event from 21 ends up being not a huge change for most uses and a huge data save for OpenAPS users.

MilosKozak:
ad profile switch: the only possibilities are 
load 12 months or so
or load 12 months first time only, cache it and then expect people will not add such records too much to the past and load profileswitch along with other treatments and merge it

MilosKozak:
ad other treatments:
sorry i didn't notice $in -> $nin :)
got it now

sulkaharo:
Can we load the last profile change object regardless of age? What's the data structure like?

jasoncalabrese:
Can we sort and limit to 1, then depending on if we get a result and how far back it is load more

MilosKozak:
there is a index of eventtype and createdAt. guessing query of limit(1) and some time limit will take about the time

MilosKozak:
so far we use only newest profile record i think

jasoncalabrese:
what's the state of this, is it something we'd want to get in the release?

jasoncalabrese:
With the new tokens I think we can add some settings to the subjects to adjust the amount of data returned.

sulkaharo:
Closing in favor of #2068

PR #1850, opened 2016-07-10
node engine bump

ELUTE:
Bumped the Node engine version to 4.x to resolve deployment failures with mlab.


jwedding:
+1 This fixes a problem with new deployments failing in ALL environments.

jasoncalabrese:
from the message posted to gitter it looked like there was another option (pinning the mongo version) than bumping the node version which is a big big change and I'm sure will have other effects.

I'm going to give that a try, can't merge to master when the build is failing

jasoncalabrese:
see #1852 for the final solution, thanks for tracking down the issue @ELUTE

PR #1844, opened 2016-07-01
Change formatting in OpenAPS pill

cjo20:
Change the formatting to make it slightly easier to read. New style looks like this:
image

PR #1856, opened 2016-07-17
Change formatting in OpenAPS Pill v2

cjo20:
Resubmitting this from a new branch, old one had a silly number of commits that were used for testing on my azure instance (I really need to set up a copy I can run locally)

Replaces https://github.com/nightscout/cgm-remote-monitor/pull/1844

### DIFF
--- a/lib/plugins/openaps.js
+++ b/lib/plugins/openaps.js
@@ -283,28 +283,54 @@ function init() {
       if (prop.lastSuggested) {
 
         var valueParts = [
-          valueString('BG: ', prop.lastSuggested.bg)
-          , valueString(', ', prop.lastSuggested.reason)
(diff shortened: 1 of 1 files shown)


cjo20:
image
This is how enacted basals currently look.

jasoncalabrese:
Sorry I haven't got back to you on this, moving BG to the bottom seems a little out of place and the extra lines feel tall when there are multiple rigs.

I think we need to be able to use more structure in the info tip, then we could use multiple font sizes or something.  I liked the indenting you had before, but we probably need to add some css support to make it wrap right.

sulkaharo:
Apologies for this having been opens for so long; can't effectively merge anymore. Please PR in again if you still think this is valid.

PR #1859, opened 2016-07-21
Rus translation update

apanasef:


apanasef:
Milos, I just added my translation of a few lines that have appeared recently

PR #1874, opened 2016-08-07
import many settings on boot from remote URL

bewest:
Allows sharing commonly used configurations amongst folks.
CC @jasoncalabrese.

### DIFF
--- a/env.js
+++ b/env.js
@@ -17,6 +17,7 @@ function config ( ) {
    */
   env.DISPLAY_UNITS = readENV('DISPLAY_UNITS', 'mg/dl');
   env.PORT = readENV('PORT', 1337);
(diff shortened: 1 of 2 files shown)


bewest:
Here's what I used to test against: https://gist.githubusercontent.com/bewest/34fdf7f628e07f6446064a66e2b30d33/raw/37ec006e6a92321212f64adf797b7ee2f9b9334b/new-defaults.json

It's simply the output of `curl -s $host/api/v1/status.json | json settings`.  This only changes the `env.settings` object, none of the others.

Am thinking about supporting a `defaultRoles` feature in the imported config, which would allow providing a completely custom table of builtin permissions.

jasoncalabrese:
Maybe we should only use the method when it's connected to a subject/token, then external settings will only affect clients and not the server

PR #1885, opened 2016-08-14
Wip/audio experiments

bewest:
## is cgm for the sighted alone?
Etude in glucose.


bewest:
See video, hear sound: https://twitter.com/bewestisdoing/status/764645661430034432

I've had some really great conversations with blind users of cgm.  Basically, using cgm without sight is hard: by default audio reports are given on the 5 minute "tick" updates, and no other times, and additionally there is no way request a report on demand, nor is there a good way to get an idea of trend.

ATM, this experiment sends all the glucose from the last two days (the default/hardcoded in NS) and converts them all into 16th notes in 3/2 time at 220 bpm.  The entire audio is done in about thirty seconds.  The only button, with `tabindex=1` is an "again" button which will replay the audio.

I'm not at all confident this is right, but my first-hand usage is promising already.  While I personally have a music degree, and lots of a Capella singing, I've already learned to orient myself in time listening to the updated announcements, as well as get a sense of overall reading.  I suspect with some guiding/orienting tones, in addition to diversifying the instrument/sounds (especially WRT missing or N/A sgv's and converting them to a percussive/snap sound), as well as limiting and duration to relevant rations, that this could benefit anyone who cannot look at cgm graph, regardless of the reason.  I'm hopeful that others will contribute their use cases.

bewest:
The "two days in 30 seconds" seems like the wrong ratio, but I'm not sure what is the right ratio, and whether or not there needs to be a way to select more than one "on demand" (what if app is closed and you want to open it and get a readout?).

PR #1896, opened 2016-08-19
Loop plugin

ps2:
This plugin supports the (Loop iOS app)[https://github.com/loudnate/Loop].  I first attempted to support Loop via the openaps plugin by making Loop post the same data structures to the NS api, but there are enough differences in how Loop collects, processes, and acts upon data, that reflecting these states accurately in the openaps plugin is problematic.   I could solve it by submitting a patch to the openaps plugin that adds conditionals at the key points, but eventually decided that it would be better to have two separate plugins that focus clearly on communicating what is going on with it's own specific system.


tests/loop.test.js
```diff
@@ -0,0 +1,237 @@
...
+           'timestamp':'2016-08-13T20:09:15Z',
+           'rate':0.875,
+           'duration':30,
+           'received':true
+        },
+        'version':'0.9.1',
+        'recommendedBolus':0,
+        'timestamp':'2016-08-13T20:09:15Z',
+        'predicted':{
+           'startDate':'2016-08-13T20:03:47Z',
```
jasoncalabrese: I like `startDate` better than the assumption we have with oref0

jasoncalabrese:
Looks good to me, similar structure as the openaps plugin.  I ran the build again from travis and got it to pass.  Going to merge since there are several people already using it and is isolated.  We can make any adjustments against dev later.

PR #1900, opened 2016-08-22
Changed the count of records coming back from API call

daveewall:
Personal experience shows over 70,000 records returning on some days.

Before:

before

After:

after

### DIFF
--- a/static/report/js/report.js
+++ b/static/report/js/report.js
@@ -558,7 +558,7 @@
         return $.Deferred().resolve();
       }
       $('#info-' + day).html('<b>'+translate('Loading CGM data of')+' '+day+' ...</b>');
-      var query = '?find[date][$gte]='+from+'&find[date][$lt]='+to+'&count=10000';
+      var query = '?find[date][$gte]='+from+'&find[date][$lt]='+to+'&count=100000';
       return $.ajax('/api/v1/entries.json'+query, {
         success: function (xhr) {
           xhr.forEach(function (element) {


sulkaharo:
That's odd. Do you mean you have up to 70k records on a single day? That's almost one measure / second, which can't possibly be right. Do you have an uploader device that's somehow causing a massive amount of data duplication? A single day using Elite or Dexcom is expected to contain exactly 288 CGM readings (one every 5 minutes).

sulkaharo:
Are you uploading CGM data directly into Mongo, or going through the REST api?

daveewall:
On a single day, yes.  I put a breakpoint on the success handler of that call and saw that many records coming back each day.  Sometimes 20000, sometimes up to 70000.  But always more than 10000.

I'm going through the REST API with the ns-upload commands, just like the docs say, but I've had this issue ever since going to an Intel Edison, but I'm not sure that's the reason.  I've rebuilt the Edison multiple times, starting on Ubilinux, now on Yocto.  Maybe there is something screwy on my side, which would explain why this wasn't an issue before, but no idea what may be causing it, and little time to figure it out.

If you have any suggestions, let me know, and I'll see what I can do to look into it, but it may take a while before I can get to it (new child and all...).  Thanks.

daveewall:
I should add that I tried creating a new database to see if the issue was data related, but on day one of the new database, it kept happening.  So I guess that does point to something in my config...?  Not sure what without investigating more, though.

sulkaharo:
You mean you're uploading CGM data from the Edison into Nightscout, not using an Android app? Do you have logic in place to only upload new records? If you're creating this many records into the Mongo database, it'll probably stop scaling for you pretty quickly.

daveewall:
I'm on a Medtronic-only system, no Dexcom.  No Android phone used.  Using the Edison to directly upload to the web site via wifi or tether to iPhone.  I assumed that the API would be the system that does the de-duplication, because the loop requests 24-hours of data (or glucose values for AMA) during every loop.  You mean that the API is storiing all 24 hours every loop run?

PR #1906, opened 2016-08-23
Show last predicted, even if lastLoop doesn't have a prediction

ps2:
If we have Loop errors to report, but no updated prediction, the graph was redrawing to drop the forecast area. In other cases (like if Loop was offline) it would keep drawing the old forecast as new BG points came in.  This updates the loop plugin to keep the old forecast onscreen until we have an updated one.

### DIFF
--- a/lib/plugins/loop.js
+++ b/lib/plugins/loop.js
@@ -87,6 +87,7 @@ function init() {
     var result = {
       lastLoop: null
       , lastEnacted: null
+      , lastPredicted: null
       , lastOkMoment: null
     };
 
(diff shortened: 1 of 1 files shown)


ps2:
What do I need to do to get this merged?  The travis build has this error:

```
tick Tue Aug 23 2016 21:17:06 GMT+0000 (UTC)
not ok 2 Entries REST api "before all" hook
  Error: timeout of 10000ms exceeded
      at null.<anonymous> (/home/travis/build/nightscout/cgm-remote-monitor/node_modules/mocha/lib/runnable.js:139:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:121:15)
got data-received event, requesting reload
Load Complete:
	 sgvs:1
Reusing MongoDB connection handler
```

Which looks like an unrelated error.  Maybe mongodb timeout?

jasoncalabrese:
I didn't see this till now, I restarted a build and will get it merged

PR #1919, opened 2016-09-01
Multiple CGM Sources using bucketed data

jasoncalabrese:


jasoncalabrese:
Do we need to extend the pebble endpoint to use the buckets for more than delta?

scottleibrand:
Just switched my main Nightscout to this branch.  Looks good so far.

jasoncalabrese:
Great, it's been working good for me too.

I need to add some tests and also think I want to adjust the name/structure of the property.  Currently we have `/api/v2/properties/bgnow`, I might have 2 properties, one just for the delta.  When I first started on this I was going show the average of the last bucket, but after playing with it I think it makes more sense to show the last actual value.  So the BGNow name doesn't make sense anymore.

PieterGit:
can't find any regression with one receiver, running for more than two weeks on this branch now

PR #1926, opened 2016-09-04
Update language.js

martijnmcb:
Minor update in dutch language.


jasoncalabrese:
There a conflict with this, might be easier to make a new PR since the change is so small.

martijnmcb:
Made a new pull req but seems like conflict still exists.  GitHub is for me a magic blackbox, have no idea how te reolve this. Change is very small and not critical.

PR #1944, opened 2016-09-10
support scanning and removing more than 10 future records, and bind to ipv6 socket

scottleibrand:


server.js
```diff
@@ -46,7 +46,7 @@ function create (app) {
 
 require('./lib/bootevent')(env).boot(function booted (ctx) {
     var app = require('./app')(env, ctx);
-    var server = create(app).listen(PORT);
+    var server = create(app).listen(PORT, '::');
```
jasoncalabrese: did you mean to include this?  I'm not sure what the behaviour will be in all cases environments

scottleibrand: Yes, the behavior without it was broken for IPv6, so it seemed a worthwhile change.  I'm happy to have it reverted while awaiting more testing if needed though.

jasoncalabrese:
@scottleibrand I added a new `HOSTNAME` env setting that you can set to `::`, can you give it a try and see if it works for you

PR #1975, opened 2016-09-26
Update for BG levels

martywright:


jasoncalabrese:

PR #1992, opened 2016-09-28
0.8.5 pull

rnewby:


jasoncalabrese:

PR #1996, opened 2016-09-29
removing mongodb3.x dependencies

SandraK82:
- remove usage of $eq operator

enables the usage of local mongodb2.6 on raspi ( https://github.com/SandraK82/deploy-ns-local-raspi )

### DIFF
--- a/lib/data/dataloader.js
+++ b/lib/data/dataloader.js
@@ -144,9 +144,7 @@ function loadTreatments (ddata, ctx, callback) {
 function loadProfileSwitchTreatments (ddata, ctx, callback) {
   var tq = {
     find: {
-      eventType: {
-        $eq: 'Profile Switch'
-      }
+      eventType: 'Profile Switch'
       , created_at: {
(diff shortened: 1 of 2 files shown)


jasoncalabrese:
I'm not sure we want to commit to supporting mongo 2, but these changes seem like a good simplification anyway.

Why can the pi run mongo 3?

jasoncalabrese:
Saw this, but does seem like a bit of a pain http://andyfelong.com/2016/01/mongodb-3-0-9-binaries-for-raspberry-pi-2-jessie/

SandraK82:
I was not able to get one of the provided binaries to run stable on my pi. Building failed for different reasons, too. Some other folks at the Germanloop group had similar problems. So I decided to analyse the needed change set to get it running...

jasoncalabrese:
I'll test this over the weekend and get it merged, but don't want to promise mongo2 support for everything going forward, but when possible we should try

SandraK82:
As long as we get no 64bit raspi3 Linux I will try to maintain the compatibility. As soon as I see non working parts I will try to fix them and simply give you another pr

PR #1999, opened 2016-09-30
updating branch from nightscout

gcastellanos:


jasoncalabrese:

Issue #1788, opened 2016-05-31
Scrollbars appear when hovering over recent treatment details and settings button 'hamburger button'


rkresha:
Copying this over from #1730 since it's a larger problem. Seeing this with 0.9.0-beta3 and dev branches

Seeing this in Chrome, Internet Explorer, and Firefox

Broken: when hovering over a recent treatment or hamburger button
![image](https://cloud.githubusercontent.com/assets/19247761/15683306/40b3a95e-2727-11e6-9dbb-bcb562cfed95.png)
![image](https://cloud.githubusercontent.com/assets/19247761/15683370/902f2ca6-2727-11e6-9dd2-cb6dc95b739b.png)

Works: when hovering over an older treatment
![image](https://cloud.githubusercontent.com/assets/19247761/15683336/6503210e-2727-11e6-9662-3ebc5895e296.png)

rkresha:
Working on this... just testing the changes to make sure they truly do fix both cases. More to come in the next few days.

rkresha:
The settings/hamburger button is good now. The treatment tooltip is still being a pain

PieterGit:
I sometimes reproduce this with the settings/hamburger button (with 0.9.0-dev-20161010, head e0d0a98). Tested with Firefox/Edge. 

I can easily reproduce this with a unauthenticated browser (I tried Edge). It seems the tooltip "settings" triggers the invalid horizontal scrolling bar. Please fix this for the final 0.9 version.

rkresha:
Treatment hover still not adjusted, but it not nearly as noticeable. Hamburger button certainly fixed imho.

rkresha:
@jasoncalabrese So if I hover over the hamburger button, the page resizes fixing the scrollbar issues somewhere between 15-25 seconds after hovering. I move the mouse away and the page resizes again within 5 - 15 seconds and brings the scrollbar back... does that help at all?

rkresha:
Fixed with #2139

PR #2069, opened 2016-10-11
Scroll bar hover issue

rkresha:
For issue #1788


jasoncalabrese:
I tried this out, but it assumes the hamburger menu is last, but careportal and bolus wizard are after if enabled

screenshot 2016-10-11 17 44 16

With dev I don't see the scrollbar in chrome when hovering:
screenshot 2016-10-11 17 48 26

rkresha:
What if you disable careportal and bolus wizard in dev. Then do you
experience the issue?

jasoncalabrese:
Yeah, no scrollbar even with careportal and boluswizard disabled, using the current Chrome 53.0.2785.143 (64-bit) on a mac

screenshot 2016-10-12 09 39 15

rkresha:
I'm on windows 10 64bit (chrome 53.0.2785.143 m) and have a vanilla branch that is a up-to-date clone of nightscout/dev and I still see the issue. I've also tried Firefox/Edge/IE 11 with the same behavior

image

rkresha:
My workaround, which was terrible, was to offset the icon so the underlying balloon text would be more offset to the left. I think the fix is to just move the balloon text to the left and leave the icon alone, but I'd need some guidance on where to adjust it.

rkresha:
You can leave this open, but I'll continue to look into the tipsy side of the house.

PR #2081, opened 2016-10-14
selin

gokhan84:


jasoncalabrese:
You might not have been logged into github when you started the update, try using http://nightscout.github.io/pages/update-fork/ again

PR #2107, opened 2016-10-16
update german parts

baembel08:
hopefully its right :)


MilosKozak:
Line 7506 dot instead of colon

PR #2139, opened 2016-10-21
Add right padding to button bar to fix scrollbar issue

rkresha:
Windows users may experience scollbars being displayed when hovering
over the right most icon on the button bar. This adds right padding to
kick the buttons over to the left more.

This is for issue #1788

### DIFF
--- a/static/css/drawer.css
+++ b/static/css/drawer.css
@@ -227,7 +227,7 @@ h1, legend,
   overflow: hidden;
 }
 #buttonbar {
-  padding-right: 10px;
+  padding-right: 15px;
   height: 44px;
   opacity: 0.75;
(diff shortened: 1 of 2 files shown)


rkresha:
@jasoncalabrese adjusted the button bar this time around instead of focusing on the hamburger button. Now working fine imo.

jasoncalabrese:
Instead of setting that in the html, it should be in the css, it already has `padding-right: 10px;`, so just change that to `15px`

https://github.com/nightscout/cgm-remote-monitor/blob/dev/static/css/drawer.css#L230

rkresha:
I should've checked, my mistake. Thanks for the spot check.

PR #2140, opened 2016-10-21
get plugins from ctx instead of initializing them again

jasoncalabrese:


sulkaharo:
Great improvement, probably saves a ton of memory / pebble invoke

PR #2175, opened 2016-10-31
minify bundle.js

moomoobloo:
Minify bundle.js to improve page load latency.

According to Pingdom speed test, this reduces download size by about 800kB, and improves page load latency by about 500ms.


sulkaharo:
Works for me on Safari & Chrome on OS X, iOS

PR #2218, opened 2016-11-14
3rd Portuguese Update

dougtakeuti:


lib/language.js
```diff
@@ -5929,6 +5936,7 @@ function init() {
       ,bg: 'Актуализирай'
       ,it: 'Aggiornamento'
       ,pl: 'Aktualizacja'
+      ,pt: 'Atualizar'
```
jasoncalabrese: @dougtakeuti build is failing, looks like it's because of the duplicate here

dougtakeuti:
Thanks, I just fixed it.

2016-11-14 14:54 GMT-02:00 Jason Calabrese <notifications@github.com>:

lib/language.js
```diff
@@ -6547,6 +6592,7 @@ function init() {
       ,nb: 'Kunngjøring'
       ,fi: 'Tiedoite'
       ,pl: 'Powiadomienie'
+      ,pt: 'Aviso'
```
jasoncalabrese: @dougtakeuti that exposed another duplicate here, this file is way too big we need to split it up

PR #2246, opened 2016-11-22
Prioritize Loop IOB over Pump IOB if both are available

dm61:
Updated iob.js per @mddub's suggestion, to prioritize Loop IOB if available


jasoncalabrese:
Thanks, I'll test the openaps case and get this merged

PR #2247, opened 2016-11-22
0.9.1 (Grilled Cheese)

jasoncalabrese:
Changes since 0.9.0
===============

  * Alexa integration
  * Lots of translation updates including plugins
  * Refactor storage and support using static openaps report files in addition to the standard mongo backend
  * New aggregation queries for example `/api/v1/count/entries/where?find[sgv]=100`
  * Import settings via url
  * Prioritize Loop IOB over Pump IOB if both are available

see https://github.com/nightscout/cgm-remote-monitor/milestone/7?closed=1 for a complete list


MilosKozak:
👍

PieterGit:
Looks good afaict. If you want to follow semver 2.0.0 rules. I think this version should be called 0.10.0 instead of 0.9.1 because there a two api improvents:
- New aggregation queries for example /api/v1/count/entries/where?find[sgv]=100
- Import settings via url

According to semver.org "MINOR version when you add functionality in a backwards-compatible manner."

Thanks for the great work on the openaps-storage!

jasoncalabrese:
Good point about semver, so far each 0.x release has been a new name. Maybe 1.0 will be the time to start following semver.

PieterGit:
Why wait until 1.0? Why not release 0.9.1 as 0.10.0 and explain in the release notes that from now on you will follow semver convention. I think it would be good to do that not only for nightscout, but also for openaps/oref0, etc. 

Because of the number of users of Nightscout and because current Nightscout is quite stable, I would really release a 1.0.0 before the end of the year (with maybe some more focus on bug fixing instead of adding features).

Does this release require node 4.6.0+ or does it now supports both 0.10+? 
That might be good to mention in the release notes / facebook post, etc. In case you require a new node version, I definely would increate the MAJOR or MINOR version part in this release.

jasoncalabrese:
This release doesn't include the dependency updates, there seems to be some issues with it on azure, hoping others that use azure will help out with that.

PR #2267, opened 2016-11-26
Alexa mmol conversion from dexterbooty

jasoncalabrese:


jasoncalabrese:
@dexterbooty I pulled you're commit in to this, I'm going to add a test too, thanks for your help

jasoncalabrese:
I'm going to merge this, we still need to add some tests, but it's going to be a little bit of work and I don't want to hold this up.

We also still need to update the responses for the AR2, OpenAPS, and Loop plugins

PR #2278, opened 2016-11-30
Alexa only needs TWO decimal points for IoB 

shanselman:
Putting ToFixed in Alexa's IoB message otherwise she'll say:

`      "text": "123 and slightly dropping as of 5 minutes ago. Your temp basal of 0 units per hour will end in 16 minutes and you have 3.415892342031398 units of insulin on board."
`

Sorry for the messy commits, you should be able to use GitHub's new feature to SQUASH AND MERGE


jasoncalabrese:
There's a failing test, just need to change it to expect 1.5 instead of 1.50, I can do it later if you don't get to it before me

jasoncalabrese:
replaced by #2279

PR #2283, opened 2016-12-02
Fix activityContrib

warstar2187:
This fixes an incorrect insulin activity calculation for DIA <> 3h, which would result in negative insulin activity (and positive BGI) at the tail end of the DIA for DIA < 3h, and erroneously high activity (and too-low BGI) for DIA > 3h. Discussion at https://gitter.im/MilosKozak/AndroidAPS?at=583cd6c116207f7b0eb97fa2

Before fix.
image

After fix.
image


jasoncalabrese:
Thanks for tracking this bug down in all the various repos

PR #2318, opened 2016-12-09
Jordan berger patch 1

jordan-berger:


jordan-berger:
Possible wifi script
[wifi.txt](https://github.com/nightscout/cgm-remote-monitor/files/641269/wifi.txt)

PR #2328, opened 2016-12-11
Show extra info when there are multiple devices upload cgm data

jasoncalabrese:


jasoncalabrese:
screenshot 2016-12-11 10 02 36

PR #2330, opened 2016-12-12
Reduced the time processDurations goes through treatments

beached:
In my benchmarking it went from an average of 16.0s for 90days of data to 2.4s.
Also created a single function for the conversions of msecs to mins and visa-versa

### DIFF
--- a/lib/times.js
+++ b/lib/times.js
@@ -40,7 +40,9 @@ function create (types) {
 }
 
 var times = {
-  week: function ( ) { return create('weeks')(1); }
+	mins_msecs: function (value) { return value*60*1000; }
+  , msecs_mins: function (value) { return value/1000/60; }
+  , week: function ( ) { return create('weeks')(1); }
   , weeks: function (value) { return create('weeks')(value); }
   , day: function ( ) { return create('days')(1); }
   , days: function (value) { return create('days')(value); }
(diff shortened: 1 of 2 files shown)


lib/times.js
```diff
@@ -40,7 +40,9 @@ function create (types) {
 }
 
 var times = {
-  week: function ( ) { return create('weeks')(1); }
+	mins_msecs: function (value) { return value*60*1000; }
```
jasoncalabrese: Maybe we can come up with better names for `mins_msecs` and `msecs_mins`, so it will be easier to remember what they do and will prevent the from getting used backwards

beached: Yeah, I when I wrote it I was just seeing how much of a diff it would make in the profiler.
What about timeMinsToMsecs  and timeMsecsToMins ?

jasoncalabrese: Maybe just `minsToMsecs` and `msecsToMins`

beached: I have it done in my daw_001 branch but I am unsure how to get it into this one.

jasoncalabrese:
Not sure why it says unknown repository, maybe wasn't pushed to your fork, but I can do it locally when I test

beached:
I accidentally uploaded my database configuration and pulled the branch

jasoncalabrese:
Hmm, just tested and openaps temp basals aren't rendering at all, are they rendering for you?

PR #2331, opened 2016-12-12
Fixed reference of undefined error

beached:
Was seeing errors in js console about delta being undefined.  Put check in


jasoncalabrese:
thanks, I just wrote that bug today and already fixed :)

beached:
It was in my way :)  Couldn't see my logging

beached:
So it is expected to have undefined there?

PR #2332, opened 2016-12-12
Removed unused CSS

beached:


jasoncalabrese:
I thought tipsy was used for the tooltips on the top bar, but it's been there for a long time and never confirmed

beached:
Chrome is complaining to me that it is unreferenced.  This is the css, not the js

jasoncalabrese:
The `tipsy.css` is needed, without it hovering over the icon's ads a line on top instead of the tooltip

screenshot 2016-12-11 22 35 19

Also `ui-darkness/jquery-ui.min.css` is used for the login and edit dialog, without it you get this (no modal background/border/title/etc):
screenshot 2016-12-11 22 39 43

PR #2333, opened 2016-12-12
Added proper display of metre bg reads

beached:
The main client window did not scale bg reads to reflect client's unit choice


beached:
Found a new bug, will fix first

beached:
Bug was already there

lib/client/renderer.js
```diff
@@ -146,7 +146,7 @@ function init (client, d3) {
       var rawbgInfo = getRawbgInfo();
 
       client.tooltip.transition().duration(TOOLTIP_TRANS_MS).style('opacity', .9);
-      client.tooltip.html('<strong>' + translate('BG')+ ':</strong> ' + client.sbx.scaleEntry(d) +
```
jasoncalabrese: good find

PR #2334, opened 2016-12-12
Bugfix, tooltip was displaying NaN for Bg value

beached:


jasoncalabrese:
I think you fixed this with #2333, and I think that's better than assuming `scaled` will already be there.

PR #2336, opened 2016-12-12
Unused CSS

beached:
Could not find a report that used this css file


jasoncalabrese:
Same as #2332 the `ui-lightness/jquery-ui.min.css` is used for the edit modal, maybe chrome doesn't see it as being used till the modal is opened

PR #2338, opened 2016-12-12
Updated forever npm dependency

beached:
forever pulls in minimatch and the old dependency has a DoS bug.  This version pulls in minimatch 3.0.3.  
https://github.com/nightscout/cgm-remote-monitor/issues/2228

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "expand-braces": "^0.1.1",
     "express": "^4.6.1",
     "express-extension-to-accept": "0.0.2",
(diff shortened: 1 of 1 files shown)


jasoncalabrese:
The Procfile used by Heroku uses forever, we should test there to make sure it works as normal

see: https://github.com/nightscout/cgm-remote-monitor/blob/master/Procfile

sulkaharo:
I think this is (finally) resolved with the recent module updates..

PR #2353, opened 2016-12-16
Implement Azure push notifications for Hybrid App!

srmoss:


env.js
```diff
@@ -113,6 +113,7 @@ function setStorage() {
   env.profile_collection = readENV('MONGO_PROFILE_COLLECTION', 'profile');
   env.devicestatus_collection = readENV('MONGO_DEVICESTATUS_COLLECTION', 'devicestatus');
   env.food_collection = readENV('MONGO_FOOD_COLLECTION', 'food');
+  env.azurepush_collection = readENV('MONGO_AZUREPUSH_COLLECTION', 'azurepush');
```
jasoncalabrese: I've been wanted a generic notifications collection, can we rename this to `notifications` instead of `azurepush` then later we can do some refactoring to be able to use it for other notification integrations.  For example the pushover receipts are currently only stored in memory, would be nice to be able to add them to a collection to survive a restart.

lib/plugins/azurepush.js
```diff
@@ -0,0 +1,323 @@
...
+            tags: tags
+          }
+        }
+      };
+
+      // 'platform': https://developer.xamarin.com/guides/xamarin-forms/platform-features/device/#Device.OS
+      switch (request.platform) {
+        case 'iOS':
+          installation.platform = 'apns';
+          installation.templates.nightscout.body = "{\"aps\":{\"alert\":\"$(message)\",\"sound\":\"$(sound)\",\"category\":\"$(category)\"},\"eventName\":\"$(eventName)\",\"group\":\"$(group)\",\"key\":\"$(key)\",\"level\":\"$(level)\",\"title\":\"$(title)\"}";
```
jasoncalabrese: if you use single quotes you could avoid all the escaping, maybe better to build as a json object and then convert to string?

env.js
srmoss: Right now, this collection stores device registrations, not notifications themselves.  The concept is similar to what we have in static environment variables for pushover, but they are stored in mongo since device registration is dynamic.

I can see two approaches to this:
1. Have a single collection for `notifications` that includes at it's top level devices and settings (what I currently store) and then has an array of actual notifications sent/receipts.
2. Use two separate collections: one for `devices` (what I currently store) and another for `notifications` which can have a key to the devices it was sent to/received by.

Thoughts?

lib/plugins/azurepush.js
srmoss: I think I can change the outer most quotes to single quotes, but can't remove the escaping or use stringify on a JSON object.  

The string for `body` is actually used in Azure to translate an actual JSON object sent in a notification (from Nightscout, which isn't stringify'd or build as a string) to the JSON payload sent to the messaging platform.  In otherwords, I'm not building a JSON object for `body` so much as I'm building a map for how Azure will translate my actual payload into a platform specific notification.

I'll at least update the outer most quotes to single quotes to keep with coding style.

env.js
jasoncalabrese: I see, was thinking the registrations were for each push, but are for the phones running the apps.  A `devices` collection would be confusing with `devicestatus`, maybe we could call it `clients`, but it is more special purpose than I was thinking.  Just don't want each new integration adding a collection if we don't need to.

lib/plugins/azurepush.js
jasoncalabrese: Thanks, thought it might be something like that, but wanted to check

env.js
srmoss: How about `pushclients` to avoid confusion with actual clients on Nightscout (used by socket.io)?

lib/plugins/azurepush.js
```diff
@@ -0,0 +1,323 @@
...
+      logTrelloIssue(env, '0x12-Register received without platform');
+      // END: *****REMOVE*****
+
+      // No platform, required parameter
+      return callback({message: 'Platform required'});
+    }
+  };
+
+  azurepush.ack = function ack(request, callback) {
+    if (request && request.level) {
```
jasoncalabrese: Even if there isn't a key for the notification, maybe the device id could be checked?

srmoss: I thought about this quite a bit... and was uneasy about the openness of an API like this too.  You'll also note that there was a separate existing API that required the caller be authenticated with an API_SECRET.

The problem is we don't currently have a way to uniquely identify notifications beyond the group and level.  The key is currently being sent in both directions, but not being used since it's not unique to every notification.  It could be used and checked against the notify table, but it's not very secure (since some notification keys are as simple as ar2_1/ar2_2).
... As a side note, is it true that snoozing is currently open to any unauthenticated client on the website?

Additionally, there is currently no requirement to have an API_SECRET input on the client to use notifications (nor do I think there should be).

One thing we could do is use a unique _id generated to store notifications in mongo (which we currently don't do, but want to do at some point) that would allow us to easily persist a uuid and check on acknowledgement.  That wouldn't be too difficult.

env.js
srmoss: It will be difficult to change in the future because we'll need a conversion.  I agree it should be generic and it's a much easier change at this stage.  I've got the change made and tested on my local branch, so I'm happy to sync it.

lib/plugins/azurepush.js
jasoncalabrese: Yeah, currently snoozing is open, and I really want to fix that now that we have a permissions system using the tokens.  I started to implement `notification:*.ack`, but it's not done.  A notification id would be great, I think that's the direction we need to go.  Also I think there's a difference between a local snooze and a system wide ack.  Anyone should be able to make the alarm they are hearing locally stop, but there should be a limit on how that effects others using the system.

If we can't find a way to lock it down more now we should only add the endpoint if the plugin is enabled so we don't add a new hole by default.

srmoss: We don't have a way to differentiate between system wide and local yet, right?  I only saw the ack function which does the same thing as snoozing in the website.

lib/plugins/azurepush.js
srmoss: Worked like a charm.  From Azure registration:
``BodyTemplate: '{"data":{"message":"$(message)","eventName":"$(eventName)","group":"$(group)","key":"$(key)","level":"$(level)","sound":"$(sound)","title":"$(title)"}}'`

env.js
jasoncalabrese: OK, agree it will be harder to change later so makes sense to try making it more generic now

lib/plugins/azurepush.js
srmoss: Also, currently this API is only available IF azurepush is in the ENABLE string AND the user had the BASE_URL configured.

jasoncalabrese: There isn't a way yet, but the main difference would be that we just wouldn't send the ack if not authorized, but still snooze the alarm locally.

I'm wondering if there's a way to use the tokens(subjects) for devices.  Could store the settings on the subject, then the admin could revoke the device later, and maybe not need the extra collection at all.

srmoss: Which extra collection are you talking about?  pushclients?  Or the notifications collection?

And let me see if I understand: instead of registering the device for push notifications generically, we would register it specific to a subject.  Then, that subject either has global ack or local ack.  Global would snooze for everyone, local would only prevent further notifications for that specific device and subject's site?

jasoncalabrese:
@srmoss @aditmer:

Can you add some information about your azurepush service, is there a repo?  What are your plans for maintaining 24x7 uptime?

Also I've been thinking about our offline discussion on trying to use nightscout.github.io/pages/test-beta for this and I think it's a bad idea to target the `dev` branch on everyones fork.  The `dev` branches should really stay clean and only have commits from the main `dev` branch.

aditmer:
@jasoncalabrese There is no other code for the azure push service.  It is all in this PR.  The centralized Azure Notification Hub has no code in it, simply the Apple Push certificate and Google Cloud Messaging API key.

RE: the dev branch.  I did not think we were merging to the dev branch now, rather this WIP branch that is solely for this purpose.  The only reason someone would fork this branch would be to set up their Nightscout site to send push notifications as part of the beta test (per our instructions on my GitHub page for the mobile app).

aditmer:
This is how the push notification system works:

### Registration for notifications

0. The user specifies in the mobile app that they want push notifications.
1. The mobile app requests permission from the user to send push notifications.
2. The mobile app requests a device token from Apple Push Notification Service (APNS) or Google Cloud Messaging (GCM).
3. The app sends that token to a web service on the Nightscout website (new server code in this PR).
4. Nightscout registers with our Azure Notification Hub using the device token and settings from the mobile app (which types of notifications are wanted).

### Delivering notifications

0. An event happens on Nightscout (Alert, Info, Announcement)
1. Nightscout checks to see if devices are registered for this type of event.
2. If they are, Nightscout sends the notification to Azure Notification Hub (ANH).
3. ANH forwards the message to APNS or GCM (no code we write is involved; we simply provide the APNS and GCM credentials to ANH to use).
4. The mobile app gets a notification.
5. The user can open the app from the notification or take action on it (snooze) from the Notification Center or lock screen.

Does this make more sense?  The only code we have written lives either in this PR or in the mobile client.  The centralized part, ANH, only has configuration in it, but it handles a lot of the leg work for us.

lib/plugins/azurepush.js
```diff
@@ -0,0 +1,323 @@
...
+    return null;
+  }
+}
+
+function setupAzurePush (env) {
+  if (env.settings.isEnabled('azurepush')) {
+    if (env.settings.baseURL) {
+      //TODO: Allow override
+      //env.extendedSettings.azurepush.hubName, env.extendedSettings.azurepush.connectionString
+      var notificationHubService = azure.createNotificationHubService('Push-Notification-Hub','Endpoint=sb://nspush.servicebus.windows.net/;SharedAccessKeyName=DefaultFullSharedAccessSignature;SharedAccessKey=VdtFDvldTypVUgroq6JtrbI162upFh7MeWcpSdJu3yE=');
```
srmoss: If others have ideas, I'm open to them.  I struggled thinking of a way to avoid having the credentials available publicly (either on a setup guide or in the code) and placing an undue burden on getting the app connected up for push notifications.

I like the idea of a separate "credentialing" service, I'm just not sure how we would implement it without having to maintain **another** centralized service (that is prone to up-time guarantees and maintenance).

Let me know if you have other ideas!

srmoss: P.S.S. We're taking this conversation offline to come up with an approach.  I'll update the code once we come up with an acceptable, secure solution.

srmoss:
OK, RE: uptime: we've created a twitter for @NightscoutPush that will allow us to tweet out when outages are reported.  This way, we can have a separate service that users can follow to be notified of potential service interruptions on the centralized push notification service.  User can follow that handle to be notified of any potential downtime.  We've monitored the service over the past 2.5 weeks and it appears we will not have any false positives/lots of traffic.

The twee that will appear will include no information about the actual report, just that a potential outage was reported (so as to protect the identity of the individual site).  The text of the tweet will be:

At {{AddedAt}}, there was a potential outage reported on the central Nightscout notification service.  Be sure to watch for alerts proactively. #nightscoutpush

srmoss:
OK, the final fix we discussed is live.  We've removed the Azure Notification Hub access key from the Git repository and from Nightscout code.  Now, the shared access key is provided by the app in a special closed source file only used to build and deploy the app.  If someone wants to deploy their own notification hub, they would just need to build and deploy with their own config file.

I think we are really close to being ready to merge this pull request.  We've had two production sites running the new production code for close to a month and had no downtime or unexpected issues!

Let me know what else you see!

hackingtype1:
Looks good, can someone w/ write access please merge? thx

sulkaharo:
The Trello code is still in, which I thought was supposed to be removed before the merge?

srmoss:
We are using Trello for error logging at least through the Beta so we can monitor any potential downtime.  We'll reconsider the logging methodology before meeting back to dev.

jasoncalabrese:
@hackingtype1 didn't know you used Nightscout

hackingtype1:
Actually, I've maintained an instance of the community versions for a long time. And are you being passive aggressive or just aggressive - or am I misreading the tone?

PR #2376, opened 2016-12-27
Added note about using the same Amazon account as you use on your Ale…

jwedding:
…xa app

Initially created my skill using a developer account separate from the app I use to manage Alexa. It wouldn't show up obviously.


jwedding:
Minor edit to the steps. Thanks.

jasoncalabrese:
Looks like this was merged directly to master, but version wasn't changed, etc.

PR #2403, opened 2017-01-10
rus lang update

apanasef:


apanasef:
this is an updated translation for the russian site. seems like all is completed for now

PR #2424, opened 2017-01-20
Upbat temperature

cjo20:
Add a "temperature" field to the upbat plugin, to allow tracking up uploader device temperature.


cjo20:
Tested this by modifying my call to `voltage` in `EdisonVoltage`:

`voltage json batteryVoltage battery | json -e "this.temperature='`sensors | grep Core | awk '{print $3}' | head -1`'" > /home/edison/myAps/uploader.json`

This adds the temperatre of the first core to uploader.json. `sensors` needs `sudo apt-get install lm-sensors`

PR #2447, opened 2017-02-07
Show correct mmol/l value in treatment tooltip

cjo20:
It seems that by the time that treatment glucose values reach `lib/client/renderer.js`, they have already been converted to mmol/L where appropriate. The code was doing a double conversion from mg/dL to mmol/L, leading to a BG of 5.4mmol/L displaying as 0.3 in the treatment tooltip.


cjo20:
Mentioning @sulkaharo as requested.

cjo20:
It seems to work with manually entered BG Check too.

image

MilosKozak:
same issue with CALs .... could you add it too?

MilosKozak:
same bug in reports/treatments. can you add it too?

PR #2455, opened 2017-02-14
Clock Mode improvements

unsoluble:
Added BG-threshold-based background colors, trend arrows, and fixed layout for (some) mobile browsers. Could use some more dynamic elements, like pulling the user's thresholds & page name from NS, but not sure how to call those. Also would love to add the alarms to this view maybe?


unsoluble:
Oh hey, just realized that the replacement apple-touch-icons I made (just for aesthetic preference) are going to be a part of this. Is there a way to mark certain changes as "ignore" when doing a PR?

MilosKozak:
not so easy. you can cherrypick and create new PR

unsoluble:
Is it easy enough for you guys on that end to ignore those at this point? Or are you suggesting it'd be better if I did a new clean PR?

unsoluble:
(Removed all the custom apple-touch-icons. These commits should only be affecting clock.html, plus the addition of the new trend arrow images.)

unsoluble:
Realized that I couldn't easily maintain custom personal changes while having an active PR on the same branch. Closing this one, will re-open on a clean branch.

PR #2464, opened 2017-02-19
Clock colors and trends

unsoluble:
Adds /clock-color.html, an alternative to /clock.html. This one changes background color to reflect the current BG range, and adds a trend arrow. Page scales nicely to fit most browser views.

Now properly grabs BG thresholds from the user profile.

uxqenu8g


MilosKozak:
/api/v1/status.json

and please keep the backgroud black

unsoluble:
Ahh, thanks for that — I'm new to APIs, wasn't sure where to look. Will play around with that soon.

Note that I've moved this code to a new page, at clock-color.html, so that the original black-background view is untouched.

tomas69:
background black colored numbers

unsoluble:
That'd be another way of doing it, sure; I prefer using the bg as color indicator, for better visibility from across the room.

jasoncalabrese:
Why was `static/images/favicon.ico` removed seems to still be used in some places

unsoluble:
Ohhh you know why, I had replaced it with my own local version (just to maintain a separate aesthetic for our family site), and then when cleaning up for the PR I think it got cleared out accidentally with all my other little non-public tweaks. I'll put the original back and re-commit. :)

jasoncalabrese:
Thanks, you should be able to just add the back to your branch

jasoncalabrese:
I like the svg arrows, the unicode arrows we use for the main view have problems on some devices, it would be great to replace them with the sgvs.

PR #2468, opened 2017-02-21
Attempt to update nightscout 21/2/17

SyGrOvE:


SyGrOvE:
I have no idea what I'm doing really - have been having lots of problems with nightscout not receiving data for a few hours at a time in the last week or 10 days, sometimes data comes back after a restart but often not, and I've had gaps of 3 hours or so where the data never comes back. Thought I should try updating next & am following instructions step by step, but the steps I'm seeing are not what the instructions say should be there. Like now, I should see a "merge pull request" button at the bottom of the page, but it's not here.

PR #2495, opened 2017-03-07
IOB & COB rendering speedup

MilosKozak:


lib/plugins/cob.js
```diff
@@ -67,30 +67,51 @@ function init(ctx) {
...
       .sortBy('mills')
       .last()
       .value();
   };
 
+  cob.COBDeviceStatusesInTimeRange = function COBDeviceStatusesInTimeRange (devicestatus, from, to) {
+
+    return _.chain(devicestatus)
+      .filter(function (cobStatus) {
+        return cobStatus.mills > from && cobStatus.mills < to;
```
sulkaharo: Should be >=? ;)

PR #2609, opened 2017-05-23
bump benv to version 3.3.0 to fix issue with updated jsdom

jasoncalabrese:
Maybe this will solve the build failures

@MilosKozak


jasoncalabrese:
@MilosKozak this fixes the build only only changes tests, so I'm going to merge it

PR #2636, opened 2017-06-15
Update

dmanders:


scottleibrand:
This is a no-op: perhaps you meant to merge from nightscout:master into dmanders:master ?

dmanders:
Yes, sorry I was trying to upgrade versions and must have hit the wrong button.
David

PR #2649, opened 2017-06-24
Fix Duration input type

unsoluble:
In the Treatment drawer, most number fields are properly set to type=“number”, except the Duration field (for Combo boluses). Fixed that.


jasoncalabrese:
I think we still want to be able to support mixed input.  We're using a module that can handle input such as "5h", and will be converted to 300 minutes.  I think there was another suggestion that would hint at using numbers and still allow a user to select a alpha keyboard.

unsoluble:
This change doesn’t preclude or restrict mixed input — it merely suggests to the browser which keyboard to present by default. Can still switch to (and enter) alpha.

jasoncalabrese:
I think it's going to vary a lot from browser to browser. With the other number inputs I'm not able to select a different keyboard.

unsoluble:
Oh, okay, that’s interesting. Which browser/OS is that? (I’ve tested on Safari & Chrome on both iOS and Mac.)

jasoncalabrese:
Maybe I'm doing it wrong, but I was trying it with safari on ios

unsoluble:
I just tried it on iOS 6, 7, 10, and 11b, and all presented the full keyboard (with numbers/symbols up first on the type=“number” fields), and could switch to alpha. Not sure what’s happening on your end?

unsoluble:
Right, so: Had a [bunch of people test this](https://www.facebook.com/groups/CGMITCOFFTOPIC/permalink/385349218528769/) on different phones, and it looks like it's Android 7 that does indeed pop the limited numeric keypad that doesn't let you swap back to alpha. So I concede, this tweak would be a regression in that case.

If nobody knows of a way to get the number keyboard to default without locking in it on Android, should probably close this one.

PR #2652, opened 2017-06-27
Update example start command to allow spaces in vars

ps2:
When running the server with environment variables like the following set in your .env file:

```
ENABLE=cob iob careportal
```

Then NS will start with ENABLE actually set to 'cob'. 

This update allows environment variables with spaces in them to be used.


jasoncalabrese:
I never used the my.env example, but I have seen it cause problems for others

PR #2657, opened 2017-06-29
Add food type string to carb entry on graph.

ps2:
Also show Absorption Time in tooltip

This works in conjunction with Loop dev, but is also backwards compatible for systems that do not report these fields.  

There are a lot of trailing whitespace removals in this diff. If you want to view the patch w/o those, append &w=1 to the url. 

If you want me to remove those changes, I can.


francescaneri:
Hi @jasoncalabrese, can you add it with a branch to test it? thanks Matteo

loudestnoise:
👍 yay emoji

PR #2678, opened 2017-07-13
update to node 6.11.1

jtsackdex:


sulkaharo:
You're probably running a very old branch of dev, if you had a local file that was still on the 0.10 series? Dev is now updated to 6.11.1

PR #2679, opened 2017-07-13
Update to node 6.11.1

jtsackdex:


sulkaharo:
Master sadly cannot be updated to 6.11 directly due to incompatibility of multiple node packages, which need to be updated.

PR #2684, opened 2017-07-16
Dev

moimemecloud:


moimemecloud:
I have a little issue with my NS version, currently it displays 9.0 beta. When I upgrade via GitHub .... nothing happens ! Could provide me with a solution please ? Thank you 
Best,
Karim

PR #2689, opened 2017-07-17
Fix clock trend arrow alignment in iOS7/8

unsoluble:
In the /clock-color.html page, the trend arrow wasn't properly centering in iOS 7 and iOS 8. Fixed.


unsoluble:
(Not entirely sure why my dev sync merge is showing up as a commit here; let me know if that's a problem.)

sulkaharo:
LGTM

PR #2707, opened 2017-07-25
Alexa translations

tanja3981:
The alexa response was hardcoded in english texts so far. I changed this so translations can be added to support alexa speaking different languages.


tanja3981:
Could anybody approve this pull request or let me know if there is anything not right?

lib/plugins/basalprofile.js
```diff
@@ -101,19 +101,38 @@ function init (ctx) {
...
-      var minutesLeft = moment(basalValue.treatment.endmills).from(moment(sbx.time));
-      response = preamble + 'temp basal of ' + basalValue.totalbasal + ' units per hour will end ' + minutesLeft;
-    } else {
-      preamble = (slots && slots.pwd && slots.pwd.value) ? slots.pwd.value + ' ' : 'Your ';
-      response = preamble + 'current basal is ' + basalValue.totalbasal + ' units per hour';
+        var basalValue = sbx.data.profile.getTempBasal(sbx.time);
+        var response = 'Unable to determine current basal';
+        var preamble = '';
+        if (basalValue.treatment) {
+            preamble = (slots && slots.pwd && slots.pwd.value) ? translate('alexaPreamble3person', {
```
sulkaharo: Will these work even in the non-Alexa use-case?

sulkaharo:
Hi! Sorry for the extremely long delay. :( I'll pull this soon to test. Can you check this all works fine with the latest dev, I recently PRd in fairly significant changes.

lib/plugins/basalprofile.js
tanja3981: Do you know where else this is used? I checked the unit tests, they are still intact. I didn't check anything else as I don't know where this might be used.

tanja3981:
After merging current DEV the unit tests don't work anymore. :-(

sulkaharo:
Ah yes, sorry. :( I know there's one more I broke with the latest PR, a few are broken for unknown reasons. If you have time to help debug, that'd be super awesome.

tanja3981:
Hi

I tried to check out what the problem might be. However, running the tests
locally there are different tests failing than on travis, and those not
working on travis do work locally.
Don't know how to help.
Do you have any suggestions? I'm using node 7.9 locally.

Best regards,
Tanja

sulkaharo:
Do you have Mongo running? If the tests can connect to a local Mongo, you should be able to reproduce the test results as is.

sulkaharo:
No wait, looks like some tests time out in Travis, which is strange. Taking a look.

tanja3981:
Sorry to come back so late. Couldn't make it this week... 
I just merged the current DEV into this branch and tests are fine now.

PR #2727, opened 2017-08-05
add missing indexes for nsclient

MilosKozak:


sulkaharo:
Note this is likely to increase the amount of problems people have with mlab scaling - we already have too make indexed fields and most of the queries fail to use the indexes so adding more will probably just make things worse overall... :(

stavlor:
https://drive.google.com/file/d/0B8GpOcukfs3-VFVJRXlRemZKWlU/view?usp=drivesdk note that not having it is already causing mLab e-mails. I added it the other day on mine.

MilosKozak:
@sulkaharo what's the issue with indexes?

sulkaharo:
In databases, the software needs to update the indexes on every insertion of new data (heavy operation) and have relevant parts of the index in memory when queries are issued, so indexes cause the server to work a lot harder, even in cases where the indexes won’t actually speed up the query - just adding an index doesn’t mean the query is automatically faster. All of this increases the probability that the free mlab tier will run out of resources, so we’re worried about performance, we should actually run all our queries through a query analyzer and re-structure the queries to use the indexes. I looked at the original indexing Ben added to the software and I’m pretty sure the database ignores the index on queries, because the query structure of the “query for anything” API causes a full table scan (going through every element in the collection) anyway.

MilosKozak:
i've added indexes for fields i'm using for dedupping in websockets communication.
https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/websocket.js#L286
do you think it will not help @sulkaharo ?

sulkaharo:
I’m pretty sure that query either goes through the entire collection or only uses the created_at index.

MilosKozak:
did you see the screenshot from stavlor?

sulkaharo:
No, where is it? You can check if indexes are helping with these instructions: https://docs.mongodb.com/v3.4/tutorial/analyze-query-plan/

MilosKozak:
https://drive.google.com/file/d/0B8GpOcukfs3-VFVJRXlRemZKWlU/view?usp=drivesdk

PR #2740, opened 2017-08-13
update time!

tangermean:


stavlor:
Note this pull request is in the wrong direction, you want to do it the otherway see this http://nightscout.github.io/pages/update-fork/

PR #2763, opened 2017-08-25
Upgrade node from 8.1.4 to 8.4.0 and store treatments in UTC (and return deduped/sorted)

PieterGit:
- Upgrade node from 8.1.4 to 8.4.0 (speed improvement)
- Nightscout fix for https://github.com/openaps/oref0/issues/600
  - Store uploaded/inserted/upserted/changed treatments with `created_at` 
  - Improve unit test for treatments with different Timezones
  - Sort treatments on `created_at` field before sending to the client

Question:
- @sulkaharo :  why did you use t.timestamp in https://github.com/nightscout/cgm-remote-monitor/blame/dev/lib/api/treatments/index.js#L42 . I would have expected t.created_at


PieterGit:
I deployed this PR and think it's ready to be tested and reviewed. If somebody can also double check whether when converting to this release does not cause duplicate treatments when upgrading to this release that would be great. I can't test that anymore because I converted earlier on a different code base.

So the test I would like to be either  (simple):
- upgrade your nightscout to this PR
- check the treatments in Nightscout and make sure you don't see any duplicates

Or more complex:
- check/save your https://yournightscout/api/v1/treatments.json?find[created_at][$gte]=2017-08-26
- upgrade your nightscout
- make sure your oref0 0.5.3 rig uploads the pump treatments
- check/save your https://yournightscout/api/v1/treatments.json?find[created_at][$gte]=2017-08-26
- make sure you don't see any duplicates

sulkaharo:
@jamorham Does xDrip normalize the data to UTC before uploading? Any other uploaders we need to double check before merging to dev?

sulkaharo:
@jasoncalabrese you might want to weigh in here too - I'd love to add data validation, only worry is will something break when this goes out.

@pazaan does the 640G uploader do any two-way synching / how does the date normalization work?

jasoncalabrese:
I think it's too late to change any data rules for uploaders, there are too many out it the wild and many will never be updated, having an extra field would be safe, but extra index size

stavlor:
Right far easier to change what we do with the data than change all the
uploaders

PieterGit:
@sulkaharo @jasoncalabrese : this PR does not enforce data validation. It converts `created_at` timestamps with timezone information to a common / canonical UTC format when storing it.  It implements the https://en.wikipedia.org/wiki/Robustness_principle "Be conservative in what you send, be liberal in what you accept". The most efficient way of implementing that is storing the `created_at` values always in UTC format.

All uploaders should accept UTC format now, because otherwise treatments created in Nightscout won't work for them (Nightscout always uses UTC format). The problem is that not all uploaders upload in UTC format (e.g. oref0, ) which causes bugs, that also trigger "bugs" in Nightscout, e.g.:
- future treatments can fail, because it select future treatments based on UTC format
- sometimes treatments can be missed when queried with: https://nightscout/api/v1/treatments.json?find[created_at][$gte]=2017-08-25

What's the suggested way forward?

pazaan:
@sulkaharo 600-series uploader doesn't do 2 way syncing at the moment. We're only just starting to upload treatments in an upcoming release, so I can make sure they adhere to whatever standard you like. I assume you want the `Z`-notation?

PieterGit:
Yes, in UTC format in the `created_at`  field, thus ending with a `Z` and not having timezone information.
Note that I don't change anything to the `timestamp` field. oref0 still uses the localtime format (with timezone), e.g. `2017-08-26T08:53:42+02:00`

sulkaharo:
@PieterGit Oops Node 8.1.4 is the latest version available in Azure. I don't have an ETA for when they update, so better no update to a later version.

PieterGit:
@sulkaharo no problems to reset to node to 8.1.4 for this PR.  Wouldn't it be better that Travis (and Nightscout) support both 8.1.4 (latest Node version on Azure) and 8.4.0 (latest Node version available Heroku).  I thought Travis can also be configured to test with multiple Node versions.

@jasoncalabrese do you still object to this PR?

sulkaharo:
Ah yes, sorry for poor choice of words earlier, didn't mean this adds validation. @jasoncalabrese the idea is that NS API would fix data as it comes in, if we determine it's fixable safely. I would add features to disregard completely invalid data though, such as CGM entries that don't contain a SGV value -> NS should guarantee some level of data validity.

@PieterGit deployment to Azure fails if package.json declares anything newer than 8.1.4. We can add 8.4.0 to Travis, but not sure how you'd use that in Heroku given it uses the version from package.json. (I'm personally using Heroku, so yes would prefer 8.4 but that's at least now a no-go to most NS users.)

PieterGit:
I reverted to Node 8.1.4 at @sulkaharo request. Can this be merged to dev?

PieterGit:
@jasoncalabrese @sulkaharo : what's the way forward with this issue. Now that 0.10 has been released, I would like to know how this needs to be merged in a future version.

I see two major options:
1  Change the internal behaviour and not increase the api version. In this case all combinations of uploaders and APS'es must be checked to see if they work for all timezones.
2. create a new v2 api interface that requires uploaders to upload in UTC format (and probably have more validations). 

I think 1 is faster, but can pose a risk. On the other hand I also found bugs in the current implementation that can lead to problems if there is a interpretation difference between uploader and APS. I noticed false postitves for future CGM records (oref0 fails working if it thinks cgm values are more than 10 minutes old) in +02:00 timezone. I think that could also work the other way around: stale date which can be interpreted as current with negative TZ offset.

What is your opinion?

PieterGit:
Postponing to 0.11.0 or later. Let's first get this right for the activity monitoring in Nightscout.

PieterGit:
Current state is still: postponed until 0.11. Probably will start to create a new API v2 version which enforces uploaders to do the the right things. That change is far too big for 0.10.x. Implementation strategy:
- make enforcing endpoints for cgm for Nightscout (and probably also for pump history)
- create a branch for uploaders such as oref0, AndroidAPS, etc. that uses these new API's
- make sure it works
- release the nightscout version with v1 and v2 API
- release the APS versions that can work with these new v2 endpoints

PR #2768, opened 2017-08-26
Graceful app launch when offline

sulkaharo:
* Trigger app on DOM ready event (as recommended by HTML5 spec)
* Detect if app is completely offline, auto-reload if needed.


PieterGit:
@sulkaharo 

Tested on Samsung S7 with 
image

Results on mobile mode: (size to 20%)
image

Results on "Desktopversion" checked on Samsung S7 chrome
image

The interface did not show an Offline warning, but note there was a "Last data received 2 hour ago" notification active at that time.

static/admin/index.html
```diff
@@ -45,8 +45,8 @@ <h1 class="translate">Admin Tools</h1>
     <b><span class="translate">Authentication status</span>: </b>
     <span id="authentication_placeholder"></span>
 
-    <script src="/js/bundle.js?v=0.10.0-dev-20170821"></script>
+    <script src="/js/bundle.js?v=8.1.4-dev-2017082601"></script>
```
jasoncalabrese: all the `0.10.0`'s got replaced with `8.1.4`, replacing node versions somewhere else?

PR #2773, opened 2017-08-28
Upgrade to nodejs 8.4.0 and fix unit tests for travis

PieterGit:
This should make deployment on Heroku (node 8.4.0) and Azure (node 8.1.4) possible.
Travis is instructed to test both node versions.

Update: Azure now supports node 8.4.0, so not allowing 8.1.4 anymore because it breaks on travis CI builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
 sudo: required
 node_js:
-  - 6
+  - "8.1.4"
+  - "8.4.0"
 matrix:
   fast_finish: true
 services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 sudo: required
-env:
-  - MONGO_CONNECTION="mongodb://localhost/test_db" storageURI="mongodb://localhost/test_db"
 node_js:
   - "8.4.0"
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: required
 env:
   - MONGO_CONNECTION="mongodb://localhost/test_db" storageURI="mongodb://localhost/test_db"
 node_js:
-  - "8.1.4"
   - "8.4.0"
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: node_js
 sudo: required
 env:
-  - MONGO_CONNECTION=mongodb://localhost/test_db
-  - storageURI=mongodb://localhost/test_db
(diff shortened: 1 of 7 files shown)


PieterGit:
Deploys and works on Heroku. Can somebody test this on Azure?
@sulkaharo  please review

sulkaharo:
@PieterGit wanna merge DEV, I fixed the unit tests :)

PieterGit:
@sulkaharo : I updated to dev, but still some tests fail. As far as I can see I had nothing to do with that...
I see @jasoncalabrese also is working on fixing tests.

PieterGit:
@jasoncalabrese @sulkaharo :  I think the main root cause was the generation of the bundle.js.  I added that to the `Makefile` and that caused the build to be finished in about a minute. Can you review my other changes?

PieterGit:
False alarm. I made an error in the `Makefile` so that the tests did not run.
The only test which is now failing for me is:
```
not ok 204 Profile editor "before each" hook for "should produce some html"
  Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
      at tests/fixtures/headless.js:160:7
      at Object.module.exports.setup (node_modules/benv/index.js:24:56)
      at Function.init [as setup] (tests/fixtures/headless.js:16:10)
      at Context.<anonymous> (tests/profileeditor.test.js:94:14)
```

PieterGit:
@sulkaharo @jasoncalabrese : Now is ok for Node 8.4.0, but it fails with Node 8.1.4.  What's the best way to go forward? Profile the two tests that used to fail, or do you have other suggestions.

Minor thing: I noticed a 404 with the /profile page to https://datejs.googlecode.com/svn/trunk/build/date.js?_=1505082730792 which was from an old jquery js. I couldn't find the code that was causing it.

sulkaharo:
Azure started supporting 8.4.0, so we could bump to that now that it works on the minimum set of platforms (locally, Heroku, Azure), if the performance issue is caused by 8.1.4. I've see the tests randomly work/fail though, so I wouldn't immediately conclude the build works on either without seeing multiple builds succeed with consistent timing (and I still thing this is an issue with something Travis does, since I've never seen the timeouts locally).

PieterGit:
Things are looking quite stable now. I minimized the PR to the bare minimum. Can you please review and merge to dev if it's ok.

sulkaharo:
LGTM. Pulled; tests all succeed

sulkaharo:
Wonder if the Travis success is actually due to this: https://medium.com/the-node-js-collection/get-ready-a-new-v8-is-coming-node-js-performance-is-changing-46a63d6da4de

PR #2780, opened 2017-09-03
wip/onlinestatus + trying to fix test failures

jasoncalabrese:


sulkaharo:
Given all the test in the branch work locally and randomly work in Travis, I'd like to merge this to get more testers on board. I've run this in production for over a week now without issues, tested on multiple Android/Win/iOS/OS X devices and haven't found issues + know other people have tested successfully as well. Once this is in, I think we can go into release packaging mode.

francescaneri:
fullsizeoutput_7d
formatting problem on iphone 6 more but there was also before. While charging you also see the new icon ;)

sulkaharo:
@francescaneri Do you mean the pump pill pushing the content down? That's due to the pump pill being so wide it doesn't fit the line with the COB and other pills. There's no easy fix, other than restricting the max width of each pill, which would then cause content to be hidden. :(

PR #2813, opened 2017-09-22
Habanero

mdomox:


sulkaharo:
You've accidentally made the PR the wrong way (from your code to nightscout, not the other way round)

PR #2815, opened 2017-09-23
Habanero 2

ifferj:


sulkaharo:
You've accidentally made the PR the wrong way (from your code to nightscout, not the other way round)

PR #2827, opened 2017-09-23
update to node 6.11.1

ak8418:


sulkaharo:

ak8418:
Thank you for fixing the problem and for all your help. 
Angela

PR #2868, opened 2017-09-28
Should fix the V2 API & remain compatible with Android nsclient

sulkaharo:


sulkaharo:
Hmm, maybe should not have beautified the js before committing...

MilosKozak:
yes :(

MilosKozak:
No, it doesn't work. intervals are processed before sending to websockets

PR #2894, opened 2017-10-06
SVG dimension fix

unsoluble:
Somehow the NONE.svg ended up with an incorrect rectangle shape; fixed
to match the aspect ratio of the other symbols.


sulkaharo:
How did this manifest itself on the site? As in - what does this fix? :)

unsoluble:
Hah, sorry, probably should have been more specific in the commit. This is one of the trend-arrow images used by the /clock-color view (and potentially other views in the future). It's an SVG rectangle, and its dimensions were causing it to fill the enclosing block, instead of staying in proportion to the other images. :)

PR #2904, opened 2017-10-09
Remove pattern regex from basal change field

unsoluble:
Having the regex require integers means that only the numeric keypad will pop on iOS. This prevents the entry of negative values, which is pretty critical for temp basals.

I'm not sure whether there's adequate sanitizing of the entry on the other side of this; that should be double-checked.

### DIFF
--- a/views/index.html
+++ b/views/index.html
@@ -339,7 +339,7 @@
                     </label>
                     <label id="percentLabel" for="percent" class="left-column short-label">
                         <span class="translate">Percent</span>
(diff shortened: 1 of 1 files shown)


sulkaharo:
LGTM; did you check the underlying math? I thought we used Medtronic-style temp basal values, but looks like not, so wonder how the regex ever made it to the UI.

unsoluble:
I'm not actually sure _where_ the underlying math is; I'd have to dig around a bit further to check.

If I had to guess, the regex was initially intended to pop a more appropriate mobile keyboard on some devices — it may indeed have been popping a better one in the past. But with it now bringing up merely the numpad, better to loosen the restriction a bit I think.

PR #2911, opened 2017-10-11
New attempt at v2 API fix

sulkaharo:


sulkaharo:
@MilosKozak This should be identical to your PR except the use for cloneDeep(); can you formally approve :)

PR #2919, opened 2017-10-15
display ZT PredBGs, and forecast tooltips

scottleibrand:
This displays the new zero-temp predicted BG array generated by https://github.com/openaps/oref0/pull/714

It also contains a number of readability improvements, including:
 - You can now tap / mouseover purple predBG tooltips to see what kind of forecast they're from
 - Tooltip targets for carbs and boluses are much narrower, so you can mouseover the thing you're actually interested in rather than something 30m earlier/later.
 - Reduces font sizes for treatment labels so they're more readable with OpenAPS SMB.

One other change we may or may not want:
 - Removes package-lock.json from git control and adds it to .gitignore to prevent merge conflicts.


.gitignore
```diff
@@ -1,6 +1,7 @@
 bower_components/
 node_modules/
 
+package-lock.json
```
sulkaharo: So as per NPM documents, this is intended to be included in the repo

sulkaharo:
Looks good; will test soon

scottleibrand:
Per our discussion, including package-lock.json in the repo prevents you from running `git pull` after `npm install`. Is there another better solution that doesn't require everyone hosting their own NS to know the ins and outs of npm package management? Leaving it out of the repo seems to work fine, as it lets npm choose the proper version of packages (following package.json) automatically. Is there a reason we need to disable such automatic behavior and force manual package management?

lib/client/renderer.js
```diff
@@ -835,7 +839,8 @@ function init (client, d3) {
         .style('fill', 'white');
 
       label.append('text')
-        .style('font-size', 40 / opts.scale)
+        // reduce the treatment label font size to make it readable with SMB
+        .style('font-size', 25 / opts.scale)
```
sulkaharo: Feels very small with non-SMB user now; this would probably feel like a regression to most users. Wonder if we could something like count how many entries there are in the timeline and set the font size dynamically.

scottleibrand: That would be good.  I'm also playing around with getting rid of the 0. and U to make the label text narrower, and possibly removing the label entirely for the smallest boluses (maybe just 0.1U ones).

scottleibrand: #2926 has the leading 0 and trailing U removed, and changes the font size almost back up to where it was.

alimhassam: Im not an open aps user (yet) but another interesting idea would be to display the bolus even at a different height in the chart based ok the IOB at the time.
Does that sound like a good idea? 
We generally don't care that much about the individual SMB but more about the totally IOB anyway I would think.
I'll try to experiment adding a different plugin that does that and maybe also draws the IOB curve as well to see how it looks. Would be a separate thing but hopefully it can help with the same problem.

scottleibrand: Oh, that's an interesting approach.  I have long liked the idea of rendering IOB and COB on the graph, as the NS Reports Day to day report does.  Plotting the insulin treatments along the IOB line, and the carb treatments along the COB line, would be a nice way to get them out of each other's way and make it easier to see everything (at the expense of cluttering things up a bit).

.gitignore
scottleibrand: Have you decided what you'd like to do here?

scottleibrand:
I've removed the visible rendering changes from this PR and moved them into the `rendering` branch: https://github.com/nightscout/cgm-remote-monitor/pull/2926

sulkaharo:
Hmm looks like the test fails on this branch are real; getting 4 fails when running the tests locally.

scottleibrand:
Any idea why?  I don't think I touched any code that should affect those...

sulkaharo:
Check the stack traces in the unit test log? Looks like d3 is now trying to access a non existing property, so maybe related to one of the changed d3 calls?

lib/client/chart.js
```diff
@@ -73,9 +73,15 @@ function init (client, d3, $) {
...
 
   function dynamicDomain() {
-    var mult = 1.3
+    // allow y-axis to extend all the way to the top of the basal area, but leave room to display highest value
+    var mult = 1.15
       , targetTop = client.settings.thresholds.bgTargetTop
-      , mgdlMax = d3.max(client.entries, function (d) { return d.mgdl; });
+      // filter to only use actual SGV's (not rawbg's) to set the view window.
+      // can switch to Logarithmic (non-dynamic) to see anything that doesn't fit in the dynamicDomain
+      , mgdlMax = d3.max(client.entries, function (d) { if ( d.type === 'sgv') { return d.mgdl; } });
```
PieterGit: @scottleibrand : shouldn't this function always return a value, if you're finding the max, you can return `0` for non sgv values. Perhaps this fixes the unit tests, see travis or https://gist.github.com/PieterGit/5aada655fb268559bdcd229f9d621de1

PieterGit:
Please don't use this PR anymore, but help fix the tests on https://github.com/nightscout/cgm-remote-monitor/pull/2991

.gitignore
PieterGit: Reverted on PR https://github.com/nightscout/cgm-remote-monitor/pull/2991 . `package-lock.json` is removed from dev branch and we now use npm-shrinkwrap.json
So no need for .gitignore.

scottleibrand:
Merged #2991 to `rendering`, so closing this in favor of #2926.

PR #2920, opened 2017-10-15
Adding some test coverage for /lib/client/renderer.js

christopher-fredregill:
I noticed the coverage report showed a pretty large gap in coverage for /lib/client/renderer.js. This is just an incremental step towards getting that 70% coverage mark up to a healthier number... only adds a test and a small change to .gitignore for VS Code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules/
 
 bundle/bundle.out.js
 
+.vscode/
 .idea/
 *.iml
 my.env
--- a/tests/client.renderer.test.js
+++ b/tests/client.renderer.test.js
@@ -1,0 +1,29 @@
+'use strict';
+
+require('should');
+let _ = require('lodash');
+
+let renderer = require('../lib/client/renderer');
+const MAX_DELTA = 0.0001;
+const PREV_CHART_WIDTHS = [
+  { width: 400, expectedScale: 3.5 }
+  , { width: 500, expectedScale: 2.625 }
+  , { width: 900, expectedScale: 1.75 }
+];
+
+describe('renderer', () => {
+  describe('bubbleScale', () => {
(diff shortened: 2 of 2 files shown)


sulkaharo:
Yayyy getting more test coverage would be amazing. Another issue we _really_ need help with is finding out what's causing the huge run time discrepancy between running tests on a dev laptop vs Travis. The long runtime in Travis is causing the tests to fail due to timing out, causing false alarms

christopher-fredregill:
Ah yeah, I don't remember if I added an issue yet for that or not. Regardless, I'd be happy to look into that.

christopher-fredregill:
@sulkaharo I actually have had trouble just getting all the tests to pass locally; seems like there are 4 tests which may just require some special setup that I haven't yet found documented anywhere in the README.md. Here are the ones I'm currently struggling with:

```
admintools.test.js
careportal.test.js
profileeditor.test.js
reports.test.js
```

In looking over the build history in Travis, it appears that one of these - `profileeditor.test.js` - also has a higher chance of failing than the others. Some questions:

1) Are there any steps missing from the README.md that you follow in order to get the 4 tests I mentioned above to pass, locally?
2) Have you considered separating out the integration and end-to-end tests from the unit tests? Tests requiring real MongoDB connections or browser environments will always be more brittle than simple unit tests, but one strategy I've seen to help alleviate issues like this is to run unit tests on all pushes, and reserve integration tests only for certain situations (merges to certain branches, releases, etc.). Something like that can limit exposure to brittle builds, and also allow more specialization in the testing tasks, themselves (more retry/wait-for logic wrapping integration tests, for example; they may take longer to run, but if you're only running them on really important commits, it may be a worthwhile tradeoff, especially if the buttressing means that the number of false negatives from CI decrease over time).

PR #3037, opened 2017-11-09
Upgrade to node 8.9.1 and npm update and some dutch translation fixes

PieterGit:
See changelog https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#8.9.1


PieterGit:
Needed to remove some duplicate lines in language.js. Don't understand why Travis complained about it,  but removing duplicate keys is always a good idea.  I also updated some dutch language translations.

PieterGit:
@sulkaharo : I tested it today, and didn't find any strange things. I think it's ready to merge to dev.

elnjensen:
FYI, I was just starting up another branch on Azure (not this one) and looking at the build logs, it says this: 

```
Node.js versions available on the platform are: 0.6.20, 0.8.2, 0.8.19, 0.8.26, 0.8.27, 0.8.28, 0.10.5, 0.10.18, 0.10.21, 0.10.24, 0.10.26, 0.10.28, 0.10.29, 0.10.31, 0.10.32, 0.10.40, 0.12.0, 0.12.2, 0.12.3, 0.12.6, 4.0.0, 4.1.0, 4.1.2, 4.2.1, 4.2.2, 4.2.3, 4.2.4, 4.3.0, 4.3.2, 4.4.0, 4.4.1, 4.4.6, 4.4.7, 4.5.0, 4.6.0, 4.6.1, 4.8.4, 5.0.0, 5.1.1, 5.3.0, 5.4.0, 5.5.0, 5.6.0, 5.7.0, 5.7.1, 5.8.0, 5.9.1, 6.0.0, 6.1.0, 6.2.2, 6.3.0, 6.5.0, 6.6.0, 6.7.0, 6.9.0, 6.9.1, 6.9.2, 6.9.4, 6.9.5, 6.10.0, 6.10.3, 6.11.1, 6.11.2, 6.11.5, 7.0.0, 7.1.0, 7.2.0, 7.3.0, 7.4.0, 7.5.0, 7.6.0, 7.7.0, 7.7.4, 7.10.0, 7.10.1, 8.0.0, 8.1.4, 8.4.0, 8.5.0, 8.7.0, 8.8.0, 8.8.1, 8.9.0.
```

So I don't see 8.9.1 listed there.  I don't know if there's some other way to add it, or if the code would safely fall back to 8.9.0, but from that info anyway, it doesn't look like 8.9.1 is available.  Again, untested with this PR, just posting the log message.

PieterGit:
@elnjensen : thanks. Please let us know when 8.9.1 is available on Azure.
@sulkaharo : what do we do with this PR? There is a (minor) security patch in NodeJS and I would like Nightscout releases to have the latest node 8.9.x LTS when we release them. I don't think there are a lot of Azure users on the dev branch (and they should be able to change the Node version). I would say: merge to dev now and wait with Nightscout release until Microsoft support node 8.9.1 on Azure.
Thoughts?

sulkaharo:
@PieterGit IMO the most critical thing here is we're at all times providing a stable runtime with no known remote exploits. The .1 release fixes things we're not using in the codebase, so theoretically we could release it already. Plus if we set in stone we always have to release with latest, we'll probably very often hit an issue where we're about to release, Node releases a new version, so we have to re-test things again, during which time they might release yet another version. :)

Let's wait for the .1 to go out into Azure, then release with that regardless of where Node is at that point.

PR #3039, opened 2017-11-09
Update bgclock.html

danielharrelson:
Add trending arrow beside glucose value.


danielharrelson:
screen shot 2017-11-09 at 6 49 37 pm

unsoluble:
Wondering if we should think about renaming these views so they make some sort of sense. :)

But of course that'd break everyone's existing setups. :(

PR #3043, opened 2017-11-11
Updating package.json to reflect current dependencies in dev

elnjensen:


PieterGit:
@elnjensen : what code is using these dependencies? Is this required for Azure deploy?
I  also pushed a PR for the latest Node version (8.9.1).   See https://github.com/nightscout/cgm-remote-monitor/pull/3037

Probably best to integrate these PR's into one.
Can you confirm that NodeJS 8.9.1 is available on Azure?

elnjensen:
@PieterGit   This branch is something totally different than a node version update.  This `wip/azurepush` branch contains code that will let a new iOS version of the Nightscout mobile app do push notifications without using Pushover.  I haven't been closely involved - this is work by @hackingtype1 and Andrew Ditmer - but much of the work was done a while ago, so I was just trying to get the `package.json` file in this branch closer to being parallel with the dependencies already present in dev so it would integrate cleanly.  The only dependency changes added here are `azure-sb` and `node-trello`.   And the "azure" in that package name relates to the message service used - runs fine on both Heroku and Azure.    So I think you should proceed with your PR - this is unrelated.   (Also note that this isn't yet a PR against dev or master - just trying to get this branch itself ready.)

PieterGit:
@elnjensen or @hackingtype1 . What's the state/progress of this PR? Do you think it's possible that this PR can be ready in at most two weeks. If not, I'll like to propose to set this feature for 0.11.0 (and not for 0.10.3)

PieterGit:
@elnjensen or @hackingtype1 . Can you give an update if this PR? Is still needed or can I close it. If it's needed can you explain for what features / code within Nightscout. I'm still a bit lost in this PR for `package.json` changes, without updates to Nightscout code to use these new packages.

PieterGit:
@elnjensen or @hackingtype1. Postponing this PR to 0.11 release. Can you please give an update? If no response I think we should close or apply this PR and then close the branch.

PieterGit:
@elnjensen or @hackingtype1 can you give an update on this issue and this branch? Please also fix the failing tests. I would like to clean stale branches in the Nightscout 0.11 release cycle.

PR #3045, opened 2017-11-11
add root mean squared out of range metric

jakepusateri:
Out of range root mean squared (https://en.wikipedia.org/wiki/Root_mean_square) is calculated by summing the square of the amount out of range for all readings, dividing by the count of readings, and taking the square root. 

A period of 100% in-range would have RMS of 0. More time/amount out of range increases the number.


jakepusateri:
Added a note about what RMS is. This metric is useful in that it can distinguish out of range 10 points for 2 hours from out of range 100 points for 2 hours.

lib/report_plugins/glucosedistribution.js
```diff
@@ -385,7 +398,10 @@ glucosedistribution.report = function report_glucosedistribution(datastorage, so
 
 	var TDC = deltaTotal / days;
 	var TDCHourly = TDC / 24.0;
-	
+
+    var RMS = Math.sqrt(RMSTotal / events);
+    console.log('RMS', RMS);
```
PieterGit: Please remove (debug) logging.

PR #3058, opened 2017-11-20
Add data-hours="4" option

scottleibrand:
With OpenAPS predictions being 4 hours in length, I've found it useful to set my NS view to +/- 4H.  This PR adds a 4H option.  To make room for the extra one, it shortens "HR" to "H" and removes "H" from the 12 and 24 hour options, which results in "2H 3H 4H 6H 12 24 ..." (which fits on my screens without wrapping).

Also includes #3048


alimhassam:
would be nice to make this customizable somehow. I find that 8h gives be a better overall view of what happened in the last meal, so i have personally replaced my 6h with 8h on my site for example.

scottleibrand:
Please do, and open a new PR when you figure it out.  I tried and failed with https://github.com/nightscout/cgm-remote-monitor/compare/4h-option...focus-range-hours

PieterGit:
Problem with this PR is that it doesn't work in other languages. Some languages have up to 6 characters to define the time period. So shortening it will give problems. For example from `languages.js`:
```
   ,'3HR' : {
      cs: '3hod'
      ,de: '3h'
      ,es: '3h'
      ,fr: '3hr'
      ,el: '3 ώρες'
      ,pt: '3h'
      ,sv: '3tim'
      ,ro: '3h'
      ,bg: '3часа'
      ,hr: '3h'
      ,it: '3ORE'
      ,dk: '3t'
      ,fi: '3h'
      ,nb: '3t'
      ,pl: '3h'
      ,ru: '3ч'
      ,sk: '3 hod'
      ,nl: '3 uur'
      ,ko: '3시간'
      ,zh_cn: '3小时'
      ,zh_tw: '3小時'
      }
```

I think we should try to get a customizable date-hours function in 0.11 and postpone this PR for now.

@scottleibrand Agree?
@alimhassam would you like to create a customizable date-hours PR?
@sulkaharo @MilosKozak @jasoncalabrese Thoughts?

PieterGit:
@Tornado-Tim Last time I checked this PR needed some polishing before it can be merged.

@scottleibrand Nightscout is not only used in the US and should be able to be used internationally. For example in Dutch the commonly used abbreviation for hour is `u` which make the UI confusing given the fact that the dosing unit `u` is also used internationally. That can cause confusion. Similar issues might occur in other languages that don't have a 2 character abbreviation for hour, like `HR` in English. 

Second problem was that last time I checked this PR the mobile layout was made horrible because of this PR (tested with Dutch).  Unfortunately I didn't post a screenshot of it. 

I don't know if that's fixed now. This might be fixed in various ways:
- make the list of hours that are show configurable . @scottleibrand didn't want to PR that
- remove the spaces in `languages.js` and make sure the wrapping goes fine on mobile screens

sulkaharo:
Implemented in #5289

Issue #3059, opened 2017-11-20
Update to latest dev 0.10.2 on heroku failed


osodebailar:
Update on latest dev failed with following log entry:

-----> Node.js app detected
-----> Creating runtime environment
       
       NPM_CONFIG_LOGLEVEL=error
       NPM_CONFIG_PRODUCTION=true
       NODE_VERBOSE=false
       NODE_ENV=production
       NODE_MODULES_CACHE=true
-----> Installing binaries
       engines.node (package.json):  8.9.1
       engines.npm (package.json):   unspecified (use default)
       
       Resolving node version 8.9.1...
       Downloading and installing node 8.9.1...
       Using default npm version: 5.5.1
-----> Restoring cache
       Skipping cache restore (new-signature)
-----> Building dependencies
       Installing node modules (package.json + shrinkwrap)
       
       > uglifyjs-webpack-plugin@0.4.6 postinstall /tmp/build_9782e35936c1c3753b9cb2d5c2415d55/osodebailar-cgm-remote-monitor-1218b81/node_modules/uglifyjs-webpack-plugin
       > node lib/post_install.js
       
       
       > Nightscout@0.10.2-release-20171109 postinstall /tmp/build_9782e35936c1c3753b9cb2d5c2415d55/osodebailar-cgm-remote-monitor-1218b81
       > webpack --config webpack.config.js && npm run-script update-buster
       
       Production environment detected, enabling UglifyJsPlugin
       Hash: a7b09b7d474e72c01133
       Version: webpack 3.8.1
       Time: 31690ms
       Asset     Size  Chunks                    Chunk Names
       js/bundle.js  1.91 MB       0  [emitted]  [big]  app
       js/bundle.js.map  10.8 MB       0  [emitted]         app
       [23] (webpack)/buildin/module.js 495 bytes {0} [built]
       [25] ./lib/language.js 342 kB {0} [built]
       [40] ./lib/units.js 291 bytes {0} [built]
       [254] ./bundle/bundle.source.js 863 bytes {0} [built]
       [256] ./static/css/drawer.css 1.01 kB {0} [built]
       [259] ./static/css/dropdown.css 1.02 kB {0} [built]
       [261] ./static/css/sgv.css 1.01 kB {0} [built]
       [274] ./lib/client/index.js 36.6 kB {0} [built]
       [440] ./lib/client/boluscalc.js 26.7 kB {0} [built]
[...]

PR #3061, opened 2017-11-20
temp workaround for https://github.com/nightscout/cgm-remote-monitor/…

increddibelly:
temp workaround for #3059 until pull request https://github.com/julien-maurel/js-storage/pull/8 is merged into js storage


increddibelly:
it's somehow not working with the older version but I'm not going to investigate; js-storage is fixed so this is no longer needed.

PR #3086, opened 2017-11-29
Smarter formatting based on amount of data

sulkaharo:
To make sure the rendering works for both SMB and non-SMB users.

@scottleibrand can you test if this still works as you intended for SMB users?

### DIFF
--- a/lib/client/renderer.js
+++ b/lib/client/renderer.js
@@ -451,9 +451,10 @@ function init (client, d3) {
       
       if ( treatment.insulin < 1 && !treatment.carbs && enteredBy.indexOf('openaps') > -1) { // don't show the unit of measurement for insulin boluses < 1 without carbs (e.g. oref0 SMB's). Otherwise lot's of small insulin only dosages are often unreadable
          unit_of_measurement = '';
+         // remove leading zeros to avoid overlap with adjacent boluses
          dosage_units = (dosage_units+"").replace(/^0/,"");
       }
-      // remove leading zeros to avoid overlap with adjacent boluses
+
       arc_data[3].element = dosage_units + unit_of_measurement;
     }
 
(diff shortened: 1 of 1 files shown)


lib/client/renderer.js
```diff
@@ -444,13 +444,17 @@ function init (client, d3) {
...
+      
       var unit_of_measurement = ' U'; // One international unit of insulin (1 IU) is shown as '1 U'
-      if ( treatment.insulin < 1 && !treatment.carbs ) { // don't show the unit of measurement for insulin boluses < 1 without carbs (e.g. oref0 SMB's). Otherwise lot's of small insulin only dosages are often unreadable
+      var enteredBy = '' + treatment.enteredBy;
+      
+      if ( treatment.insulin < 1 && !treatment.carbs && enteredBy.indexOf('openaps') > -1) { // don't show the unit of measurement for insulin boluses < 1 without carbs (e.g. oref0 SMB's). Otherwise lot's of small insulin only dosages are often unreadable
          unit_of_measurement = '';
+         dosage_units = (dosage_units+"").replace(/^0/,"");
       }
       // remove leading zeros to avoid overlap with adjacent boluses
```
scottleibrand: Comment should move up too.

scottleibrand:
Is this counting the number of treatments on-screen?  Looks like the result is a larger font size for our typical number of visible treatments, but still readable.  Might be slightly better to increase the spacing between the circle and the text, but also maybe not worth worrying about.

sulkaharo:
how many treatments / day does SMB create for you / day on average?

scottleibrand:
Looks like we're averaging about 150 SMBs per day with no-bolus when the sensor is working well.  Only half that in the last 24h due to jumpy sensor and resulting less frequent larger SMBs.

sulkaharo:
Right, with those amounts this should produce exactly the same font size as your original change, so odd if you’re seeing something different.

scottleibrand:
Is it based on the number of treatments per day?  Or the number on-screen?

scottleibrand:
It looks like it's using the number on-screen, and reducing the font size as I shift from the 2h and 4h to the 12h and 24h graphs and it has to fit more treatments on-screen.  If that's the desired behavior, I think that's fine.

PR #3114, opened 2017-12-03
Fix bg units in updated Loop pill

dm61:


francescaneri:
@dm61 sensitivity can also be added? Thanks Matteo

dm61:
@francescaneri Loop currently does not upload ISF to NS, so we are not able to add that to the pill. I've added min-max and eventual predicted bg, with units correctly matching user preference. That's about as much as we can show in the pill with the current version of Loop.

sulkaharo:
@dm61 has this been tested / is this ready for merge?

dm61:
@sulkaharo yes, this one is ready to merge - thanks

PR #2350, opened 2016-12-16
Google Home integration

rickfriele:
Google Home integration using API.ai. Attached is zip file of the Nightscout agent I created which you can import. Once the agent has been created from the zip file, you'll want to create the webhook used to fulfill the voice commands. When editing the Nightscout agent in the left hand menu select "Fulfillment" and set the URL to https://<YOUR_SITE>/api/v1/googlehome. Make sure to set the `api_secret` header to the hashed value of your `api_secret` so it's able to read from the API. Once the webhook is created select "Integrations" in the left hand menu and turn on "Actions on Google". Choose an invocation name and a voice preference. I chose "Nightscout" as the invocation name. Select "Authorize" and verify your Google credentials to either test on your own Google Home or using the web simulator. The authorization lasts for 30 minutes before you need to reauthorize since this is just a development build. 

You can invoke the Nightscout agent by speaking "Hey Google, talk to Nightscout". The Nightscout agent will then introduce him or herself. You can then ask things like "What's my blood sugar?", "What's Ricks' blood sugar?", "What's my forecast?", "How is my pump battery?", "How many carbs do I have on board?", etc. Most of these questions require the corresponding plugin to be enabled besides the basic "What's my blood sugar?".

Here's a quick demo getting my current blood sugar and projected blood sugar: https://www.youtube.com/watch?v=W8K5rv7Fe4g

Nightscout Agent which can be imported directly to API.ai when creating a new agent: [Nightscout.zip](https://github.com/nightscout/cgm-remote-monitor/files/656530/Nightscout.zip)

PR #3123, opened 2017-12-04
Wip/google home

vivekvelso:


sulkaharo:
So actually - looks like most of the code currently named as Google Home specific adds functionality that could be shared between Home and Alexa & other similar integrations? Can we make these generic?

sulkaharo:
The original PR is still open, this was closed as it was simply a dupe https://github.com/nightscout/cgm-remote-monitor/pull/2350

kevinfmanning:
ty

PR #3154, opened 2017-12-08
Npm update 

PieterGit:
I did a npm update, because a newer version of NodeJS has been released (8.9.2).
No need to change the cgm-remote-monitor code, because we specify NodeJS as 8.9.x now.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -59,9 +59,9 @@
       "integrity": "sha1-/ts5T58OAqqXaOcCvaI7UF+ufh8="
     },
     "ajv": {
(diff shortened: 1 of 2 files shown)


PieterGit:
The openssl vulnerability seems rather small... https://www.openssl.org/news/secadv/20171207.txt
Deploys fine under Heroku.

stavlor:
Pretty sure azure is still only supporting up to 8.9.0 at the moment but 8.9.x should work

PieterGit:
@sulkaharo : i've been running with this update for 14 days without problems. can you merge it to dev?

PR #3190, opened 2017-12-22
Upstream npm bug `Cannot read property '0' of undefined` fixed. Use npm 5.6.x when available

PieterGit:
@elnjensen @sulkaharo 

NPM has a known bug in 5.5.1 that causes the install to randomly fail
During installation/deployment some users got an error `Cannot read property '0' of undefined`
The downgrade to npm 5.2.0 was due to many users having these issues with npm 5.5.1

Because of old versions in Azure the following ended up with these versions in the `package.json`
```
 "engines": {
    "node": "8.9.x",
    "npm": "5.2.0 || 5.3.0 || 5.0.3"
  },
```

See discussion at: https://gitter.im/nightscout/public?at=5a21fcea87680e6230be5df8
According to the upstream npm issue https://github.com/npm/npm/issues/17858 has been fixed in npm 5.6.0, so we should use npm 5.6.x again. Because npm might not be available we also settle for npm 5.2.0, 5.3.0 or 5.0.3 if npm 5.6.x is not available (for example on Azure).

Note: when upgrading you might need `rm -rf node_modules` to clean out the modules.


PieterGit:
Travis CI Unit tests ok, and deploys fine on Heroku.  Didn't find problems yet.

@elnjensen @sulkaharo  Is `"npm": "5.6.x || 5.2.0 || 5.3.0 || 5.0.3"` the best set, or has Azure upgraded their npm stack, so we can drop either 5.2.0, 5.3.,0 or 5.0.3 from this list?

PieterGit:
I noticed Travis still uses `npm 5.5.1` for some reason.

.travis.yml
```diff
@@ -2,7 +2,8 @@ language: node_js
 sudo: required
 dist: trusty
 node_js:
-  - "8.9.1"
+  - "8.9.3"
+before_install: if [[ `npm --version` != "5.6" ]]; then npm install -g npm@latest; npm --version; fi
```
sulkaharo: Does this work in Azure?

PR #3196, opened 2017-12-27
Update bower.json

blocklist:
In Field "Version", there are 2 " at the end, so the Value is not right:

`$ bower install
bower                       EMALFORMED Failed to read ./bower.json

Additional error details:
Unexpected string in JSON at position 60

`


sulkaharo:
Closing in favor of having removed bower.json altogether, as we've removed Bower and this was just a remnant that wasn't actually used

PR #3211, opened 2018-01-01
Add "Views" section to Readme

unsoluble:
The three available alternate views should be explicitly mentioned in the Readme.


sulkaharo:
Do all these views support mmol correctly? I recall there were some issues with some of the clock views, which probably should be fixed before adding to public docs

unsoluble:
Yup, automatic unit support on all of them.

Only issue I'm aware of is that bgclock.html doesn't yet have proper viewport scaling, but it's the one that's already linked directly in the main options panel, so it's already pretty public. (Fixing the CSS for that one is next on my list.)

PR #3213, opened 2018-01-01
Habanero update (#2)

tylerwilson317:
hadn't changed Azure setting to 8.5.0 until after I updated to Habanero.  My mistake.


tylerwilson317:
I just fixed my Azure setting to 8.5.0  ...never did that before updating the fork here.  oops

PR #3342, opened 2018-02-09
de fix

AdrianLxM:


PieterGit:
@AdrianLxM : applied to dev. Thanks for the PR

PR #3351, opened 2018-02-11
Autotune first day consumed to establish conditions

jpcunningh:
The standalone Autotune algorithms pull data from before the time period being tuned so the IOB and COB are accurately reflected at the beginning of the tuning period.

This change accomplishes the same thing by consuming the first day in the From / To fields to set the conditions going into the second day and onward that are to be tuned.

### DIFF
--- a/lib/report_plugins/autotune.js
+++ b/lib/report_plugins/autotune.js
@@ -184,6 +184,11 @@ autotune.report = function report_autotune(datastorage, sorteddaystoshow, option
     return;
   }
 
+  if (sorteddaystoshow.length < 2) {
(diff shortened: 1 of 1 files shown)


PieterGit:
@jpcunningh Can you confirm this PR is ready to merge to dev? Has this been tested by you or others?

jpcunningh:
This is for the autotune branch instead of dev.

@scottleibrand, what are your thoughts on the logic above that describes why the change is needed?

scottleibrand:
Does autotune in NS run 4am-4am, or midnight-midnight?  How many hours from the previous "day" are used/needed?

jpcunningh:
I verified it is midnight to midnight.

PieterGit:
@scottleibrand any objections to merge this to the autotune branch?

PR #3360, opened 2018-02-14
Fix rawbg calculations

jpcunningh:
Change the rawbg calculations to simply use the noise value to determine whether to use the calibrated filtered value vs the calibrated unfiltered value to display. It was creating a ratio that had the effect of exaggerating acceleration when BG was changing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,8 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
   An option plugin to enable adding foods from database in Bolus Wizard and enable .
 
 ##### `rawbg` (Raw BG)
-  Calculates BG using sensor and calibration records from and displays an alternate BG values and noise levels.
+  Calculates BG using sensor and calibration records from and displays an alternate BG values and noise levels. Defaults that can be adjusted with [extended setting](#extended-settings)
+  * `SMOOTH` (`false`) - If `SMOOTH` is false, rawbg will be calculated such that it exagerates direction changes to provide early visible indication of glucose trend changes.
 
 ##### `iob` (Insulin-on-Board)
   Adds the IOB pill visualization in the client and calculates values that used by other plugins.  Uses treatments with insulin doses and the `dia` and `sens` fields from the [treatment profile](#treatment-profile).
--- a/README.md
+++ b/README.md
@@ -320,6 +320,7 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
 ##### `rawbg` (Raw BG)
   Calculates BG using sensor and calibration records from and displays an alternate BG values and noise levels. Defaults that can be adjusted with [extended setting](#extended-settings)
   * `SMOOTH` (`false`) - If `SMOOTH` is false, rawbg will be calculated such that it exagerates direction changes to provide early visible indication of glucose trend changes.
+  * `UNFILTERED` (`false`) - If `UNFILTERED` is false, rawbg will be calculated using the raw filtered value. If `UNFILTERED` is true, rawbg will be calculated using the raw unfiltered value.
 
 ##### `iob` (Insulin-on-Board)
   Adds the IOB pill visualization in the client and calculates values that used by other plugins.  Uses treatments with insulin doses and the `dia` and `sens` fields from the [treatment profile](#treatment-profile).
--- a/README.md
+++ b/README.md
@@ -319,8 +319,10 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
 
 ##### `rawbg` (Raw BG)
   Calculates BG using sensor and calibration records from and displays an alternate BG values and noise levels. Defaults that can be adjusted with [extended setting](#extended-settings)
-  * `SMOOTH` (`false`) - If `SMOOTH` is false, rawbg will be calculated such that it exagerates direction changes to provide early visible indication of glucose trend changes.
-  * `UNFILTERED` (`false`) - If `UNFILTERED` is false, rawbg will be calculated using the raw filtered value. If `UNFILTERED` is true, rawbg will be calculated using the raw unfiltered value.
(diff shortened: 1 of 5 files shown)


jpcunningh:
Yes. We've been running it for about 3 months now.

I submitted this modification because the old code was always exaggerating SGV trajectory changes which was providing incorrect data to the user. 

Regarding when new sensors are inserted, I am running Lookout, which I changed to send a unity calibration value to NS when a new sensor is inserted. This allows continuing to send the SGV entries with the calibrated glucose value set to 5, and NS displays the raw values exactly as they come out of the transmitter rather than using the previous sensor's last know calibration value.

sulkaharo:
We have a lot of users still on G4 and other solutions; need someone to test how those behave

jpcunningh:
I agree.

I don't think it will behave poorly, but I know from at least one
conversation I've seen on the Loop group there are some who thought the
exaggerated trajectory changes were predictions, which they weren't. The
different behavior will be thought to be a loss of the predictive logic
rather than correcting an error.

PieterGit:
@jpcunningh We use a G4 but no raw and a single rig that is connected to the CGM by USB. I believe that will use the CGM readings from the device. 
 Is it possible to find testers for this feature, or should we postpone it to 0.11 Nightscout release?

@sulkaharo What side effects are you worried about with this PR?

jpcunningh:
Since the prior discussions, I've implemented sending raw values to NS from
Lookout during the sensor warm-up periods and the code works as expected.

I'm as confident as I can be that the code works correctly. I've used it
uninterrupted without issues since the PR was created.

I would recommend merging it to dev.

PieterGit:
@jpcunningh : i think this can be merged to dev, but would like a 👍  on it from the other core committers like @sulkaharo or @jasoncalabrese or @MilosKozak

Tornado-Tim:
Support the move to merge into dev as I want to test :)

jpcunningh:
I think we determine the rational for the original behavior over in gitter's oref0/hardware-dev channel, so more discussion is probably needed. 

It unsmoothes the raw data and doing so makes the white dots for raw bg visible more often and gives early indication of trend trajectory changes. Whereas, I am more interested in accurate plotting of the values rather than use them for trend indications.

PieterGit:
Based on @jpcunningh remark I postponed this to 0.11 milestone. I think this can be merged to dev as soon as 0.10.3 is out of the door.
@Tornado-Tim you can already test by using the https://github.com/jpcunningh/cgm-remote-monitor/tree/wip/fix-rawbg-calculations branch

jpcunningh:
@PieterGit, to move this forward, I modified it to be controllable by an extended setting for the rawbg plugin that defaults to existing behavior and added some text in the README to describe the new option.

I've tested the various option configuration scenarios and believe this is good to merge into dev.

jpcunningh:
I think I've missed a place that needs to be called with extended settings, so please don't merge just yet.

jpcunningh:
@PieterGit, it's functioning as intended, now. Without setting the new extended settings environment variables, rawbg continues to operate as it did. The new extended settings environment variables provide control over how the rawbg values are calculated.

jasoncalabrese:
I haven't tested this, and since I'm using the G6 now, but the code changes look good

PieterGit:
This PR has been integrated with the https://github.com/nightscout/cgm-remote-monitor/tree/needs_testing 
branch. Please test that branch and report at this issue or at nightscout/cgm-remote-monitor#4169

lib/plugins/rawbg.js
```diff
@@ -49,14 +56,18 @@ function init (ctx) {
...
     var cleaned = cleanValues(sgv, cal);
 
+    var prefs = rawbg.getPrefs(sbx);
+
     if (cleaned.slope === 0 || cleaned.unfiltered === 0 || cleaned.scale === 0) {
       raw = 0;
-    } else if (cleaned.filtered === 0 || sgv.mgdl < 40) {
+    } else if (cleaned.filtered === 0 || sgv.mgdl < 40 || prefs.unfiltered) {
       raw = cleaned.scale * (cleaned.unfiltered - cleaned.intercept) / cleaned.slope;
+    } else if (prefs.smooth) {
```
PieterGit: This part of the code will never be reached if `unfiltered==true`.  Is this inteded?

README.md
```diff
@@ -318,7 +318,9 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
   An option plugin to enable adding foods from database in Bolus Wizard and enable .
 
 ##### `rawbg` (Raw BG)
-  Calculates BG using sensor and calibration records from and displays an alternate BG values and noise levels.
+  Calculates BG using sensor and calibration records from and displays an alternate BG values and noise levels. Defaults that can be adjusted with [extended setting](#extended-settings)
```
PieterGit: can you elaborate on what the difference is between the default behaviour (both are false), and if either `SMOOTH` or `UNFILTERED` is true. Shouldn't be documented that one of these settings must set to be true (and not both)?

PieterGit: Can you elaborate on what setting you would advice in what situation?

lib/plugins/rawbg.js
jpcunningh: Yes, the smooth preference allows the user to select how the raw `filtered` data is displayed, smooth or exaggerated.

If the user wants either of those plotted (both have some type of filtering applied), that would be opposite of the user wanting the raw `unfiltered` values plotted.

jpcunningh:
@PieterGit, I changed it to a tri-state variable and updated the README. Great suggestion.

Does the new setting help? Thanks!

PieterGit:
This PR has been tested as part of the `need_testing` branch of https://github.com/nightscout/cgm-remote-monitor/pull/4169
@blocklist_twitter @balshor @sulkaharo @jpcunningh @unsoluble @veryfancy @balshor @tynbendad have run this PR for a week.
No negative stuff popped up. Gitter link: https://gitter.im/nightscout/public?at=5c2f33d12863d8612bd78eb0

If you encounter success or failure with this PR, please reopen the issue and share your findings.

Closing PR, will be merged with https://github.com/nightscout/cgm-remote-monitor/pull/4169

PR #2778, opened 2017-08-30
Fix "disaplyed" typo

dbeasy:
Fix typo for hover tip for "Raw BG" display in the static/index.html (settings drawer).

I'm new to git and can't find a branch/fork of mine to merge and to not include the  static/report/index.html changes too. Those don't have to be changed in dev or master. I didn't focus group test changing the sorting. I just use the "treatments" tab a lot to fix incomplete documentation from the pebble watch and I like "most recent" at the top.

Thanks

PR #3369, opened 2018-02-18
Fix "disaplyed" typo

PieterGit:
cherry picking fixes from https://github.com/nightscout/cgm-remote-monitor/pull/2778


PieterGit:
The English key was ok, so the key wasn't translated.

PR #3400, opened 2018-02-26
Fix tooltips in drawer

PieterGit:
Tooltips in the drawer don't work.  They don't show up.
I think it's caused by the bower to webpack upgrade.
I migrated https://www.npmjs.com/package/jquery.tipsy to https://www.npmjs.com/package/jquery.tooltips
because that has webpack support.

I hoped that fixed the probem, but unfortunately it does not work yet.  

@sulkaharo : can you have a look if you can spot the error?

### DIFF
--- a/bundle/bundle.source.js
+++ b/bundle/bundle.source.js
@@ -1,7 +1,7 @@
 import '../static/css/drawer.css';
 import '../static/css/dropdown.css';
 import '../static/css/sgv.css';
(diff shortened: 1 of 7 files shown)


PieterGit:
@sulkaharo : they didn't show up because a browser caching issue. I can now confirm that this PR fixes the issue. Can you please review, so that we can merge it to dev?

PR #3403, opened 2018-03-01
adding Loop as automated uploader

Kdisimone:
Loop is also an automated uploader of treatments.  Had an issue where temp basals set by Loop and uploaded to treatments were auto-snoozing BG alarms.  This PR resolves the error shown below.

screen-shot-2018-02-28-at-10 12 34-am

### DIFF
--- a/lib/plugins/treatmentnotify.js
+++ b/lib/plugins/treatmentnotify.js
@@ -15,13 +15,13 @@ function init() {
     , pluginType: 'notification'
   };
 
-  //automated treatments from OpenAPS shouldn't trigger notifications or snooze alarms
+  //automated treatments from OpenAPS or Loop shouldn't trigger notifications or snooze alarms
   function filterTreatments (sbx) {
     var treatments = sbx.data.treatments;
 
     treatments = _.filter(treatments, function notOpenAPS (treatment) {
       var ok = true;
-      if (treatment.enteredBy && treatment.enteredBy.indexOf('openaps://') === 0) {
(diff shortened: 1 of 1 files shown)


lib/plugins/treatmentnotify.js
```diff
@@ -15,13 +15,13 @@ function init() {
...
 
-  //automated treatments from OpenAPS shouldn't trigger notifications or snooze alarms
+  //automated treatments from OpenAPS or Loop shouldn't trigger notifications or snooze alarms
   function filterTreatments (sbx) {
     var treatments = sbx.data.treatments;
 
     treatments = _.filter(treatments, function notOpenAPS (treatment) {
       var ok = true;
-      if (treatment.enteredBy && treatment.enteredBy.indexOf('openaps://') === 0) {
+      if (treatment.enteredBy && treatment.enteredBy.indexOf('openaps://' || 'loop://') === 0) {
```
dabear: This is not correct @Kdisimone . You cannot use OR on a string. Suggested fix:

if (treatment.enteredBy && (treatment.enteredBy.indexOf('openaps://') === 0 || treatment.enteredBy.indexOf('loop://'))

Kdisimone: Thank you. I actually thought I’d closed this out because when I tested it the fix hadn’t worked. I’ll make the edit in the next couple hours and comment back. Thank you!!

Kdisimone: Or rather submit a new PR 😉

PR #3405, opened 2018-03-01
Update language.js

hartlijn:
added NL language on missing codes

PR #3406, opened 2018-03-02
Revert "Update language.js"

PieterGit:
Reverts nightscout/cgm-remote-monitor#3405


PieterGit:
Wrong base. Should have been applied to dev first. Reverting and will apply it to dev.

PR #3423, opened 2018-03-06
mgdl -> mg/dl

AdrianLxM:
Nighscout according to @MilosKozak and from what it sends to AndroidAPS uses "mg/dl", not "mgdl"

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
   Treatment Profile Fields:
 
   * `timezone` (Time Zone) - time zone local to the patient. *Should be set.*
-  * `units` (Profile Units) - blood glucose units used in the profile, either "mgdl" or "mmol"
+  * `units` (Profile Units) - blood glucose units used in the profile, either "mg/dl" or "mmol"
   * `dia` (Insulin duration) - value should be the duration of insulin action to use in calculating how much insulin is left active. Defaults to 3 hours.
   * `carbs_hr` (Carbs per Hour) - The number of carbs that are processed per hour, for more information see [#DIYPS](http://diyps.org/2014/05/29/determining-your-carbohydrate-absorption-rate-diyps-lessons-learned/).
   * `carbratio` (Carb Ratio) - grams per unit of insulin.


PieterGit:
@AdrianLxM : the code uses `mgdl` for `mg/dL` and `mmol` for `mmol/L`.  So this PR will currently break the units setting. Can you change it so that it is clear that you should use `mgdl` or `mmol`, but that the units are `mg/dL` and `mmol/L`.

Note that liter is often abbreviated with a capital `L`, to avoid confusion with a `1` or `l`.

See https://github.com/nightscout/cgm-remote-monitor/search?utf8=%E2%9C%93&q=units+mgdl+mmol&type= for occurences of the unit in the current code.

README.md
```diff
@@ -483,7 +483,7 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
   Treatment Profile Fields:
 
   * `timezone` (Time Zone) - time zone local to the patient. *Should be set.*
-  * `units` (Profile Units) - blood glucose units used in the profile, either "mgdl" or "mmol"
+  * `units` (Profile Units) - blood glucose units used in the profile, either "mg/dl" or "mmol"
```
PieterGit: Something like. Use `mgdl` for `mg/dL` and use `mmol` for `mmol/L`

AdrianLxM:
The profiles created by NightScout use "mg/dl" and "mmol". This is why AndroidAPS checks for them.
Having "mgdl" and "mmol" in the documentation lead to the problem that the 600series uploader uploaded "mgdl" and was not detected.

Storing the profile again from Nightscout leads to the correct (?) "mg/dl" that will be detected correctly.

AdrianLxM:
An example from the tests where units is set to "mg/dl": https://github.com/nightscout/cgm-remote-monitor/blob/9fce5db976f3a3e00087014b81b212213cf25e88/tests/api.treatments.test.js#L39

So it seems that like with the display units, the internal representation is "mg/dl" and "mmol".

AdrianLxM:
Any news to this?
Using "mgdl" now after years of using "mg/dl" would break many apps working with NightScout. This is why I suggest only updating the documentation.

sulkaharo:
Nightscout code uses "mgdl" for the internal representation and for interpreting this variable, so if we change the documentation, we need to change multiple parts in the logic to understand both formats. The user-facing presentation uses "mg/dl". What's the specific API that AndroidAPS uses that sees "mg/dl"?

PieterGit:
@AdrianLxM can you have a look at @sulkaharo question above. If not answered, I think I'll postpone this to Nightscout 0.11 as it's not clear of what that best way is to fix this for both Nightscout and AndroidAPS

PieterGit:
@AdrianLxM Postponed this to 0.11.0 because we're about to release a new version and this change probably needs more thought. Please have a look at my question of two weeks ago and drop in at https://gitter.im/nightscout/public to discuss this

PieterGit:
@AdrianLxM and @MilosKozak : can we close this PR or does is there anything we can do know to resolve it?

PieterGit:
@AdrianLxM and @MilosKozak last call for this PR. If there is no response soon I'll close this due to inactivity. Postponing this to 0.12 release for now.

@sulkaharo question of May 18th has not been answered, see https://github.com/nightscout/cgm-remote-monitor/pull/3423#issuecomment-390104500

AdrianLxM:
I neither use Nightscout at the moment nor the 600 Series Uploader.

Back when I reported that either the behavior of NS or the documentation was wrong as they didn't fit.

PR #3464, opened 2018-03-23
bugfix/issue3462

tanja3981:


gonium:
I ran into the same issue (invalid ID when AndroidAPS recorded a profile change) - this pull request fixed it.

tanja3981:
Would it be possible to get this bugfix merged into dev, as more and more people experience this problem? 
If not, cause someone explain to me why the pull request isn't approved so that I could change whatever necessary? 
@sulkaharo @jasoncalabrese

Issue #3499, opened 2018-04-04
Web page view leaks memory and eventually crashes


jimrandomh:
If you leave the web-page view of Nightscout open, memory usage of the tab it's in gradually increases until eventually the tab becomes unresponsive and dies.

The renderer seems to be creating duplicate SVG elements (or alternatively, not deleting old SVG elements) and this is at least part, maybe all of the problem. I am working on a pull request to address it.

PieterGit:
@jimrandomh I assume https://github.com/nightscout/cgm-remote-monitor/pull/3500 fixed this issue? If not, please feel free to reopen this issue.

PR #3500, opened 2018-04-05
Fix a memory leak in Nightscout chart

jimrandomh:
The chart would create SVG objects for treatment event records, then later recreate them without deleting the old ones. This resulted in a huge number of duplicates accumulating, which (in my setup) would cause a crash due to excessive memory usage and DOM size in about a day.

Tested by deploying to my personal install and poking around. This corresponds to issue #3499.


sulkaharo:
Yay great! Testing

jimrandomh:
(codacy-bot comment is because I accidentally put some unrelated commits on the same branch as this, then undid them.)

PR #3584, opened 2018-05-19
Fixes in german language

winni67:
The nightscout answer already contains these deleted words.


winni67:
Only two small changes in one file. Is it possible to merge this into dev soon? Thanks in advance.

PR #3601, opened 2018-05-25
Add token to nav URLs

schmitzn:
When you access a nightscout page with a token, the token is not included in the urls of the nav menu.  For example, if you have the index page opened, you cannot switch to the reports page without modifying the url yourself.

I've added a few lines to the browser-settings.js to fix this. I've left out the admin page because I don't know whether it makes sense to access that page via token.

This is my first PR (except for translations). I hope I haven't missed anything important.

### DIFF
--- a/lib/client/browser-settings.js
+++ b/lib/client/browser-settings.js
@@ -94,6 +94,13 @@ function init (client, serverSettings, $) {
 
     $('#editprofilelink').toggle(settings.isEnabled('iob') || settings.isEnabled('cob') || settings.isEnabled('bwp') || settings.isEnabled('basal'));
 
+    var token = client.browserUtils.queryParms().token;
+    if (token) {
+      $('#showreportslink').attr("href", $('#showreportslink').attr("href") + '?token=' + token);
+      $('#editprofilelink').attr("href", $('#editprofilelink').attr("href") + '?token=' + token);
+      $('#editfoodlink').attr("href", $('#editfoodlink').attr("href") + '?token=' + token);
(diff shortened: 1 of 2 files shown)


PieterGit:
@schmitzn thanks for the PR. I think the token should work for an hour with the jsonwebtoken. I don't know why this isn't the case.  I doubt this is the way to fix it, because there are a lot more URL's.

@sulkaharo do you know why the user can't navigate when using a token?

PieterGit:
Postponing this to 0.11 because I think it needs more work than proposed in this PR.

PieterGit:
It's a pity I need to do this, but I'm postponing this to 0.12 release. Adding the token is not the solution, because people can also be authentication with `API_SECRET` or `jsonwebtoken`. Adding the needs help label, hoping somebody can spend some time to dig into this issue.

schmitzn:
@PieterGit tested it with current ``dev`` branch, the token is still being ignored in navigation. I've applied the changes of this PR to my own installation which fixes it for me. But I think this issue could be relevant for other users as well.

funkstille:
@schmitzn 

You missed one link for the colorclock... just for completeness... I did not use that feature ;)

$('#clockcolorlink').attr("href", $('#clockcolorlink').attr("href") + '?token=' + token);

Your fix works fine for me. 

Only the hardcoded close ("X") buttons f.e. in Reports page do not pass the link backwards.
But that also depend on my browser setting, since I always open links to new tabs.
And since it is hardcoded it is not that easy to address it with a link "manipulation" in browser-settings.

sulkaharo:
This was implemented in a separate PR

PR #3636, opened 2018-06-19
Update to node 6.11.1

carrfl:


PieterGit:
@carrfl thanks for the PR. Instead of upgrading Node from 0.11 to 6.11.1 we should release the dev quite soon. That works with Node 8 and Node 10.

@sulkaharo Agree?

PR #3476, opened 2018-03-27
Various updates to fix Azure deploys.

telma2006:
Fix npm version 5.6.x.
Reverse the --production and --scripts-prepend-node-path arguments on npm install.
Use graceful-fs@4.1.11 to solve missing internal/util/types in postinstall.
Remove blank line in webpack.config.js as a dummy commit for postinstall testing.

PR #3642, opened 2018-06-25
Various updates to fix Azure deploys.

balshor:
Fix npm version 5.6.x.
Reverse the --production and --scripts-prepend-node-path arguments on npm install.
Use graceful-fs@4.1.11 to solve missing internal/util/types in postinstall.
Remove blank line in webpack.config.js as a dummy commit for postinstall testing.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -113,7 +113,7 @@ selectNodeVersion
 # 3. Install npm packages
 if [ -e "$DEPLOYMENT_TARGET/package.json" ]; then
   cd "$DEPLOYMENT_TARGET"
-  eval $NPM_CMD install --scripts-prepend-node-path --production
+  eval $NPM_CMD install --production --scripts-prepend-node-path
   exitWithMessageOnError "npm failed"
   cd - > /dev/null
 fi
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4,269 +4,11 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@mrmlnc/readdir-enhanced": {
-      "version": "2.2.1",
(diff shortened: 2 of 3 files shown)


balshor:
Tests are passing locally:
    # tests 288
    # pass 288
    # fail 0

Not sure why they didn't pass on https://github.com/nightscout/cgm-remote-monitor/pull/3476

PieterGit:
@balshor : thanks for the PR.

Can you elaborate why you need the dependency for `graceful-fs`. It pulls quite some other packages. According to it's pages it a drop in replacement for `fs`. I wonder if it's not an npm issue your facing on Azure, and if it wouldn't be better to either to upgrade to node 5.8 or 6.1 (if that's available on Azure).

In Travis `.travis.yml`. we have this line: 
`before_install: if [[ `npm --version` != "5.8.0" ]]; then npm install -g npm@latest; npm --version; fi` 
to upgrade npm 5.6 because we had other issue with it.

Will perhaps setting ` "npm": "6.1.x || 5.8.x || 5.7.x || 5.6.x"` to ` "npm": "6.1.x || 5.8.x` in `package.json`   fix this issue on Azure? I rather will do that than adding packages which we don't reference in our software.

@sulkaharo or @MilosKozak any ideas?

balshor:
It looks like I can update Nightscout with npm 5.6.0 now.  I've removed graceful-fs from this PR.

There was a race condition in earlier versions of npm that cause issues with uninstalling fsevents, which is Mac OS X only.  graceful-fs has a bunch of better fs management on Windows, so that was avoiding the issue for me.  See https://github.com/npm/npm/issues/17671 for the gory details.

FWIW, Azure App Services does not currently have any npm versions more recent than 5.6.0.

PR #3700, opened 2018-07-15
fix the openaps/loop auto-trearment filter

jasoncalabrese:


jasoncalabrese:
I first made this based on my old branch, and then after talking to @Kdisimone found that she had a PR (https://github.com/nightscout/cgm-remote-monitor/pull/3403) to fix the same issue.  That PR was merged by mistake since the `indexOf` wasn't right.

PR #3706, opened 2018-07-18
Fixed Swedish typos

whooze:


PieterGit:
@whooze thanks for the fix. This will be in the 0.10.3 release

PR #3707, opened 2018-07-18
upgrades, moment-locales-webpack-plugin integration, easier bundle-analyzer, dutch translations

PieterGit:
1. webpack 4.16.0 gives problem with report test on Windows

```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments:
[0] _isAMomentObject: true, _isUTC: true, _useUTC: true, _l: undefined, _i: T00:00:00, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (E:\Git\src\cgm-remote-monitor\tmp\js\bundle.js:1:5989)
```

fixed it with an upgrade to 4.16.1

2. a lot of moment-locales are packacked with webpack. Limit to language that are available in Nightscout. Note that `dk` is not available in moment.js. This decreases
      js/bundle.js from 1.86 MiB to 1.72 MiB. 

3. make it possible to use `npm run bundle-analyzer` so that webpack.config.js does not have to be changed in order to run the bundle-analyzer

4. add missing Dutch translations

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -9974,6 +9974,7 @@ function init() {
       ,ru: 'Избыток инсулина равного %1U, необходимого для достижения нижнего целевого значения, углеводы не будут учтены'
       ,sv: 'Överskott av insulin motsvarande %1U mer än nödvändigt för att nå lågt målvärde, kolhydrater ej medräknade'
       ,nb: 'Insulin tilsvarende %1U mer enn det trengs for å nå lavt mål, karbohydrater ikke medregnet'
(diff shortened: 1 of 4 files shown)


PieterGit:
More info on 2. the moment-locals.

Here's an overview of the bundle before the PR:
image

And here is an overview of the bundle after this PR:
image

PieterGit:
Detail: 

Before: 
image

After: 
image

PR #3708, opened 2018-07-18
Made swedish 100% + a few corrections

whooze:
**Found these strings on my site, but not in languages.js**

* "temp basals must be rendered to display this report"
* "netIOB stats"
* "Pump Battery Change"

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -1816,7 +1816,7 @@ function init() {
       ,fr: '% de valeurs'
       ,el: '% των μετρήσεων'
       ,pt: '% de valores'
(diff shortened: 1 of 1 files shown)


lib/language.js
```diff
@@ -9870,6 +9870,7 @@ function init() {
       ,fr: 'Information nécessaire manquante'
       ,ru: 'Отсутствует необходимая информация'
       ,bg: 'Липсва необходима информация'
+      ,sv: "Nödvändig information saknas'
```
PieterGit: please fix this quote. Use single `'` instead of `"`

PR #3207, opened 2018-01-01
More clock options

unsoluble:
Should have links to both bgclock.html and clock-color.html in the main link panel. Might be a tidier way of putting them both on one line, but didn't want to mess with the CSS.

PR #3709, opened 2018-07-19
speech and  Warning, Pump Battery Low Alarm issue during Clock testing...

PieterGit:
During testing of https://github.com/nightscout/cgm-remote-monitor/pull/3207 I found several issues that need attention:

1. If speech plugin is not enabled it will cause problems in the console, e.g.
`index.js:600 Uncaught TypeError: Cannot read property 'speech' of undefined`

2. I had serious problems with `Warning, Pump Battery Low` alarm. 
It was is audible, but cannot be surpressed or disabled in the UI. I started adding the UI for disabling it, but it does not seem to work yet. There is some strange issue with these pump alarms. Temporary workaround is `PUMP_ENABLE_ALERTS=false`, but of course `PUMP_ENABLE_ALERTS=true` should not lead to alarms that cannot be stopped and cannot be disabled in the UI.

3. The clock views don't work with `AUTH_DEFAULT_ROLES=denied`. Blank screen and: `/api/v1/entries/current.json (401 Unauthorized)`  and `/api/v1/status.js 401 (Unauthorized)`. In the console: `DENIED no-token api:status:read`


PieterGit:
Still an error when `speech` is not in `ENABLE` 

image

unsoluble:
Getting the clock views to respect the authentication roles is beyond my ability. But they never did, so that'd be new functionality regardless. I'd suggest saving that for 0.11, unless it's a quick fix.

PieterGit:
@unsoluble I know it's a different problem, so therefore I created a new PR.

I just found out that problem 1 and 2 are related. If I disable the speech code in `lib/client/index.js`
https://github.com/nightscout/cgm-remote-monitor/pull/3709/files#diff-5f3d9018a260bb33cee03e0c67fdf95bR600 then the pump battery low alarm works as expected. That is: I can postpone the alarm.

Testcase: Use a near empty pump battery that will trigger `Warning, Pump Battery Low alarm.`. Remove `speech` from `ENABLE` and the pump low battery alarm will be triggered you will not be able to postpone the alarm. This is a blocker for 0.10.3.

I have no problem in postponing issue 3 (clock view doesn't work) to 0.11.

I'm assigning this issue to @sulkaharo and I hope he has time to look into this. Perhaps postponing speech plugin implementatie to 0.11 is an option as well.

PieterGit:
@sulkaharo I would suggest to postpone the pump battery low alarm to 0.11. I was having troubles with disabling alarms so I noted this alarm was not configurable yet.

PR #3724, opened 2018-07-24
Server-side localization fixes, better speech api use

sulkaharo:


sulkaharo:
I should still commit a change that removes the require(levels) from all plugins in favor of using the one in sbx

lib/client/index.js
```diff
@@ -770,8 +774,9 @@ client.load = function load(serverSettings, callback) {
       console.info('generating timeAgoAlarm', alarm);
       container.addClass('alarming-timeago');
       var display = client.timeago.calcDisplay(client.sbx.lastSGVEntry(), client.sbx.time);
+      var translate = client.translate;
       var notify = {
-        title: 'Last data received ' + [display.value, display.label].join(' ')
+        title: translate('Last data received') + ' ' + display.value + ' ' + translate(display.label)
```
PieterGit: Why not reuse 'Last entry {0} minutes ago' or create a 'Last data received {0} minutes ago`

sulkaharo: If you check the logic for timeago plugin, it actually returns dynamic time units based on how much time has passed, so once the time goes over 60 minutes, the system says "1 hour ago". Not sure if this really is needed though.

PieterGit: I have no problems with leaving it like this for now.

PR #3737, opened 2018-07-27
Update Italian Translation

hichameyessou:


PieterGit:
@hichameyessou thanks for updating the italian language

PR #3744, opened 2018-07-28
Bulk Devicestatus Delete

jpcunningh:
This adds a new admin tools button to delete devicestatus records older than 1 month rather than deleting the entire collection.


lib/server/devicestatus.js
```diff
@@ -69,6 +70,72 @@ function storage (collection, ctx) {
...
+    var dayStr;
+    var retVal = null;
+
+    var firstErr = null,
+        numDeletes = 0,
+        totalDeletes = 0,
+        totalRecordsDeleted = 0;
+
+    var endDate = moment(date);
+    endMonth = endDate.month() + 1;
```
scottleibrand: I'm not seeing how this stops at one month ago.  The way I'm reading it, it deletes up to the current month and day?  What am I missing?

jpcunningh: @scottleibrand, the date argument to the function is the day to delete up to. `moment(date)` converts the date string to a moment object so if the user sent an HTTP DELETE request to `$NS_URL/api/v1/devicestatus_delete/2018-02-12`, `endDate` would be set to `02/12/2018`.

jpcunningh:
@PieterGit, can you point me to a location I can read on how and where are the swagger files used so I can test updates to them? 

Thank you!

PieterGit:
look here:
- https://github.com/nightscout/cgm-remote-monitor/blob/dev/swagger.json
- https://github.com/nightscout/cgm-remote-monitor/blob/dev/swagger.yaml
- https://editor.swagger.io/
- https://github.com/OAI/OpenAPI-Specification

You can create the json with the swagger editor ui based on the yaml.
The API should be available on, https://YOUR-SITE.com/api-docs.html 
but I see it doesn't work at this moment, so we need to fix that bug 🐛

jpcunningh:
Rather than a new api, I am closing this and opening a new one that enhances the capability of the existing devicestatus api.

PieterGit:
The user documentation for this feature can be improved. I logged issue nightscout/documentation#2 for that.

PR #3745, opened 2018-07-28
Make bold in tooltips bolder and only show 3 digits for pump reservoir

PieterGit:
@jasoncalabrese reports differences in bold labels in tooltips. https://gitter.im/nightscout/public?at=5b5c2209bf75446606694d14

I examined them with Chrome (use Inspect, hover over a tooltip and use `F8` to pause Javascript and find  a word from the tooltip and find it in the HTML). I found the following differences between jquery.tipsy and jquery.tooltips.:
- some padding differences 
- difference with rounded corners 
- word-wrap: break-word
I fixed them in this PR and made the boldness (`font-weight`) of the tooltips pill labels `900` (very bold).

Example Chrome view (0.10.3-dev before this PR):
image

The CSS path is `html body div.tooltip strong`

Second problem fixed in this PR is a wrapping problem with the pump pill because the word `ago` in dutch is rather long (`geleden`)  and causes the pills to wrap. By only showing three significant digits for the pump resorvoir, this is solved (e.g. 105U, 90.1U, 8.12U, 0.123U). 

@jasoncalabrese and @sulkaharo : please review

### DIFF
--- a/static/css/jquery.tooltips.css
+++ b/static/css/jquery.tooltips.css
@@ -1,9 +1,13 @@
+.tooltip strong, b {
+    font-weight: 900
+}
 .tooltip {
   position: absolute;
   font-size: 10px;
-  padding: 3px 8px;
+  padding: 3px 5px;
   opacity: 0.9;
   z-index: 9999;
+  word-wrap: break-word;
 }
 .tooltip .tooltip-inner {
   background-color: #000;
@@ -12,6 +16,9 @@
   padding: 3px 8px;
   text-align: center;
   box-shadow: 0 0 5px 0 black;
+  border-radius: 3px;
+  -moz-border-radius: 3px;
+  -webkit-border-radius: 3px;
 }
 .tooltip .tooltip-arrow {
   position: absolute;
--- a/static/css/jquery.tooltips.css
+++ b/static/css/jquery.tooltips.css
@@ -1,5 +1,5 @@
-.tooltip strong, b {
-    font-weight: 900
+.tooltip strong, .tooltip b {
+    font-weight: 700
 }
 .tooltip {
   position: absolute;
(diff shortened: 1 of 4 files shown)


static/css/jquery.tooltips.css
```diff
@@ -1,9 +1,13 @@
+.tooltip strong, b {
```
jasoncalabrese: I think this should be `.tooltip strong, .tooltip b {`, but it's not having any effect for me.  I don't think we have an 900 weight font.

I tried changing https://github.com/nightscout/cgm-remote-monitor/blob/52e1ff054c98a6c0ee90d6f3b4d5c299e2b6dd40/static/css/main.css#L2 to 
`@import url('https://fonts.googleapis.com/css?family=Ubuntu:400,700');`, based on https://fonts.google.com/specimen/Ubuntu?selection.family=Ubuntu:400,700 using `font-weight: 700`, thats the heavyist they seem to have for the font.  In the old version that I'm still running we were using `font-family: 'Open Sans', Helvetica, Arial, sans-serif;` on the body element.  I think @sulkaharo did some [optimizations](https://github.com/nightscout/cgm-remote-monitor/commit/936d631870132ac931daaa75d04aa46961bcc5c7), but now it's https://github.com/nightscout/cgm-remote-monitor/blob/52e1ff054c98a6c0ee90d6f3b4d5c299e2b6dd40/static/css/main.css#L74

PieterGit: you're right. fixed and merged your PR in. I don't know if we use `b` within `tooltip`s

PieterGit:
@jasoncalabrese I merged the fixes from https://github.com/nightscout/cgm-remote-monitor/pull/3746 into this PR. Can confirm that bold fonts show up ok. Agree to merge this PR?

PR #3751, opened 2018-07-30
Log error on device status insert & don't crash

sulkaharo:
(Don't merge yet!)

### DIFF
--- a/lib/api/profile/index.js
+++ b/lib/api/profile/index.js
@@ -44,7 +44,6 @@ function configure (app, wares, ctx) {
                     res.json(created.ops);
                     console.log('Profile created', created);
                 }
-
             });
         });
 
--- a/lib/authorization/storage.js
+++ b/lib/authorization/storage.js
@@ -28,6 +28,10 @@ function init (env, ctx) {
         obj.created_at = (new Date()).toISOString();
       }
       collection.insert(obj, function (err, doc) {
+        if (err != null && err.message) {
+          console.log('Data insertion error', err.message);
+          return;
+        }
         storage.reload(function loaded() {
           fn(null, doc.ops);
         });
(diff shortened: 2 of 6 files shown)


sulkaharo:
Is there a good way to surface this to users?

jasoncalabrese:
I think we should still return a 500 so the client deal with it.  Is this mostly for device status? I was wondering it we should rename the old collection on startup if it's too big, then start using a new collection and ignore the old.

sulkaharo:
At least for mLab, looks like all inserts to all collections start failing at the same time, so it’s not specific to the device status collection, other than it being the collection that seems to get the most data.

sulkaharo:
@jasoncalabrese so, instead of just returning, throw the error?

PieterGit:
@jasoncalabrese @sulkaharo 
I think logging  an error for 0.10.3 in the console is better than crashing Heroku. We can do the 500 server error stuff in 0.11. Thoughts?

jasoncalabrese:
What's the behavior when the db is full "crashing heroku" sounds extreme...

I think to return a 500, we just need to call the callback passing an error something like `fn(new Error('insert failed'), null)` instead of the `fn(null, doc.ops)` that gets call on success

PieterGit:
@jpcunningh I don't mind if we PR this with this PR or with https://github.com/nightscout/cgm-remote-monitor/pull/3989 . I assume @jasoncalabrese and @sulkaharo are quite busy at this moment, because I haven't seen them a lot the last few weeks.  I will accept any tested PR that will log an error and return 500 to the client 😄

jpcunningh:
@PieterGit, I created a new PR #4004 that returns 500 to the client.

PieterGit:
Merged https://github.com/nightscout/cgm-remote-monitor/pull/4004 instead. Closing this PR

PR #3738, opened 2018-07-27
Add more Hebrew localizations

kretch:
Changed a few numbers to words in the Hebrew translation

PR #3754, opened 2018-07-31
add hebrew translations

PieterGit:
fixes https://github.com/nightscout/cgm-remote-monitor/pull/3738 (and resolve merge conflicts and fix spacing).

Please mind the spacing of the file and make sure you use the last `lib/language.js` and see if there are open language translations PR's. In this case a lot of people added translations, so I fixed it for you.

@kretch thanks for creating the PR. Can you add the missing hebrew translations?


PieterGit:
@kretch. Thanks for the translations. There are still a lot of strings that need to be transldated. Can you translate more?

PR #3800, opened 2018-08-07
habanero purple

patik2:


herzogmedia:
what are you trying to accomplish with all those prs? If you want to update your nightscout instance, pull **from** the nightscout master **into** your master not the other way around. You can always ask for help in the gitter-chat as well: https://gitter.im/nightscout/public

Issue #3784, opened 2018-08-06
NS fails to load under iOS 9.3.5 or earlier


unsoluble:
Looks like maybe a problem in the Node bundle, but I'm not familiar enough with that stack to suggest anything else. Error being caught by the console is here:

![screen shot 2018-08-06 at 9 55 11 am](https://user-images.githubusercontent.com/801582/43730142-39682f06-995f-11e8-98c5-ddf3a921bdd3.png)
This failure appears to occur on everything prior to and including iOS 9.3.5. Same sites run fine under iOS 10.3.3 and above; the break likely occurs at the 9-10 transition.

PieterGit:
@redcyclist1 also reports this for Ipad3, see https://gitter.im/nightscout/public?at=5b687b46e8698a2dca165f55

PieterGit:
@unsoluble @RedCyclist1 note that older apple devices don't have security updates and 
therefore your personal details aren't completely safe/secure. I think we should warn people 
with older browsers/devices and we can't keep support for older devices indefinitely. 

I did some research for you. I think this is caused by these dependencies:
```
+-- webpack@4.16.3
| +-- node-libs-browser@2.1.0
| | +-- buffer@4.9.1
``` 

The javascripts comes from the `buffer` package. They still seem to support  iOS 9+.
You might help by creating an issue on the upstream project https://github.com/feross/buffer
If there is a buffer version that has the fix this version must be included by the `node-libs-browser` project and that `node-libs-browser` must be used by `webpack`.

PieterGit:
@unsoluble @redcyclist1 can you give https://github.com/nightscout/cgm-remote-monitor/pull/3806 a try? I tested it on iphone 4s and it works, but I'm not sure reverting to webpack 3 is a good idea.

So it seems to be a webpack problem we're hitting, because 
```
+-- webpack@3.12.0
| +-- node-libs-browser@2.1.0
| | +-- buffer@4.9.1
```
works on ios 9.

unsoluble:
Works for me on a 9.3.5 device.

I'm still hoping someone can determine whether or not the entire bundle.js is actually required for clock views. Paring that down would be a fair solution that allows for keeping up with the webpack advances but also lets the simple displays run on older hardware.

PieterGit:
[...]

PR #3806, opened 2018-08-07
add browserupdate and make 0.10.3 work on ios9 (including webpack 3 downgrade)

PieterGit:
fix for https://github.com/nightscout/cgm-remote-monitor/issues/3784

reverts to webpack@3 and adds browers-update message on older browsers.

tested with iphone 4s (ios 9.3.5)

not ready to merge, drops momentLocales fix for now and need to make changes for mode=development

webpack@3 does not support ES6, which xdrip-js.js plugin already used.

### DIFF
--- a/bundle/bundle.source.js
+++ b/bundle/bundle.source.js
@@ -4,7 +4,6 @@ import '../static/css/sgv.css';
 
 
 $ = require("jquery");
-
 require('jquery-ui-bundle');
 
 window._ = require('lodash');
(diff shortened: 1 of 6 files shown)


PieterGit:
Closing PR. Webpack 3 was not the cause for not working. The xdrip js was using ES6 syntax (`let`). 
Please use https://github.com/nightscout/cgm-remote-monitor/pull/3863 instead.

PR #3819, opened 2018-08-09
Format webpack config & exclude node_modules directory

nielsvanvelzen:
I started using Nightscout a few weeks ago but when I tried to update to the latest version a few days ago it didn't work because my server (that was compiling it) only had 512MB memory.

After some digging I found out that this was caused by the webpack config not excluding `node_modules` from the ruels for image + css loading.

When testing on my Windows machine memory usage went from a peak of 1.5GB to 200MB. Compile times got speed up too.

After testing I can confirm that everything is still working (tested the admin & clocks in particular)


PieterGit:
@nielsvanvelzen thanks for finding this bug. Please target PR's at the `dev` branch. I will test this PR later or perhaps @sulkaharo beats me.

PieterGit:
Reimplemented with https://github.com/nightscout/cgm-remote-monitor/pull/3953 (without reformatting). This PR signficantly reduces memory usage and built time:
```
  Δt 109764ms (436 modules hidden) [without this PR]
  Δt 22199ms (436 modules hidden) [with this PR]
```

Closing this PR. We'll merge it with https://github.com/nightscout/cgm-remote-monitor/pull/3953

PieterGit:
@nielsvanvelzen thanks for finding and PR'ing this bug. Sorry it took so long to take action.

PR #3821, opened 2018-08-09
Devicestatus Delete Enhancement

jpcunningh:
Add ability to delete devicestatus entries based on query.

### DIFF
--- a/lib/admin_plugins/cleanstatusdb.js
+++ b/lib/admin_plugins/cleanstatusdb.js
@@ -117,7 +117,7 @@ cleanstatusdb.actions[1].code =  function deleteOldRecords(client, callback) {
   $status.hide().text(translate('Deleting records ...')).fadeIn('slow');
   $.ajax({
       method: 'DELETE'
-    , url: '/api/v1/devicestatus/' + dateStr
+    , url: '/api/v1/devicestatus/?find[created_at][$lte]=' + dateStr
     , headers: client.headers()
     , success: function (retVal) {
       $status.hide().text(translate('%1 records deleted',{ params: [retVal.n] })).fadeIn('slow');
--- a/lib/admin_plugins/cleanstatusdb.js
+++ b/lib/admin_plugins/cleanstatusdb.js
@@ -75,7 +75,6 @@ cleanstatusdb.actions[0].code =  function deleteRecords(client, callback) {
     }
   });
 
-
 cleanstatusdb.actions[1].init = function init(client, callback) {
   var translate = client.translate;
   var $status = $('#admin_' + cleanstatusdb.name + '_1_status');
--- a/lib/admin_plugins/cleanstatusdb.js
+++ b/lib/admin_plugins/cleanstatusdb.js
@@ -74,6 +74,7 @@ cleanstatusdb.actions[0].code =  function deleteRecords(client, callback) {
       callback();
     }
   });
+};
 
 cleanstatusdb.actions[1].init = function init(client, callback) {
   var translate = client.translate;
@@ -99,9 +100,7 @@ cleanstatusdb.actions[1].code =  function deleteOldRecords(client, callback) {
 
   if (isNaN(numDays) || (numDays <= 0)) {
     alert(translate('%1 is not a valid number', { params: [$('#admin_devicestatus_days').val()] }));
-    if (callback) {
-      callback();
-    }
+    if (callback) { callback(); }
     return;
   }
   var endDate = moment().subtract(numDays, 'days');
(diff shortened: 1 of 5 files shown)


jpcunningh:
@PieterGit, I think I'm starting to get an idea of how to edit the swagger files, but I see devicestatus doesn't have any data in the swagger files.

Is that by design or because nobody has gotten around to it? I may can fill in Lookout / Logger and OpenAPS data, but I don't know about the other devices that send data to NS.

Thanks!

PieterGit:
I think that's just not documented (yet). I think @sulkaharo can confirm.
We should check that changes to the API are documented, before the PR is merged. If you want to documented what you think you can, please do. We'll ask others to do the devices that you don't do, or put a TODO in the swagger.

jpcunningh:
@PieterGit, thoughts on merging this PR?

lib/api/devicestatus/index.js
```diff
@@ -54,6 +55,28 @@ function configure (app, wares, ctx) {
       });
     });
 
+    // delete record that match query
+    api.delete('/devicestatus/', ctx.authorization.isPermitted('api:devicestatus:delete'), function(req, res) {
+      var query = req.query;
+      if (!query.count) {
+        query.count = 10
```
PieterGit: Shouldn't we set the default to for `count=0`, so that the user must explicitly remove records?

swagger.json
```diff
@@ -640,6 +640,175 @@
...
+            "description": "The query used to find entries, support nested query syntax, for\nexample `find[created_at][$gte]=2015-08-27`.  All find parameters\nare interpreted as strings.\n",
+            "required": false,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "count",
+            "in": "query",
+            "description": "Number of records to return.",
```
PieterGit: i think this should be `Number of records to delete`? And shouldn't we make `required=true`?

lib/admin_plugins/cleanstatusdb.js
```diff
@@ -67,3 +75,58 @@ cleanstatusdb.actions[0].code =  function deleteRecords(client, callback) {
...
+
+  if (callback) { callback(); }
+};
+
+cleanstatusdb.actions[1].code =  function deleteOldRecords(client, callback) {
+  var translate = client.translate;
+  var $status = $('#admin_' + cleanstatusdb.name + '_1_status');
+  var numDays = Number($('#admin_devicestatus_days').val());
+
+  if (isNaN(numDays) || (numDays <= 0)) {
```
PieterGit: The numDays check doesn't match 1..365 in the range in the HTML input tag (zero is allowed and 366 also). 
I think a range of 1 to 365 should be checked in the the input, in the javacript and also on the API-side

lib/api/devicestatus/index.js
jpcunningh: @PieterGit, I've looked through some of the documentation for mongodb and it doesn't appear the count has any impact on remove().

I'll remove the count so it is not confusing.

swagger.json
jpcunningh: Same as above, there is no way to limit the number of records being deleted.

lib/admin_plugins/cleanstatusdb.js
jpcunningh: @PieterGit, the API side is based on the generic find capabilities in the mongodb api.  After looking at this again, I thought it best not to artificially limit the maximum number of days the user could choose to keep.

If the user has a paid version of mLab or some other hosting solution that enables to to maintain a much larger database, they may want to keep 2 years, 3 years, etc.

PieterGit:
@gregorbel reports an error with this PR "I tried PR #3821 but it is not obvious what this PR does (a screenshot would be great). I assume this add a option to delete device status but keeping Last 30 days. I tried it and it displayed "error". when I clicked delete all device status within NS it has been successfully done.". Source: https://gitter.im/nightscout/public?at=5b8e2ac3ac25fd11b5da2e9d

@jpcunningh can you have a look at that and report back?

jpcunningh:
@PieterGit, I wasn't able to duplicate the issue @gregorybel described.  My guess is that it may be related to the mongodb timeout set to 30 seconds.

The image below shows the page where the user can set the number of days to retain in the devicestatus collection, and deleting everything older than those number of days.  Clicking the Delete Old Documents executes the delete.
delete old documents

Upon success, the page appears similar to the image below, displaying the total number of deleted records.
delete old documents result

I suppose to move forward, we'll need another tester that exhibits the same problem, but doesn't need an immediate resolution so we can troubleshoot.

Does anybody have a test database that we can fill up with 100,000's of records to see if it is the timeout that may have caused the error?

PieterGit:
@jpcunningh @sulkaharo should we just merge this to dev and let users test it there? All testers have trouble with Git/Github or just delete the devicestatus in Mlab Mongodb.  I don't have the time to do testing this on a seperate mongo db. This feature is really on of the major problems people have with Nightscout, so targetting this at 0.11 release

jpcunningh:
@PieterGit, I am aware of the disappearing button. Based on the structure of the new feature matching the existing delete all device status, it appeared that was expected behavior.

I'll take a look at it again today with fresh eyes.

jpcunningh:
@PieterGit, fresh eyes helped! It was expected behavior for the existing delete button, but I found a flag I could add to keep the button in place for the new delete button.

I think it's ready to merge.

PieterGit:
@ShaggydogUK tested this PR and @jpcunningh managed to fix the problem with the disappearing button. Merging this PR to dev (finally)

PR #3839, opened 2018-08-14
Name too long, please switch back to BWP

viderehh:


PieterGit:
@viderehh This was changed by @lifeexcellence . 

Can you (and perhaps other German translators like @herzogmedia and @schmitzn  ) find consensus on whether or not this must be abbreviated or not?

viderehh:
in my case it takes to much space on my phone screen, and after a while i know whats behind bwp, iob,cob, etc...
in the english translation, the short name is also taken, so why dont do this for all!?
like to hear whats the other will say.. 
cheers

PieterGit:
I agree that sometimes a shorter name is ok. But I would like translators to agree. Please also change the base of this PR from `master` to `dev`.  We only take changes to `dev`.

herzogmedia:
yeah, well, I can see the point. If we change it back, we should add the abbreviation back in the Menu-Button in parantheses, so it's easier to see where it is coming from. Also it would be nice to have the full name of every item in the tooltip. Would that be possible @PieterGit ?

schmitzn:
I'd also prefer the short form. And I really like the idea of @herzogmedia with the tooltip.

PieterGit:
@schmitzn @herzogmedia For creating a tooltip edit the dev branch versions of https://github.com/nightscout/cgm-remote-monitor/blob/52e1ff054c98a6c0ee90d6f3b4d5c299e2b6dd40/lib/plugins/boluswizardpreview.js and https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/tests/boluswizardpreview.test.js
and please submit a new PR.

For tooltips we use jquery.tooltips (see https://www.npmjs.com/package/jquery.tooltips ) because it's the successor of tipsy. see https://github.com/nightscout/cgm-remote-monitor/search?q=tooltips&unscoped_q=tooltips for more information on the integration.

cukibeza:
Merge pull request #3839 from viderehh/patch-1 German language change. Name too long, please switch back to BWP

PR #3846, opened 2018-08-19
ensure treatments created_at in ISO format

### LINKED ISSUE #2242: Reports not including Insulin Treatments despite showing on NS Home Page
I noticed that my NS Report does not include all of my insulin treatments. The insulin shows on the NS site home page graph, but if I select Today - Show under Reports, two of the overnight insulin treatments are not included, nor are they included in the Bolus Insulin total, or on the Treatments Report page.

If I tick the IOB option on the Reports page, I do see that the two, missing insulin treatments are detected in IOB but the insulin labels are not displayed for these two, nor are they included in the total bolus insulin for the day.
missinginsulintreatmentsinreports

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39423361-reports-not-including-insulin-treatments-despite-showing-on-ns-home-page?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2559: String comparison for future treatments
It looks like future treatments uses a string comparison to work out whether a treatment is in the future or not (https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/admin_plugins/futureitems.js#L64). 

This seems to fail when the treatments entered in to mLab are 'zoned' (e.g 2017-04-28T15:05:41+01:00) as the string comparison [$gt] doesn't have any concept of timezones. This causes treatments to appear as future treatments when they really aren't (and they display properly on the main nightscout graph).

Future entries seems to be based on a unix timestamp, which doesn't suffer from this problem.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44641952-string-comparison-for-future-treatments?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F2451736&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lawley:
Because string comparisons are used to retrieve treatments data (e.g., future entries $gte) they always need to be represented with the same timezone.  This commit ensures use of UTC

this commit fixes #2559
this commit fixes #2242

### DIFF
--- a/lib/server/treatments.js
+++ b/lib/server/treatments.js
@@ -144,7 +144,8 @@ function prepareData(obj) {
 
   var results = {
     //TODO: validate format of created_at
-    created_at: obj.created_at || new Date().toISOString()
+    // Perform timezone conversion in case it is required
+    created_at: (obj.created_at ? new Date(obj.created_at) : new Date()).toISOString()
     , preBolusCarbs: ''
   };
 


lawley:
Or one could convert queries to UTC as well?  

Best would have been to use a proper date type in the db rather than a String.

I'm not mongo-literate or I'd have looked at such a refactoring.

Michael

PieterGit:
@lawley I think this is quite some work and needs more work and discussion.  changing the queries to UTC requires changes to all Nightscout clients. 

Postponing to 0.12.x. Please target changes at dev and not at master. If you think this will be implementable in 0.11 timeframe.

sulkaharo:
@lawley the latest release now fixes this, thanks for getting the ball rolling!

PR #3863, opened 2018-08-25
redo brower-update pr. Warn user about outdated browser.

PieterGit:
This PR includes the browser-update npm packages and integrates it into Nightscout. This will warn users with older browsers that they should update.
I redid the PR from https://github.com/nightscout/cgm-remote-monitor/pull/3806 without the webpack downgrade from 4.x to 3.x. The reason that stock 0.10.3 was not working with older browsers was because of ES6 features used by xdrip plugin and was not caused by webpack.

Because older browsers are unsafe and the user should be protect the data in Nightscout, we should warn users about the risks of unsafe browsers.


PieterGit:
Tested with Iphone 4s with Chrome and Safari and it works for me.

@unsoluble @RedCyclist1 @sulkaharo @MilosKozak @jpcunningh 
Can you please review and perhaps test this PR?

unsoluble:
I can't tell from skimming the code here how exactly this warning manifests — how does it appear to the end user, and when does it trigger?

PieterGit:
@unsoluble It uses the scripts from the https://browser-update.org/ initiative.  I used the following settings:

image

I added `no_permanent_hide:true` so that the users will be warned after 150  hours (6.25 days) instead of having an option to permanent hide the message.

unsoluble:
Hmm. Since the one legitimate use case for older browsers is in always-on display fixtures, I’d argue that the permanent dismissal should be available — it’d be untenable to have to go around the house dismissing a warning on all your displays every 6 days.

PieterGit:
@unsoluble Using old and insecure browsers isn't a legitimate use case. It's a use case of saving costs and accepting the risk by the user of misuse of personal data with possible severe consequences (e.g. if Nightscout is also used with an APS).  I think a weekly reminder that the user is having an insecure setup is not too much. The popup is quite unobtrusive and much more user friendly than blocking these users.

Under EU GDPR laws, a software product that processes personal information should implement "Data protection by design and by default" (GDPR article 25), see https://gdpr-info.eu/art-25-gdpr/

Unsafe browsers violate approx 4 of the top 10 most common security practices ("current state of the art" in GDPR-terms), see https://www.owasp.org/index.php/Top_10-2017_Top_10
- A3:2017-Sensitive Data Exposure
- A6:2017-Security Misconfiguration
- A7:2017-Cross-Site Scripting (XSS)
- A9:2017-Using Components with Known Vulnerabilities

The A9 sums up the risks quite good:  If a vulnerable component is exploited, such an attack can facilitate serious data loss or server takeover. Applications and APIs using components with known vulnerabilities may undermine application defenses and enable various attacks and impacts.

An alternative would be to introduce environments variables like `PROTECT_PERSONAL_INFORMATION=false` and/or `UNSAFE_ALLOW_PERMANENT_HIDE_UNSAFE_BROWSERS=true` and then allow permanent hiding of security warnings.  That way the users explicitly asks the software to be unsafe and the user dismisses the right that the software will protect personal information. I would not like configuration settings that allow users to do unsafe stuff.

I would say that a weekly reminder is the right balance of allowing users to use cheap setups, and informing them about the risks they are taking and implementing security and privacy by default in the Nightscout software.

I would like to have the opinion of @sulkaharo and @MilosKozak and @jasoncalabrese on this PR and on whether or not to allow users to dismiss security warnings if they choose to use an unsecure/unsafe setup.

unsoluble:
Thanks for the comprehensive notes there. I take your points on the security concerns. Still think there's something to be said for the hundreds of folks out there who can't afford to buy new devices (a 5th-gen iPod Touch, the oldest that runs iOS 9, is $170 US), but am interested to hear what the others think.

PieterGit:
@sulkaharo, @MilosKozak, @jasoncalabrese  or @jpcunningh : can you please give your opinion on this PR. Please share your thoughts on showing this warning to users with unsupported browsers and whether or not we should give those users the possibility to skip the old browser warning forever or for a week only. If I don't see input I will merge it to dev, to include it in 0.11 next week. (I'm open to discussion, but don't want PR's to be open for too long).

jpcunningh:
@PieterGit, I'm on board with this PR and good with a weekly notification since it's easy to dismiss.

Would a custom notification be worthwhile such as:
```
text:'Your browser, {brow_name}, is too old. It may not have critical security patches, and Nightscout may not operate correctly: &lt;a{up_but}&gt;update&lt;/a&gt; or &lt;a{ignore_but}&gt;ignore&lt;/a&gt;.'
```

PieterGit:
@jpcunningh I decided not to change the default message, because otherwise we should add translations for all languages we support in `lang_xx` in `bundle/bundle.source.js`. In my case the default Dutch language is quite explanatory and I don't want to add all kind of language translations in `bundle/bundle.source.js`. I added some more information on older browsers in the readme to clarify.

@unsoluble If people want to have no warning they should set `no_permanent_hide:false` in `bundle/bundle.source.js`. I'm against not implementing security by default and privacy by default in Nightscout and weakening the product not to do the right thing of doing a weekly notification.

PieterGit:
Postponing to Nightscout 0.12.x.

PieterGit:
Postponed to 0.13.x release. If it doesn't make that, we will close the PR.

sulkaharo:
Closing as this needs refactoring to merge cleanly and with the transpiling, there's much less need for the feature

PR #3888, opened 2018-09-01
Add browser requirements to Readme

unsoluble:
(plus fix a few broken links)

Since NS won't load in older browsers, should lay this out up front in the docs.


PieterGit:
@unsoluble sorry for not merging sooner. Thanks for the doc update.

PR #3899, opened 2018-09-04
DayToDay report cleanup

bassettb:
Only show decimal places if they are nonzero. Clean up rotation code. Don't show insulin units for carb entries.
Instead of translating and then rotating around a point with offset 10, 0,  rotate and then move to a location 10*sqrt(2)/2, 10*sqrt(2)/2 offset.  Only have to specify the offset once, same end result.


lib/report_plugins/daytoday.js
```diff
@@ -15,6 +15,7 @@ function init() {
   return daytoday;
 }
 
+var showInsulinUnitsWithCarbTreatment = false;
```
PieterGit: how can `showInsulinUnitsWithCarbTreatment` get `true`? Adding a flag with default false sounds like adding dead code.

PieterGit:
@bassettb can you have a look at my comment about: `showInsulinUnitsWithCarbTreatment `. Setting this to the 0.11 milestone.

lib/report_plugins/daytoday.js
bassettb: Agreed.  My problem is that I see no value to that extra text.  But it's in the code, so maybe someone uses it.  I figured if people wanted it, they could fork and set the boolean to true.  Do you think anyone uses it?

PieterGit:
@bassettb Can you remove the code that can't be executed  `showInsulinUnitsWithCarbTreatment = false`? If so we can merge this for the 0.11 release. Otherwise this PR will be postponed.

PieterGit:
@bassettb the extra lines are still in the PR. something went wrong with your fixes. Please remove dead code with `showInsulinUnitsWithCarbTreatment ` in the PR

bassettb:
Sorry, I didn't see your comments until now.  I removed the dead code.

bassettb:
Will create smaller PRs.

PR #3901, opened 2018-09-05
Pivotal Web Services Cloud Support

thadc23:
This PR adds a manifest file, environment variables, and instructions for deploying Nightscout on the Pivotal Web Services cloud platform.


PieterGit:
Another note: please target your PR at the `dev` branch.  Can somebody test this?

package.json
```diff
@@ -97,18 +98,17 @@
     "style-loader": "^0.20.2",
     "traverse": "^0.6.6",
     "uglify-js": "^3.4.6",
-    "uuid": "^3.2.1"
-  },
-  "devDependencies": {
```
PieterGit: Something went wrong here. you removed the devDependencies

manifest.yml
```diff
@@ -0,0 +1,24 @@
+---
+## PWS information section
```
PieterGit: doest this file needs to be named `manifest.yml` or is that configurable? Please give more elaboration on this file an it's usage. People should not change (but copy) files that are under source control. So please rename this file with something with example or move to docs folder.

env.js
```diff
@@ -90,7 +91,13 @@ function setVersion() {
 }
 
 function setStorage() {
-  env.storageURI = readENV('STORAGE_URI') || readENV('MONGO_CONNECTION') || readENV('MONGO') || readENV('MONGOLAB_URI') || readENV('MONGODB_URI');
+  var appEnv = cfenv.getAppEnv({vcap: {services: {'mlab': [{name: 'thad-mlab', credentials: {uri: 'mongodb://CloudFoundry_lqdvkfrq_lsi1kb2n_3b3qo8fq:bDiPYaa3ObIjI7ymkF5h1kExt9Ca3fN2@ds137102.mlab.com:37102/CloudFoundry_lqdvkfrq_lsi1kb2n'}}]}}});
```
PieterGit: please fix and make configurable (and change your mlab credentials)

package.json
thadc23: so I pushed them up into the `dependencies` for the sake of the buildpack on PWS.  I have an open issue with them about why they can't see devDeps when packaging.  Moving them makes it so the person deploying this app only has to run a single command with a single CLI to get Nightscout deployed.  Leaving webpack in the devDeps section would require them to run an `npm install` before deploying.  I was attempting to make it as easy as possible on the user.

manifest.yml
thadc23: It can be named anything as the filename can be passed to the deploy command of the cloud foundry CLI.  Putting it at the root level with the default name makes the resulting command easier on the user. I figured that since everyone who uses this has their own fork, they could just edit that file in their own fork.

package.json
PieterGit: What we need to do is publish the npm package with the generated webpack javscript included and use that for these kind of stuff. Current version requires `npm install` which isn't very strange. Isn't there a way to let the cf install execute the `npm install`?

manifest.yml
forddonald: Correct. A `cf push` from within the root directory to PWS will find and use a `manifest.yml` file if it is named as such. If named differently, you have to explicitly point at it. So having a default `manifest.yml` file at root is an accelerator to the PWS user to spare them from having to load in environment variables (and tune the app with mem and disk settings) by hand later.

package.json
PieterGit: Is the symlink problem a pivotal problem? We can't change our dependencies for a bug in the deployment system of a hosting provider. We need to find/implement a better solution.

thadc23: I think it is a buildpack issue, but I haven't gotten a response from that issue yet either.  There is a workaround that would require an `npm install` be run before the push.  I will change the dev dependencies back and add that section to the new PWS README.

PieterGit: @thadc23 please let me know if you can find a solution. I'm planning on publishing a cgm-remote-monitor npm for 0.12 and then we don't need the devDependencies because they will be used with the npm packaging.

PieterGit:
@thadc23 Can you give an update on this PR? There are some conflicting files, especially the README.md and swagger files. Can you resolve these?

thadc23:
@PieterGit I will work on these conflicts, sorry missed the request.

PieterGit:
@thadc23 Please let me know if it's finished. I hope @forddonald can then test your code changes.

env.js
```diff
@@ -90,7 +91,13 @@ function setVersion() {
...
 
 function setStorage() {
-  env.storageURI = readENV('STORAGE_URI') || readENV('MONGO_CONNECTION') || readENV('MONGO') || readENV('MONGOLAB_URI') || readENV('MONGODB_URI');
+  var appEnv = cfenv.getAppEnv();
+  var svcs = appEnv.services;
+  var uri = null;
+  if (svcs.hasOwnProperty('mlab')) {
+    uri = svcs.mlab[0].credentials.uri;
+  }
+  env.storageURI = readENV('STORAGE_URI') || readENV('MONGO_CONNECTION') || readENV('MONGO') || readENV('MONGOLAB_URI') || readENV('MONGODB_URI') || uri;
```
PieterGit: Please refactor so that people who don't have `cf` or `svcs.hasOwnProperty('mlab')` are not affected by this feature. 
You're creating an `uri`, which is only used for a very small percentage of the users. Please use `if` statement to only use that when people use Pivotal Web Services Cloud Support.

PieterGit:
@thadc23 I reviewed your code and found one small issue I would like you to fix. If the fix is done and tested and confirmed working by  @forddonald  I'll merge it to dev.

manifest.yml
PieterGit: @forddonald @thadc23 I think that would be better. I don't like people to put credentials in files, because history has proven that they will end up in SCM systems and will be made publicly available. Given the fact that Nightscout saves personal information we should try to prevent that.

PieterGit:
@thadc23 Can you confirm that this PR is ready? Can you resolve the `npm-shrinkwrap.json` conflict?
@forddonald Can you confirm you tested the latest version of the PR?
If so, I can merge it to `dev` so that it can be part of the 0.11 release.

PieterGit:
@thadc23 Can you give an update. Did you test this PR, is this ready for testing?
@forddonald Can you confirm that this PR works for you?
This PR needs to be tested before we can include it in the 0.11 release.

PieterGit:
@thadc23 @forddonald no response for 25 days, so postponing to the 0.12 release.  Can you please give an update on the state of this PR. At least this must be tested and confirmed by another user before we can merge this PR.

forddonald:
Sorry, thought I had confirmed this. Confirmed working on PWS with the additions above.

unsoluble:
Kinda feel like the `nightscout-factory` process is a black box into which users are feeding their PWS credentials. Should be an explicit privacy policy somewhere maybe? Or is the source for that app also open and available for scrutiny?

PieterGit:
@forddonald @thadc23 Are you still interested in finishing this PR? Please have a look at mine and @unsoluble comments above and please resolve conflicts and confirm you fixed the review remarks and the stuff is tested and ready to merge.

thadc23:
@unsoluble Since we have had trouble getting it open-sourced, we will just remove all references to `nightscout-factory` and go back to the manual instructions to not hold this up any longer.  We will go back to the drawing board on how we could help ease the deployment and upgrades on Cloud Foundry.

@PieterGit  We are interested in getting the PWS support merged in with the manual steps for deployment.  We will work to finish this up and re-test with all the latest changes.

sulkaharo:
Closing now given original contributors are not active. I can still merge this if this is made compatible with most recent dev - please reopen the PR if you're willing to go through the effort. Note the NS documentation now lives in https://nightscout.github.io and it'd be great to have install docs PR if this goes live.

PR #3945, opened 2018-09-22
0.10.3 try number 2

TecMunky:


TecMunky:
Ok

TecMunky:
wrong direction -delete

PR #3953, opened 2018-09-29
upgrade node to 8.12.0, upgrade npm packages (express, supertest, webpack, swagger-ui-dist) and fix memory problem

PieterGit:
also reimplement #3819 (webpack) without reformatting
Todo: test on Azure and confirm that nodejs 8.12.0 is available there.


bassettb:
I tested this out on my server running on a t3.micro with 1GB RAM.  Unfortunately, it does not appear to fix the out of memory issue when running webpack.  With swapfile off, it locks up the system like it did before.  With swapfile on it uses all of the RAM and some swap.

PieterGit:
@bassettb thanks for testing. I think I missed some of the changes in https://github.com/nightscout/cgm-remote-monitor/pull/3819 because of the reformattign.
Could you retest that PR and report back?

PieterGit:
@bassettb I found I missed some of @nielsvanvelzen fixes. Now reapplied them with the Nightscout default formatting ident of 2. Can you please retest and report back if it works (or not). I'm running this branch on Heroku and can't find any problems.

PieterGit:
Merged to dev. Don't know if this works under Azure which currently still has Node 8.11.1.

bassettb:
I tested it just now and it's still using over 750MB of RAM when running webpack.

PR #3954, opened 2018-09-29
Fix CGM voltageb battery warning level to match xDrip+

jpcunningh:


lib/plugins/xdrip-js.js
```diff
@@ -172,7 +172,7 @@ function init(ctx) {
         result.level = levels.WARN;
       }
 
-      if (sensorInfo.xdripjs.voltageb && (sensorInfo.xdripjs.voltageb < prefs.warnBatV)) {
+      if (sensorInfo.xdripjs.voltageb && (sensorInfo.xdripjs.voltageb < (prefs.warnBatV - 10))) {
```
PieterGit: shouldn't the prefs.warnBatV not be lowered by 10 then?

jpcunningh: voltagea test uses the prefs.warnBatV value as is (not shown in the snippet above), but voltageb is offset to 10 less than the voltagea test value.

PieterGit:
No response. Looks good. If somebody can confirm this PR of @jpcunningh is ok, please do so in this issue.

Issue #3603, opened 2018-05-26
Treatment edit/delete failures


mpangburn:
Both the reports GUI and the API are very unreliable with respect to treatment edits and deletions.
When going to Reports -> Treatments and using the green pencil to edit or the red x to delete, there is no indication of success, and confirming an edit or delete appears to have no effect.

Similarly, when invoking the API with PUT/DELETE requests, the outcomes are unpredictable. Sometimes an update succeeds normally, but more often it produces an additional treatment with the same ID without deleting the prior. Deletions appear to fail consistently. These calls to the API return successful HTTP responses but do not achieve the desired effect.

Nightscout v0.10.2

PieterGit:
Do the problems occur with all major browsers (Chrome, Firefox, Safari). Do you see errosr in the Javascript console somehow.

razor459:
v0.10.3 in Safari on MacOS
It appears the issue is that it does not display the updated data. Clearing the cache and refreshing the page shows that the edit succeeded.

Problem also exist when editing Treatments via Report in Nightscout and NIghtscout X iOS apps

PieterGit:
@razor459 @mpangburn This has been fixed with https://github.com/nightscout/cgm-remote-monitor/pull/3957 . Please confirm that this works for you. Closing this issue. Feel free to reopen it if the problem persist. You need to install latest `dev` branch sinds 0.11 has not been released yet.

PR #3957, opened 2018-09-29
Fixed edit/delete treatment bug (in reports)

komarserjio:
Fix for GUI issue described in #3603:
"When going to Reports -> Treatments and using the green pencil to edit or the red x to delete, there is no indication of success, and confirming an edit or delete appears to have no effect."

Fix preventing aggressive caching of fetched treatment data and make sure a browser makes a new request after editing/deleting instead of using cached data.

Fun fact. The bug is not reproducible if Chrome Dev Tools are opened. The browser's cache policy differs in this case.


PieterGit:
@komarserjio Nice catch. Thanks for fixing.

PieterGit:
Tested with dev and works as expected. @komarserjio thanks for the fix.

PR #3970, opened 2018-10-04
Create issue templates for NS repo

danamlewis:
This creates issue templates for the NS repo. To see an example, the oref0 (OpenAPS code) repository uses these as well: https://github.com/openaps/oref0/issues/new/choose

Example from oref0 of how the issue template page looks like


PieterGit:
@danamlewis Thanks. Applied to master and dev for immediate use.

PR #3973, opened 2018-10-05
Using labels with for attributes on report page

D9ping:
UI change on report page.
This will e.g. make clicking on the text close to a checkbox select the proper checkbox etc. by using the label tag with the for attribute.


PieterGit:
@D9ping Looks good to me, but I noticed you only did a couple of these and I would like to have a consistent UI within Nightscout. Can you use also check if you can do the whole codebase, e.g. use https://github.com/nightscout/cgm-remote-monitor/search?q=%3Cinput+type%3D%22checkbox%22&unscoped_q=%3Cinput+type%3D%22checkbox%22

D9ping:
I will start over pushing from my own branch instead of from dev into dev directly.

PR #3997, opened 2018-10-14
Update language.js  - Fix the wrong Russian

mitrei:
Fix the wrong writing of the word 'рассчет' in four lines.

The correct writing is 'расчет'.


PieterGit:
@mitrei Thanks for reapplying the fix. Applying this.

Issue #4028, opened 2018-10-28
Bug in time intervals


Kywalh:
**Describe the bug**
Trying to update the basal profile today, I noticed that I wasn't able to add the interval "23:00" since the droplist was limited to 22:30.
I thought I was limited in number of intervals so I tried to delete some and create 23:00 again but unsuccessfully... the drop list still shows 22:30 max time...

**To Reproduce**
simply try to add "23:00"

**Expected behavior**
I was expecting, as before, to add "23:00"

**Screenshots**
first initial attempt:
![44927066_10215292952189801_1790805340124610560_n 1](https://user-images.githubusercontent.com/28366579/47614625-1dd7b380-daa3-11e8-97aa-cabef22bffc7.jpg)

Second attempt after deleting some intervals:

![44927066_10215292952189801_1790805340124610560_n](https://user-images.githubusercontent.com/28366579/47614626-1dd7b380-daa3-11e8-82df-303e0fd913f6.jpg)

**Your setup information**
- What version of Nightscout: 0.10.3
- What type of CGM, and how do you get your data there? Not applicable
- Is your issue specific to a browser : Chrome 70.0.3538.77 and Android 8.0 - Chrome 70

**Additional context**
Could it be related to wintertime change in Europe ?

DigitalDan1:
@Kywalh I noticed the same issue today, and it's been reported before (https://github.com/nightscout/cgm-remote-monitor/issues/3469).  It's due to the way in which the 30 minute intervals are formatted and added to the select lists; it counts on from midnight on the current day adding 30 minutes for each entry, then formats the date and time to just show the hours and minutes.  Today in the UK, at 1 am, daylight saving time ended so there were two entries at 01:00 and two at 01:30 (you would have seen them in the drop down lists for earlier times), it also means the last two entries at 23:00 and 23:30 are missing.  My pull request should fix it.

Kywalh:
@ DigitalDan1 many thanks !!!!!

PieterGit:
@Kywalh Can you confirm that this bug has been fixed by https://github.com/nightscout/cgm-remote-monitor/pull/4029 of @DigitalDan1 ? Or does it still exist? Please give an update, otherwise, we will postpone this to Nightscout 0.12.0

Kywalh:
Hello @PieterGit,
How can I test it ? 
Can I simply change the date to a DST one ?

DigitalDan1:
[...]

Issue #3469, opened 2018-03-25
Profile editor times incorrect after daylight savings time change


Pogman:
version 0.10.2-release-20171201

Clocks went forward +1 hour at 1am on 25/3/2018

Opened profile editor at 2pm.

The profile shows blank times and the drop down time selection jumps between the +1 hour time change.

Expected behaviour is not to use TZ or daylight savings for time selections within profiles.

![ns-0 10 2-bug-profile-1](https://user-images.githubusercontent.com/7413594/37875706-051d5808-303b-11e8-971a-622a360aa52c.JPG)

Dropdown time menu:

![ns-0 10 2-bug-profile-2](https://user-images.githubusercontent.com/7413594/37875702-f4e2aeb6-303a-11e8-9652-20cfb8885445.JPG)

triplem:
I do have the same issue. The total per day is not calculated as well. The whole profile cannot be edited anymore. No additional profile can be added nor deleted.

tanja3981:
Stacktrace from dev console:
TypeError: time is null[Weitere Informationen]
profileeditor.js:628:9
toMinutesFromMidnight
http://tanjascgms.herokuapp.com/profile/js/profileeditor.js:628:9
timeDiffMinutes
http://tanjascgms.herokuapp.com/profile/js/profileeditor.js:205:20
refreshTotalBasal
http://tanjascgms.herokuapp.com/profile/js/profileeditor.js:219:16
initProfile
http://tanjascgms.herokuapp.com/profile/js/profileeditor.js:242:5
initRecord
http://tanjascgms.herokuapp.com/profile/js/profileeditor.js:199:5
initeditor
http://tanjascgms.herokuapp.com/profile/js/profileeditor.js:182:5
l
jquery.js:3187
fireWith
jquery.js:3317
i
jquery.js:8757
t/<
jquery.js:9123

MilosKozak:
yes, 2 days every year it doesn't work ;)

PieterGit:
@MilosKozak @Pogman @triplem @tanja3981 This bug is now fixed with https://github.com/nightscout/cgm-remote-monitor/pull/4029

Closing issue.

PR #4029, opened 2018-10-28
Fix daylight saving bug in profile editor

DigitalDan1:
In the profile editor, if the clocks have gone back/forward on the current day, i.e. daylight saving changed today, then the drop down lists for the From time show duplicate/missing values.  This can cause profiles to be saved with invalid ranges.

This change ensures the values in the select/drop-down lists range from 00:00 to 11:30 pm (or 23:00) by using the UTC time as the starting point for the offset and time value and display formatting.

Fixes https://github.com/nightscout/cgm-remote-monitor/issues/4028
and https://github.com/nightscout/cgm-remote-monitor/issues/3469


PieterGit:
@DigitalDan1 Tested in `Europe/Amsterdam` and Wintertime has begon yesterday. Fix works as expected. Thanks for fixing this. Applied to dev.

Issue #4056, opened 2018-11-08
Profile edit not working


vickster1:
**Describe the bug**
I open my profile edit page. It takes 30s to load. Sometimes the drop down option to choose a different profile to edit is shown, other times it is missing (pic where it is missing attached).

Scroll down the page, and it is a blank, black screen, which loads after another 30-60secs.
I choose a feature to edit, and click on it. I get no response. I wait, and usually the cursor shows after a while. 
However, editing remains very difficult - either the page doesn’t respond, or I make a change and it immediately reverts to the original page shown when I opened it.
If I keep trying I can generally get to make a few changes over 15 minutes- at most 3 or 4 in this time, and sometimes none.

**Your setup information**
- What version of Nightscout: dev 0.11.0, but I’ve had this with 0.10.3 & older versions. I saw it in 2016 & 2017, but then it was occurring on my Samsung phone & worked mostly OK on desktop & Apple devices.
- What type of CGM, and how do you get your data there? Dexcom G5 & xdrip+ on Sony mobile phone
- Is your issue specific to a browser: see above - previously it was but now affects all browsers

I have ensured my mlab database is compressed on disk & deleted device status entries.

![ffbb72cd-ced8-4582-9632-42efb022b064](https://user-images.githubusercontent.com/38795035/48199687-cebc3900-e354-11e8-970c-f5ede2686867.jpeg)

DigitalDan1:
I've had the same problems.  I'm pretty sure my change (https://github.com/nightscout/cgm-remote-monitor/pull/4074) will fix it.

vickster1:
Also fixed in latest Dev from today 😀

PieterGit:
Fixed with https://github.com/nightscout/cgm-remote-monitor/pull/4074

PR #4074, opened 2018-11-16
Workaround for re-loading bug in profile editor: don't show profile form during save

DigitalDan1:
This fixes a problem I noticed in the profile editor where it re-loads and re-draws every time there's an update from the server (due to socket.io emit calling the callback).

This was causing issues where the user could be making changes to their profile, then the data would re-load and re-draw, so they'd lose their changes and have to start again.

It was also causing the save button's event to be hooked up multiple times, which meant saving was taking longer and longer, and made it more likely that any further changes made by the user were lost.  Users would have seen the user interface update whilst editing, replacing their values with a previous version (from whichever save came back last).  This change (crudely) hides the form whilst the save is happening to make sure the user doesn't carry on making changes that will be lost when the re-draw happens.

This _should_ fix https://github.com/nightscout/cgm-remote-monitor/issues/4056.


sulkaharo:
So hey - I've never seen this behavior, so I'm wondering what the root cause for the reload is? Or at least - in our instance, I can edit the profiles and the page doesn't reload and I've never lost edits. Which browser are you testing on?

DigitalDan1:
Hey @sulkaharo, I've seen this on Windows 10 Chrome and Edge, and iPhone XS Max safari.  When it happens there are regular "Client connected to server" logs in the console.  I think it's more likely to happen when there's a lot of data being transferred, maybe a timeout is causing socket IO to reconnect?

Basically the 'connect' event is fired on the initial connection _and_ any subsequent re-connection, so there's always a chance the callback could be called more than once.

It's easy enough to reproduce: just switch your network connection off then on again.

PieterGit:
@DigitalDan1 thanks for this PR.
I was able to reproduce the fact that `socket.io` updates can revert changes to the profile. 
I would classify hiding screen elements as a workaround and we should fix the root cause of this problem with Nightscout 0.12. I created https://github.com/nightscout/cgm-remote-monitor/issues/4252 for that.

Since this issue can cause problems for users I applied this, so that we can include it in Nightscout 0.11.

DigitalDan1:
@PieterGit, sounds good, thanks.

Just want to capture a few comments here for anyone (including me) that may come back to work on a more elegant fix.

I agree the route cause should be fixed, but also think every page should gracefully handle a socket IO reconnect (I’m pretty sure the admin page has the same issue).

Hiding screen elements addresses a different problem; if the save takes a long time the user can carry on making changes that are lost when the save actually finishes. I’d usually show a dialog (like a Bootstrap Modal) whilst waiting for the save to return. Hiding the form is crude, but does the job for now.

Issue #4053, opened 2018-11-08
day to day NS report is not showing all data


thomasmenges:
**Describe the bug**
The day to day NS report has some problems. 
After using Glimp instead of XDRIP (I had to change because of connecting problems to my CGM system libre (miaomiao) the day to day report shows no BG data any more. But these values are shown correctly in the NS overview page. 

So, the values must be stored in mongo db. 

If you are using the new function of NS reporting (http://www.nightscout-reporter.zreptil.de/) all values are shown correctly. That means that the values are stored in MongoDB (NS) correctly but the day to day report have some problems showing the values.

What can be done to solve these problems ?

**Screenshots**
![image](https://user-images.githubusercontent.com/37514390/48305601-092cfe00-e52e-11e8-8c74-9b927c88f75b.png)

![image](https://user-images.githubusercontent.com/37514390/48305635-b9026b80-e52e-11e8-9512-ab756802cbd0.png)

**Your setup information**
I'm using NS 0.11.0-dev  on   ns.10be.de
I'm using libre (miomio --> glimp) and AAPS
The problem still exists in every browser!!!!!

zehnBE:
can you look with browser "elemte untersuchen" (control + O  or Chrome control + shift + I) and look into the "network"-Tab and on "response", which the message/errors was?

sulkaharo:
This is likely already fixed in the develop branch - are you running the release or dev version? If release, can you test using develop to see if that works?

thomasmenges:
I'm using the dev branch as mentioned before.
After switching to the dev branch the problems with profile switches seem to be solved but the values from GLIMP are still missing.
So, I made some checks and I found some interesting things:

1) The missing values only occur when I'm using Glimp with connection to miaomiao. The values are delivered to NS and can be seen directliy in the NS overview (screenshot 1) but the values aren't shown in the day to day report. All other reports don't have these problems.
2) When I'm using Glimp are take the BG value with NFC instead using miaomiao the value is stored in NS and can be seen in the day to day report. Especially on Nov 12th you can see some BG values (taken with NFC ) (between 01.00 and 06.00 a.m) and the other ones later come from miaomiao --> no data is shown in the report (see the new screenshots below).

conclusion:
Yes, there seems to be some differences between Glimp (with and without miaomiao) but where must be  nevertheless a problem in the day to day report !!!!!!
[...]

PR #4087, opened 2018-11-21
Fix entries api null dereference

jpcunningh:


PieterGit:
@jpcunningh LGTM. Can you add information in what context / situation `lastEntryDate` got `null`?

sulkaharo:
Is this related to https://github.com/nightscout/cgm-remote-monitor/issues/4053 ? Looks like data from Glimp is a bit broken and we need to add validation to prevent broken data from getting into the system in the first place.

jpcunningh:
I don’t use Glimp, so I’m not sure if it is directly related. I noticed the
stack traces in my log file and put in a check to at least keep it from
crashing.

sulkaharo:
Can you pull out what data entries are causing the issue? I'd love to know where this data is coming from, so we can fix the underlying problem.

jpcunningh:
@sulkaharo, I think I found the entries that were causing the issue.

I was testing code in Lookout and accidentally uploaded a record that should have gone into the treatments collection into the entries collection instead. Treatments uses created_at instead of date and dateString for the time field.

You are right that this error could have been avoided by using data validation on insert.

PieterGit:
@jpcunningh @sulkaharo what to do with this PR. I think we should add validation, but I think we must do that for the 0.12 release. Do want to merge this in the meantime for 0.11, or just skip it. Please let me know what to do.

jpcunningh:
Let's close this PR and wait for the validation since I created my own problem by inserting invalid data. :smile:

Issue #4080, opened 2018-11-20
API bug: DELETE devicestatus and treatsments


VladislavLasmann:
Using the Nightscout-API to delete the types **devicestatus** and **treatsments** we noticed that the function remove(_id, fn) doesn't work properly. The delete seems to have no effect.

------------------------------------------
**Treatments:** 
 _[https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/server/treatments.js](url)
Line: 100_
The original code looks like this:
`function remove (_id, fn) {
    api( ).remove({ '_id': new ObjectID(_id) }, fn);
    ctx.bus.emit('data-received');
}`

Our solution:
`function remove (_id, fn) {
    api( ).remove({ '_id': _id }, fn);
    ctx.bus.emit('data-received');
}`

**Devicestatus:**
 _[https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/server/devicestatus.js](url)
Line: 62_
The original code looks like this:
`function remove (_id, fn) {
    var filter;
    if (_id === '*') {
      filter = {};
    } else {
      filter = { '_id': new ObjectID(_id) };
    }
    return api( ).remove(filter, fn);
  }`

Our solution:
`function remove (_id, fn) {
    return api( ).remove('_id': _id , fn);
  }`

------------------------------------------
Using this fix, the delete function works.

jpcunningh:
@VladislavLasmann, I believe that would break the ability to delete all of the devicestatus records as used in the admintools.

jpcunningh:
#4088 should correct this issue.

I'm not sure why it doesn't work as it was written. I don't believe the PR should cause a different result, but it works and preserves the benefits of using ObjectID and the '*' option for devicestatus.

PR #4088, opened 2018-11-21
Fix devicestatus and treatments id delete

jpcunningh:
Fix for https://github.com/nightscout/cgm-remote-monitor/issues/4080

### DIFF
--- a/lib/server/activity.js
+++ b/lib/server/activity.js
@@ -61,7 +61,8 @@ function storage (env, ctx) {
   }
   
   function remove (_id, fn) {
-    return api( ).remove({ '_id': new ObjectID(_id) }, fn);
+    var objId = new ObjectID(_id);
+    return api( ).remove({ '_id': objId }, fn);
   }
 
   function api ( ) {
--- a/lib/server/devicestatus.js
+++ b/lib/server/devicestatus.js
@@ -73,7 +73,8 @@ function storage (collection, ctx) {
     if (_id === '*') {
       filter = {};
     } else {
-      filter = { '_id': new ObjectID(_id) };
+      var objId = new ObjectID(_id);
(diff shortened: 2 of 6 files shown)


PieterGit:
@jpcunningh Do you understand why you need to create a new variable and refer to that in the JSON 
part of the code.
image

From reading the code I would assume that the new code is equivalent (but introduces a new temporary variable). Am I missing something?

candreassen:
Omg stop sending these threads. I hetbloke 50 emails a day.  Stop stop stop please

sulkaharo:
@candreassen if you get emails from changes in Nigthscout, it means your account is actively "watching" the Nightscout repository and thus will get an email from every change. If you want the email to stop, go to the Github Nightscout page (for example, click the "view in Github" link in the email with this message) and find a button on top of the page that says "Unwatch" and either choose "Not Watching" or "Ignore" from the menu.

sulkaharo:
@PieterGit looks like calling the remove() method call with `new ObjectID()` results in the remove() call being called with a function pointer that returns the ObjectID, not the new ObjectID(), which causes the call to fail. Using the temporary variable fixes this as the remove() function is then called with ObjectID, as expected.

jpcunningh:
Thank you for the explanation, @sulkaharo! I didn't know why one way worked vs. the other.

PieterGit:
@sulkaharo thanks for the explanation. I hate merging stuff I don't understand.
@jpcunningh thanks for the fix

PieterGit:
The user documentation for this feature can be improved. I logged issue https://github.com/nightscout/documentation/issues/2 for that.

staceyc2578:
How do I stop getting email notifications

Issue #4046, opened 2018-11-04
Remove sgvdata and mqtt or upgrade and maintain it


PieterGit:
Follow up on: https://gitter.im/nightscout/public?at=5bde0b3a5905a919554c36e3

Current sgvdata module has the following issues:
- it has a small security issue with protobufjs that needs to be upgraded
- it's not a npm package, but uses git://github.com/ktind/sgvdata.git#wip/protobuf
- it's not maintained
- the code is hard to grasp

I tried upgrading it and integrated it back into Nightscout. If integrated with Nightscout it fails on a mqtt test of Nightscout that uses the possibility to use dexcom hexcoded string into json. See code at 
https://github.com/ktind/sgvdata/pull/1

```
  it('should parse protobuffers', function () {
    // based on Nightscout tests/mqtt.test.js
    var pb = require('../').sync.protobuf;
    var payload = new Buffer('0a1108b70110d6d1fa6318f08df963200428011a1d323031352d30382d32335432323a35333a35352e3634392d30373a303020d7d1fa6328004a1508e0920b10c0850b18b20120d5d1fa6328ef8df963620a534d34313837393135306a053638393250', 'hex');
    var result = pb.parse(payload);
    result.direction.should.equal('NONE');
  });
```

it fails with:

```
$ make
NODE_ENV=test node_modules/.bin/mocha --require blanket  -R 'travis-cov' tests/*test*.js
Tests failed.

Error: Illegal wire type for field Message.Field .Practical8601.direction: 2 (0 expected)Error: Illegal wire type for field Message.Field .Practical8601.direction: 2 (0 expected)
    at Field.ProtoBuf.Reflect.FieldPrototype.decode (node_modules\protobufjs\dist\protobuf.js:3704:23)
    at Message.ProtoBuf.Reflect.MessagePrototype.decode (node_modules\protobufjs\dist\protobuf.js:3263:45)
    at Function.Message.decode (node_modules\protobufjs\dist\protobuf.js:2952:37)
    at Function.parse (lib\protobuf.js:49:19)
    at Context.<anonymous> (tests\test_protobuf.js:46:21)
make: *** [Makefile:6: travis-cov] Fout 1
```

[...]

PR #4091, opened 2018-11-21
remove mqtt and svgdata, update packages and fontello and helmet improvements

PieterGit:
Fixes https://github.com/nightscout/cgm-remote-monitor/issues/4046
Also ran `npm update` and hand updated several packages to newer major or newer minor versions.
Results: 548 additions and 2,262 deletions.

I really like the: 

```
audited 5393 packages in 53.167s
found 0 vulnerabilities
```
output. Also https://observatory.mozilla.org/ is much more happy with the security headers added by helmet package. I could not find any problems with oref0 rigs or API-calls yet.


PieterGit:
@sulkaharo or @jasoncalabrese can you review and re implement the CSV/TSB stuff without dependency on svgdata? For now I return 403 errors.

app.js
```diff
@@ -14,22 +15,32 @@ function create(env, ctx) {
     var appInfo = env.name + ' ' + env.version;
     app.set('title', appInfo);
     app.enable('trust proxy'); // Allows req.secure test on heroku https connections.
-    if (process.env.INSECURE_USE_HTTP !== 'true') {
+    if (env.settings.isEnabled('insecureUseHttp')) {
```
PieterGit: can anybody explain why this does not seem to work?

PieterGit: fixed thanks to Antony @acron0

PieterGit:
https://observatory.mozilla.org results improved from F to A+
image

PieterGit:
@sulkaharo @MilosKozak @jasoncalabrese @jpcunningh can one of you review this PR?

I hope one of you can re implement the CSV/TSB stuff without dependency on `svgdata`.

Implementing Content Security Policy will be quite some work and I think I postpone that to 0.12.
I see quite some CSP errors on the console.

I hope one of you can explain why I can't seem to be able to use 
`if (env.settings.isEnabled('secureCsp')) {` and must use `if (process.env.SECURE_CSP == 'true') {`
I can't understand why the `isEnabled` does not seem to work for me.

app.js
```diff
@@ -14,22 +14,38 @@ function create(env, ctx) {
...
         app.use((req, res, next) => {
         if (req.header('x-forwarded-proto') !== 'https')
-            res.redirect(`https://${req.header('host')}${req.url}`)
+            res.redirect(`https://${req.header('host')}${req.url}`);
         else
             next()
         })
-        if (process.env.SECURE_HTTP_HEADERS == 'true') {
-            const helmet = require('helmet')
+        //if (env.settings.isEnabled('secureHstsHeader')) { // Add HSTS (HTTP Strict Transport Security) header by
```
PieterGit: This syntax does not seem to work for some reason.

app.js
```diff
@@ -14,22 +14,38 @@ function create(env, ctx) {
...
             app.use(helmet({
                 hsts: {
                     maxAge: 31536000,
-                    includeSubDomains: true,
-                    preload: true
+                    includeSubDomains: includeSubDomainsValue,
+                    preload: preloadValue
                 }
             }))
+            //if (env.settings.isEnabled('secureCsp')) { // Add Content-Security-Policy directive by default
```
PieterGit: Same here... This syntax does not seem to work for some reason.

app.js
```diff
@@ -14,22 +14,38 @@ function create(env, ctx) {
     var appInfo = env.name + ' ' + env.version;
     app.set('title', appInfo);
     app.enable('trust proxy'); // Allows req.secure test on heroku https connections.
-    if (process.env.INSECURE_USE_HTTP !== 'true') {
+    if (!env.settings.isEnabled('insecureUseHttp')) {
```
PieterGit: This works and redirects the site if accessed over http

app.js
jweismann: @PieterGit env.settings.isEnabled didn't work for me either so that's why I introduced the process.env construct instead. Speaking of which, I don't think CORS works as expected with the current env.settings.isEnabled construct but it's not something I use so I am not the right person to test the functionality. Are we aware of users that uses CORS today?

PR #4114, opened 2018-12-02
Adds back the support for loading Entries with CSV and TEXT formats

sulkaharo:
Adds back the ability load Entries as CSV and tab separted txt files, without introducing new module dependencies. The features were lost when the sgvdata module dependency was removed.


sulkaharo:
Crash fixed. HTML, PNG and SVG output are not supported, didn't add those

PR #4116, opened 2018-12-03
Heroku template tweaks

unsoluble:
While deploying a bunch of test apps to work on the Heroku guide rewrite, noticed that the template text could use some tidying up and clarification.

Reworded most of the descriptions for clarity and simplicity, re-ordered the vars in the code to match the alpha sort on the presented page, set defaults where values have defaults, and added a BRIDGE_SERVER line.

I didn't _remove_ anything, but I'm wondering if maybe we should — many of those vars are unnecessary for the vast majority of new users, and anyone who needs them will be tinkering with the vars later anyway, and can add them then. Thoughts? (I'm thinking in particular of DISABLE, INSECURE_USE_HTTP, the MAKER and PUSHOVER vars, SECURE_CSP, and the HSTS stuff.)


app.json
```diff
@@ -2,180 +2,150 @@
...
+      "description": "If you are bridging from the CareLink service, and are anywhere *outside* the US, change this to EU. ('US' or 'EU')",
+      "value": "US",
       "required": false
     },
-    "ALARM_TIMEAGO_URGENT_MINS": {
-      "description": "Browser default minutes since last reading to trigger an urgent alarm.",
-      "value": "30",
-      "required": false
+    "MONGO_COLLECTION": {
+      "description": "The mLab collection where CGM data is stored.",
```
PieterGit: mLab is a MongoDB saas. Nightscout should be vendor neutral for MongoDB, so use MongoDB collection instead of mlab.

PieterGit:
@unsoluble I would include all the environment variables that the code uses. So follow Python Zen: "Explicit is better than implicit.". I'm fine with merging this for 0.11

@jpcunningh or @sulkaharo can you als 👍  or 👎  this for 0.11?

unsoluble:
> @unsoluble I would include all the environment variables that the code uses.

There are *many* vars that are in the code that weren't ever in the setup template. I think as an initial onboarding tool it should remain fairly lean. If folks thing we should be presenting every var up front, that should probably come with a re-work of the template into an Essential/Advanced sort of layout if possible (but definitely not happening for this release).

package.json
```diff
@@ -76,8 +76,8 @@
     "lodash": "^4.17.11",
     "long": "^4.0.0",
     "mime": "^2.4.0",
-    "minimed-connect-to-nightscout": "^1.1.1",
-    "moment": "^2.22.2",
+    "minimed-connect-to-nightscout": "https://github.com/szymjaw/minimed-connect-to-nightscout.git#dev-servers",
```
PieterGit: we only allow released versions (due to npm security hashes).

unsoluble: I incorporated that line from @szymjaw 's changes in #4190, assuming they knew what they were doing; perhaps they could comment/fix.

If we don't get a response before you need to merge, will just revert this to the original line.

PieterGit: If the authors of that PR don't upgrade `minimed-connect-to-nightscout` in the coming days, please revert to 1.1.1.

szymjaw: I have no contact with @mddub from several monts. I do not want to assign myself to authorship, I only modified the code. 
I created npm package "minimed-connect-to-nightscout-szymjaw". I created dev-guardian-connect branch for cgm-remote-monitor and it working on heroku.

unsoluble: Where does this leave us, @PieterGit ?

PieterGit: @unsoluble please revert minimed changes in this PR. then I can merge to dev and release
We will upgrade minimed-connect-to-nightscout in 0.11.x or 0.12.x when https://github.com/mddub/minimed-connect-to-nightscout/pull/11 is finished by @szymjaw

unsoluble: @PieterGit Turns out the support for EU accounts wasn't ever present in the previous version of minimed-connect-to-nightscout; that was added in @szymjaw 's version. Have updated the docs to indicate this, but we should attempt to include this sooner than later.

PR #4129, opened 2018-12-09
Remove unused dependencies

sulkaharo:
Scanned the sources with `depcheck` after recent changes and found we had were importing `ajv, `long`, `swagger-ui-dist` and `uuid` even though those now have no usages. Tested all base functionality without the modules in place and things seem to work as expected. Additionaly removing old bundle script related to browserify


package.json
```diff
@@ -91,10 +89,8 @@
     "shiro-trie": "^0.4.8",
     "simple-statistics": "^0.7.0",
     "socket.io": "^2.1.1",
-    "swagger-ui-dist": "^3.20.1",
```
PieterGit: swagger-ui-dist is needed for /api-docs.html

PieterGit: see https://github.com/nightscout/cgm-remote-monitor/blob/dev/static/api-docs.html and https://github.com/nightscout/cgm-remote-monitor/search?q=swagger-ui-dist&unscoped_q=swagger-ui-dist

package.json
```diff
@@ -74,7 +73,6 @@
     "js-storage": "^1.0.4",
     "jsonwebtoken": "^8.4.0",
     "lodash": "^4.17.11",
-    "long": "^4.0.0",
```
PieterGit: This was required for `lib/server/mqtt.js`. Can be removed now

sulkaharo:
Doh, next up - check why the swagger dependency doesn’t report correctly.

PieterGit:
about swagger. probably best to use something like `const swaggerUiAssetPath = require("swagger-ui-dist").getAbsoluteFSPath()` around https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/env.js#L28
source https://www.npmjs.com/package/swagger-ui-dist
that way depcheck should detect it
or drop the enviroment variable and integrate it with https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/app.js#L144

PieterGit:
reimplemented with https://github.com/nightscout/cgm-remote-monitor/pull/4134

PieterGit:
superseded by https://github.com/nightscout/cgm-remote-monitor/pull/4155

PR #4161, opened 2018-12-29
Upload treatments (Bolus and Card) from Omnipod PDM IBF files directly to Nightscout.

balshor:
This WIP branch is an MVP for uploading data directly from an Omnipod PDM file to Nightscout. 
 Currently, I'm just grabbing the Bolus and Carb records, converting them into treatments, and sticking them into the DB.

I'll be running this branch on my personal Nightscout site to see how it goes.  This PR is mostly intended to show this feature to the Nightscout developers and get feedback.

Use case: I've found entering Carb/Bolus information into both Nightscout and the Omnipod PDM to be rather annoying, especially since there are multiple people who help with treatments.  It's much easier for me to simply pull the data from the PDM and upload it on demand.  Currently, I use a small program to process the file locally and dump it directly into MongoDB.  Having this functionality in Nightscout will allow other caretakers to easily upload the data.

New dependencies are on ibf-file-reader (to parse the Omnipod PDM files) and multer (to assist with file uploads).


sulkaharo:
I’d rather see the conversion happening in an uploaded before data enters Nightscout than add the parser in. If we do add the parsing capabilities, ideally this would not add any other NPM modules as dependencies than the parser itself.

balshor:
@PieterGit What configuration did you have in mind?

@sulkaharo I'm not sure what you mean.  Where should the conversion be happening?

balshor:
@jweismann Your setup sounds similar to my existing strategy of just processing the file locally (I have some code written around jomnipod) and shoving the treatments directly into the DB.  While this works, it takes some setup and training to use, so I'm the only person uploading data and I do it somewhat infrequently.  Adding the upload feature to Nightscout would let other people easily manage the Omnipod upload without having to deal with installing and running programs locally, which is important for my personal situation.

It sounds like there's a (very reasonable) question about whether something like Tidepool-uploader fits into the mission of the Nightscout cgm-remote-monitor.  Currently, the PR is only in proof-of-concept state (travelling over the holidays demonstrated that I need to deal with timezones) and I haven't invested too much time into it.  I'm happy to integrate it with a currently-hypothetical upload-plugin framework, to simply keep it as a personal Nightscout customization that the larger project believes is out-of-scope, or whatever else the maintainers think is the right path.

sulkaharo:
Oops apologies, missed the earlier question - I'd be fully supportive of Tidepool Uploader -style solutions that use the REST API to enter data in formats expected by Nightscout. What I'd prefer to not happen is adding new parser dependencies into the codebase that allow uploading data in non-Nightscout formats. So from Nightscout maintenance / development perspective it'd be much better to implement a separate tool that reads the Omnipod files, converts those to Nightscout JSON files and uploads those to Nightscout using the existing REST API.

balshor:
Fair enough, I'll go ahead and keep this on my personal repo and close the PR.

PR #4162, opened 2018-12-29
bgclock 24h fix

unsoluble:
Clock display code was generating "24:xx" instead of "00:xx". Fixed, and also ensured 12-hour time would still do "12:xx".


views/bgclock.html
```diff
@@ -124,7 +124,13 @@
 			// Generate and insert the clock.
 			var timeDivisor = (window.serverSettings.settings.timeFormat) ? window.serverSettings.settings.timeFormat : 12;
 			var today = new Date(),
-			h = today.getHours() % timeDivisor || timeDivisor,
+			h = today.getHours() % timeDivisor;
```
PieterGit: I would prefer to do the time formatting in the same way within Nightscout and use a formatting function instead of doing the formatting inline.

In the frontend we use a rather outdated version of the d3 library, see https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/lib/client/index.js#L509
I don't mind switching to `moment` javascript for both basic viewing of the time and the bgclock function. See e.g.

http://momentjs.com/docs/#/parsing/string-format/
https://stackoverflow.com/questions/12970284/moment-js-24h-format

unsoluble: Sure. I'm going to need guidance on how to properly include a library like that; not clear on the broad implications.

unsoluble: Can we get this one into .11 and then do the clock rebuilds alongside the other view changes we've got slated for .12?

PieterGit:
Merging for now for 0.11. bigger clock refactors should be done in Nightscout 0.12

PR #4163, opened 2018-12-29
bugfix, if we stick to the original logic it seems to work as expected

jweismann:
cf. https://github.com/nightscout/cgm-remote-monitor/pull/4044.


PieterGit:
@jweismann I improved the fix and PR'ed it with https://github.com/nightscout/cgm-remote-monitor/pull/4253

Can please see if that PR works for you. That's a better fix than this PR.

PieterGit:
Clsoing. Fixed with https://github.com/nightscout/cgm-remote-monitor/pull/4253

jweismann:
Just tested this option and both cases works like a charm and default is to redirect as we agreed so all good now - thanks, PieterGit!

PieterGit:
@jweismann thanks for confirming that my fixes work. Thank you for initiating this PR.

PR #4178, opened 2019-01-02
Adding a Heads Up Display 

haspden:
Adding a Heads up Display page as per
[https://diabeat-this.com/xdrip-car-hud-head-up-display-page/](https://diabeat-this.com/xdrip-car-hud-head-up-display-page/)

This allows for the time & BG values to be shown on the windscreen of a car without cluttering the drivers view.

This is simply using css changes with scale to perform verticle flipping (across the horizonatal (X) axis.

Added the following code to the body css section & changed the font color to #ffffff
```
-moz-transform: scale(1, -1);
-webkit-transform: scale(1, -1);
-o-transform: scale(1, -1);
-ms-transform: scale(1, -1);
transform: scale(1, -1);
```


unsoluble:
Note that the clock in here is going to have the same 24h display problem that's fixed over in #4162 — should probably incorporate those fixes at least as a temporary measure until we get all the clocks unified as Pieter suggests.

haspden:
@unsoluble the clock was ditched in the latest version. Realised it was unnecessary. Although the script is still there the DIV doesn't exist any more. So the 12/24 hour clock issue isn't present

unsoluble:
Okay, if the clock is going to go, best to remove all reference to it.
(And I *think* your branch conflicts here are a result of working off a non-up-to-date dev branch.)

haspden:
How can I scrap this PR? I'll make a new on on a fork of the deb branch

⁣Sent from TypeApp ​

unsoluble:
Can just close it with the button at the bottom of this page on GitHub. PR the current /dev into your own dev, create a new clean HUD branch off that, then PR that into the upstream dev.

haspden:
Thanks @unsoluble

PR #4187, opened 2019-01-04
Add hud.html files to view

haspden:
Add heads up display view file to views folder


haspden:
@unsoluble finally got around to submitting that head up display file

can view the view at 
https://haspdenbloodglucose.herokuapp.com/hud.html

if you wanna see what it looks like

views/hud.html
```diff
@@ -0,0 +1,132 @@
...
+		<div class="inner">
+			<div id="trend">
+				<div id="bgnow"></div>
+				<div id="arrowDiv"><img id="arrow" src=""/></div>
+			</div>
+		</div>
+	</main>
+	
+	<script src="/api/v1/status.js"></script>
+	<script src="/js/bundle.js?v=2c1zd4vahmonsjvk"></script>
```
PieterGit: don't use hardcode bundle.js links

PieterGit: use  `<script src="/js/bundle.js?v=<%= locals.cachebuster %>"></script>`  instead

views/hud.html
```diff
@@ -0,0 +1,132 @@
+<!DOCTYPE html>
```
PieterGit: your overindenting. please follow the style rules of https://github.com/nightscout/cgm-remote-monitor/blob/master/CONTRIBUTING.md#style-guide

I know see we don't apply them to most of the other static html files as well. If you can PR that within this PR please do so.

views/hud.html
```diff
@@ -0,0 +1,132 @@
...
+			var staleMinutes = 13;										
+			var threshold = 1000 * 60 * staleMinutes;
+			
+			// Toggle stale if necessary.
+			$('#bgnow').toggleClass('stale', (now - last > threshold));
+		}
+		function init ( ) {
+			console.log('init');
+			query( );
+			setInterval(query, 1 * 60 * 1000);
```
PieterGit: see https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/lib/client/index.js on how to use `socket.io` instead of a busy wait.

views/hud.html
```diff
@@ -0,0 +1,132 @@
...
+	<main>
+		<div class="inner">
+			<div id="trend">
+				<div id="bgnow"></div>
+				<div id="arrowDiv"><img id="arrow" src=""/></div>
+			</div>
+		</div>
+	</main>
+	
+	<script src="/api/v1/status.js"></script>
```
PieterGit: append a unique identifier such as '/api/v1/status.json?t=' + new Date().getTime(); to prevent cache issues, similar to https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/lib/client/index.js#L46

PieterGit:
@haspden if you can quickly follow up on my review comments this feature might make it into the 0.11 release.

@sulkaharo @jpcunningh can one of you also review this PR?

views/hud.html
haspden: I think this is out of my skillset unfortunately... :/ any pointers @PieterGit ?

PieterGit: @haspden I already gave you a pointer 😄. You can always extend your skill set by asking for help. To be more specific have a look at this part of the client code https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/client/index.js#L897 . You need to set up a socket, and implement a custom `dataUpdate` method. 

Perhaps we need to make a minimal client-handler javascript so that that can be re-used/extended by the main Nightscout screens, and the `clock.html` and `hud.html` views. That would also solve the problems with token-based authentication. Since this method is also approved for `clock.html` we might postpone this part to 0.12 release.

I really hope @sulkaharo or @jasoncalabrese can join this discussion

unsoluble: Since this is essentially just a clone of the three existing (and stable) clock views, I think it’s reasonable to implement it as-is for 0.11 and then work on re-architecting all of the views (if that’s agreed to be appropriate) for 12.

unsoluble:
@haspden Check out our work on #4369 — wanna get this HUD view refactored to match?

unsoluble:
@joby73 Hi, really don't need to be adding your approval to all of these PRs, thanks.

PR #4189, opened 2019-01-06
Immediate full UI update after data load

sulkaharo:
This fix causes the entire UI to update after new data has landed to the client from the websocket

* Have the UI fully update immediately after new data update
* Debounce the Mongo loading for 5 seconds after data has been updated to reduce load on Mongo


PieterGit:
LGTM. Will test it when I have the time to do so (won't be today).

sulkaharo:
Merging; we've had this live for days now and haven't experienced issues

PR #4206, opened 2019-01-15
Update IFTTT maker-setup.md docs

Dave9111:
Altered the IFTTT maker-setup.md file to bring it up to date with the current IFTTT website methods and images.  
Added a local image folder so doc images can be better maintained
Update the main Readme.md file so it reflects the IFTTT doc revisions


README.md
```diff
@@ -512,29 +512,28 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
     * `PUSHOVER_ANNOUNCEMENT_KEY` - An optional Pushover user/group key, will be used for system wide user generated announcements.  If not defined this will fallback to `PUSHOVER_USER_KEY` or `PUSHOVER_ALARM_KEY`.  This also support a space delimited list of keys. To disable Announcement pushes set this to `off`.
     * `BASE_URL` - Used for pushover callbacks, usually the URL of your Nightscout site, use https when possible.
     * `API_SECRET` - Used for signing the pushover callback request for acknowledgments.
-
+
```
PieterGit: please remove extra tabs or spaces

Dave9111: Hi Pieter,
I copied the original markdown into an external editor and then rewrote everything.   The content was entirely out of date.   So a line by line comparison regarding spaces and tabs is incorrect as this isn't code.  ( for which that would make perfect sense)   I tried to mimic the original style, but thats about it.
I did a new setup of IFTTT from scratch again while I wrote that to make sure it is correct.   IMO, there was no other way to do it, since as I said, the original was so out of date.   I would encourage someone else to do the same, but it is time consuming.   Or you can just put it up and redirect any comments back to me.  I'd be happy to make any corrections if errors are found.   But I suspect it is correct as is.    Thanks for looking that over!   My suggestion to not use IFTTT and use Pushover instead is worthy of some evaluation.   The reason being that perhaps I ran into a week+ long glitch at IFTTT or something similar.    I really don't know.    Thanks, Dave

Dave9111: Pieter,  I didn't forget about doing a doc on the new features of the MongoDB maintenance interface on NS.   I have a copy of DEV loaded on a Heroku site but I haven't tested it yet with my daughters Openaps rig and Dexcom, so I don't yet feel qualified to write this doc up.    That said I will still get to it.  Thanks!    Dave

PR #4217, opened 2019-01-20
Remove unused alarm-pumpbatterylow-browser client setting

jpcunningh:
I've tested and still receive pump battery low alarms with this code removed.

### DIFF
--- a/lib/client/browser-settings.js
+++ b/lib/client/browser-settings.js
@@ -36,7 +36,6 @@ function init (client, serverSettings, $) {
     $('#alarm-timeagowarnmins-browser').val(settings.alarmTimeagoWarnMins);
     $('#alarm-timeagourgent-browser').prop('checked', settings.alarmTimeagoUrgent);
     $('#alarm-timeagourgentmins-browser').val(settings.alarmTimeagoUrgentMins);
-    $('#alarm-pumpbatterylow-browser').prop('checked', settings.alarmPumpBatteryLow);
 
     $('#nightmode-browser').prop('checked', settings.nightMode);
     $('#editmode-browser').prop('checked', settings.editMode);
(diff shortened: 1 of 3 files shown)


sulkaharo:
Did you check the git history of how these landed into the codebase?

PieterGit:
@jpcunningh @sulkaharo 
I  did a git history check. I added it in the final part of the 0.10.3 release. We had some issues with the speech plugin (solved now). It was added with commit https://github.com/nightscout/cgm-remote-monitor/commit/1769f2d837bb4e781d28647cebe974725ca702cd#diff-d878eabd64038e07845b240469aa46ca and it was part of https://github.com/nightscout/cgm-remote-monitor/commit/1769f2d837bb4e781d28647cebe974725ca702cd#diff-d878eabd64038e07845b240469aa46ca

From that issue:
I had serious problems with `Warning, Pump Battery Low` alarm. 
It was is audible, but cannot be surpressed or disabled in the UI. I started adding the UI for disabling it, but it does not seem to work yet. There is some strange issue with these pump alarms. Temporary workaround is `PUMP_ENABLE_ALERTS=false`, but of course `PUMP_ENABLE_ALERTS=true` should not lead to alarms that cannot be stopped and cannot be disabled in the UI.

I just had an empty battery, and currently it seems the battery alarm does not respect the setting in the UI.

PieterGit:
@jpcunningh @sulkaharo do you have any idea why the `alarm-pumpbatterylow-browser` does not work. I rather fix it that it works, than remove it.

bebergamota:
Cancelar. unsubscribe please


Em qui, 6 de fev de 2020 13:06, Jeremy Cunningham <notifications@github.com>
escreveu:

PR #4225, opened 2019-01-22
Update README.md for SSL config details

renegadeandy:
Added clarity to ssl config details, and specifics for the use of letsencrypt


PieterGit:
@renegadeandy thanks for the fix. merging for 0.11

PR #4215, opened 2019-01-19
Loopalyzer enhanced

lixgbg:
A PR with an enhancement to Nightscout reports Loopalyzer. This PR displays the relevant profiles (basal, CR, ISF) for the days analyzed below the charts. Very convenient to have the profiles data when analyzing how to tweak various ratios.

This PR can replace #3629 which was the original one and this PR should be merged into #4169.

/Henrik Sjöstrand

PR #4235, opened 2019-01-26
loopalyzer enhanced

PieterGit:
Extends https://github.com/nightscout/cgm-remote-monitor/pull/4215 of @lixgbg and fixes some problems of that PR.

Differences/fixes: 
- Fixed profile loading. Loads my profiles and does not get stuck in `Rendering...`
- Shows Profilenames
- changed `Show profiles table` to unchecked by default
- Fixed problem that it doesn't show the latest profile name if the last two profiles had the same basal, csf, isf

Fixes (not related to the original PR):
- drop `Unable to find element for #chartContainer` warning, because it's bugging the console log with the reports view and is a false positive. 
- `npm update`, upgrade moment, mongodb, swagger-ui-dist, uglify-js, supertest, webpack, webpack-cli

Todo: 

- set `laDebug = false` just before merge


PieterGit:
@jpcunningh or @sulkaharo Can you also review these changes. If I can fix the latest problems, I think this is ready to merge to dev. I would like to get your opinion on this PR and if it's ok to merge it to `dev` (and thus release it with Nightscout 0.11)

PieterGit:
@lixgbg Can you explain a bit more of what timeshifting does? It's not clear in the UI that it's only shown if you select multiple days. If you have only a single day it won't show the checkbox `Timeshift on meals larger than X g carbs consumed between Y and Z`.

I think this is ready for merging to dev.

lixgbg:
@PieterGit Sorry for not being responsive, I've been traveling and in job workshops for two weeks. I have been working on improving the tolerance to bad profiles, such as yours, and also to improve the deduplication. So it seems we've been working in parallel. 

I have not had a chance to look at your/this PR and verify that it works as I intended. I'm not sure about the fix to the deduplication you made. The purpose is to show a profile only when it has distinguishing changes to basal, cr, isf, but not startDate (Loop e.g. often creates copies of your profile with just a new startDate). Did your update fix any bugs there?

I'll test your version with my profiles and see if the deduplication works as expected, asap.


Regarding Timeshift -> If you have breakfast at say 7am one day and at 8am another day. Then when showing Loopalyzer over these two days you can enable Timeshift from e.g. 06:00. It will then find the first carbs logged after 06:00 on each day and when they were logged. Then it will calculate the "average time" you had these carbs, so 07:30am in this example. It will then timeshift, move, everything (carbs, bolus, basal, BG) etc the first day by +30 mins and the second day by -30 mins, so it looks like you had the breakfast at the exact same time both days. This removes the smearing of BG response when you average multiple days. It then displays this period after the "new breakfast time" with gray background so you can see more clearly. Not that ALL values for the days are timeshifted, so also values for e.g. evening snack are shifted +30min/-30min in this example.

PieterGit:
@lixgbg I made it possible to show the name of the profile. That is what the users needs for looking up the profiles (not the startDate of the profile). As far as I can tell the deduplication still works as expected.  Please check my changes, and see if you need to add changes to your PR (I don't mind which PR we use to merge). 

We should do more explanation of the timeshift function, and it should be visible regardless of whether the user has selected one day or more. It's not a good UI habit to hide/show UI elements. Perhaps it's better to use a greyed out, to show that the UI elemement is not enabled for a single day.

lixgbg:
Hi, thanks for your work. However I don't agree on the profile name. I'm a Loop user and Loop sets profile name = 'Default' to ALL profiles it creates from within the Loop app. So in my NS all profiles have the same name, hence showing it doesn't add much for Loop users. But I've kept it as your version, so displaying both name and date.

> Fixed problem that it doesn't show the latest profile name if the last two profiles had the same basal, csf, isf
This is exactly what the deduplication is supposed to do. If you have multiple profiles over the days you are showing you only want to display the oldest of the ones that have identical basal, carbratio, isf, but when a profile that has a different setting appears it should be displayed.

So if (only isf changing, for brevity):
startDate = monday, isf = 1.0
startDate = tuesday, isf = 1.0
startDate = wednesday, isf = 2.0
startDate = thursday, isf = 2.0
startDate = friday, isf = 2.0
startDate = saturday, isf = 1.0
startDate = sunday, isf = 1.0

then you want to show Monday, Wednesday and Saturday. These are when there is a change to the profile which distinguishes it from the previous one. So deduplication should remove the others.

I have pulled your version and working on debugging the deduplication. Will fix and push when I have a working version.

/Henrik

PieterGit:
@lixgbg pulled the changes on this PR to https://github.com/nightscout/cgm-remote-monitor/pull/4215 
Closing this PR and we will work together on finishing https://github.com/nightscout/cgm-remote-monitor/pull/4215

PR #4238, opened 2019-01-27
fix: Temp basal input fix for browsers that don't follow standards (Samsung phones, Kindle Fire)

unsoluble:
Way back in #2904, we removed some input styling code from the temp basal change percentage field, because it was forcing a numeric keypad to pop up on mobile browsers, and most of those keypads don't have a `-` key for entering negative amounts. All browsers tested at that time popped full keyboards after this change.

Getting reports now though that some browsers (Samsung phones, Kindle Fire) are still popping the numeric keypad, since the field has `type="number"`. Removing that input type designation to allow for negative values on all devices.

(This means a small regression in UX, as all users will now need to switch to the number display on their standard mobile keyboards before entering a value each time, but I think it's better than excluding devices from entering anything at all.)


PieterGit:
Can you be more specific? Which browsers cause these problems? Can you provide a link to the reports?

I rather that we adhere to the HTML specs and see if we can be specific in this case (preventing wrong input). See https://html.spec.whatwg.org/multipage/input.html#number-state-(type=number)

Did you also try to set `min` to a negative float and does it makes a difference to set the `step` attribute?.

I'm 👎  on merging this commit for now. Let's first see if we can find a way that those mobile browsers works, without removing `type=number`

unsoluble:
Reported on Facebook, with details in DM, so can't directly link. Kindle Fire, and the default keyboard on a Samsung S6 and S8 (Chrome or Dolphin; Firefox offers a `-` key).

Can't easily iterate change attempts when I don't have the devices to test on. But I've spent the last couple hours trying to determine how each device decides which keyboard to present, and there's no indication that any of the sub-attributes like `min` or `step` have any impact on that.

It's pretty clear that each individual browser has its own internal logic for reacting to the `type=number` tag, and I don't know if we'd ever be able to put enough conditional tweaks in there to get them all to pop the right keyboard every time. Only way to guarantee the ability to enter negatives is to allow for the full keyboard from the outset, I think.

But anyway, fine with putting this on hold until we can either dig up more functional detail or decide to scrap it.

PieterGit:
@unsoluble Found https://stackoverflow.com/questions/21677695/input-type-number-not-possible-to-insert-negative-numbers-on-mobile-phone . Normal users should not suffer from other users having broken phones. Otherwise we should add an extra environment, but parsing that also takes up CPU-cycles I would like to avoid.

PieterGit:
@unsoluble I agree it's quite a large user audience. Is this the only number field that people that has possible use of the minus key? I think this field is not used a lot. If that's the case I think it's better to drop the number requirement, but we should make sure the form has appropriate errorhandling in case an invalid character is entered.

unsoluble:
As far as I can tell, in the Careportal anyway, this is the only field that needs to accept negative values.

Do you happen to know where the form-processing code is located? I have a heck of a time navigating the flow of data in this project. :)

sulkaharo:
Oh thanks Samsung :D

unsoluble:
Think we could get this one into 0.11.2? I'm not going to be adding form validation to places where there isn't any already; that's a bigger job than I can take on right now. But this entry tweak should still be made, I think.

PieterGit:
@unsoluble I don't mind that this will be fixed in the next release. I intend to create a `0.12.0-rc.0` this weekend, to comply with semver rules. @sulkaharo agree with merging this Samsung only fix.

unsoluble:
This one didn't make it into 0.12 for some reason — can we merge it now?

PieterGit:
@unsoluble this one missed 0.11.x and we should have been able to merge it with 0.12.x
Since I really would like to follow semver I think we should merge it to dev if we know we don't have another 0.12.x patch release.

PieterGit:
@unsoluble 
Since we messed up the 0.12.x semver rules anyway, let's not be too strict.
let's not bother about semver rules for now. Merging.

PR #4250, opened 2019-02-01
API V3

PetrOndrusek:


lib/api3/const.json
```diff
@@ -0,0 +1,3 @@
+{
+  "API3_VERSION": "3.0.1"
```
PieterGit: we should use semver versioning, so I think this is `3.0.0-alpha` is more appropriate. We should change it with the release of Nightscout 0.12 to `3.0.0`

PetrOndrusek: I agree with your suggestion fully. The 3.0.1 version came from swagger online editor but we don't really need the versioning until the first official release.
So what next? Will you cancel this PR, please ? I can then prepare the version modification + some new code to review.

someoneAnyone:
I'm very much interested in the progress of the new API. If I can be of any assistance, please let me know. My app Nightscouter currently uses api/v1 a great deal.

PetrOndrusek:
@someoneAnyone Hi Pete, although the first version of APIv3 has not been released yet, I would say that it is now in testing phase, because I'm not aware of any open issues or requests for change.
@tanja3981 is the first developer who started to use APIv3, she has been connecting Android APS to NS via this new API.
If you would like to know more, you can contact me directly (https://www.facebook.com/petr.ondrusek.73) or check the docs:
https://app.swaggerhub.com/apis/PetrOndrusek/Nightscout-APIv3/3.0.0

someoneAnyone:
Thanks for sharing I'll take a look. If at all interested, I currently consume the v1 endpoints and save it in to this data format in iOS. https://gist.github.com/someoneAnyone/61e82310226028bbafd8750968775f89

PieterGit:
@PetrOndrusek can you give an update of the state of this PR and your plans with it going forward. We just released 0.12.x and should decide if we target API3 work in a next release. Since most clients probably need adaption to work with API3, we should plan the migration strategy.

PetrOndrusek:
@PieterGit Hi, APIv3 itself is prepared (including mocha tests and swagger documentation) and Socket.IO support for APIv3 is done without tests. So I would like to finish theses tests and add some more documentation, but generally the branch can be merged even now. So everything will be ready for the next release definitely.

PetrOndrusek:
From my perspective APIv3 seems to be ready for it's alpha release, I would recommend squashing before merging (there were many refactorings).

dv1:
Any updates on this?

sulkaharo:
@dv1 yup just merged it after spending an hour reading the code :D

PR #4265, opened 2019-02-06
Change install guide

unsoluble:
Just noticed that the main installation guide listed in the Readme is the [OpenAPS](https://openaps.readthedocs.io/en/latest/docs/While%20You%20Wait%20For%20Gear/nightscout-setup.html#nightscout-setup-with-heroku) one.

While we do want to get our guides all up on Readthedocs eventually, the guide over on the [wiki](http://www.nightscout.info/wiki/welcome/set-up-nightscout-using-heroku) is currently more appropriate for new users. (In particular, the OpenAPS guide instructs switching to `dev` off the bat, which is not good for general purpose installs.)


PieterGit:
Closing PR. Is integrated into dev with https://github.com/nightscout/cgm-remote-monitor/pull/4270

PR #4325, opened 2019-02-10
Sipe-update

Raijuk:


Raijuk:
I try to update nightscout. Why I can't merge like it is instructed...

PR #4349, opened 2019-02-12
Croatian translation

OpossumGit:
Croatian translation inside


PieterGit:
@OpossumGit thanks. merged to dev and will be part of the 0.11.2 
Thanks for joining as a contributor.

PR #4392, opened 2019-02-19
Changed Dockerfile.example to Alpine, Update to node 10.15.1

m-a-v:
Using the Alpine image is recommended for production use. In this way the image size can be reduced drastically (219 MB vs 1045 MB).

With node 10.15.1 it was only possible to install using the parameter --unsafe-perm. A better approach is to use a user without privileges (e.g. node).

### DIFF
--- a/Dockerfile.example
+++ b/Dockerfile.example
@@ -1,17 +1,17 @@
-FROM node:8.9.1
+FROM node:10.15.1-alpine
 
 MAINTAINER Nightscout Contributors
-
-RUN apt-get update && \
-  apt-get -y dist-upgrade
 
 RUN mkdir -p /opt/app
 ADD . /opt/app
 WORKDIR /opt/app
+RUN chown -R node:node /opt/app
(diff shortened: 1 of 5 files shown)


PieterGit:
@m-a-v you should always PR against the `dev` branch. I changed that for you.
@maskac can you confirm this PR is working for you. If it's confirmed that it works, I'll merge it to dev and release it with the next Nightscout version.

PieterGit:
@m-a-v Can you check if a `FROM node:10` also works. That way we don't have to change this file if there is an update of Node.

m-a-v:
@PieterGit Tested it with `FROM node:10-alpine` and it works perfectly. As stated above it is better to use the Alpine base image.

@PieterGit Redoing the PR means a complete new PR based on the dev branch?

PieterGit:
Merged with https://github.com/nightscout/cgm-remote-monitor/pull/4409 
Closing this PR

PR #4407, opened 2019-02-24
Wrong Korean translation update

jungsomyeong:
Wrong Korean translation update


PieterGit:
Looks good. If you want to do more maintaining of the Korean language, please also PR https://github.com/nightscout/cgm-remote-monitor/blob/dev/CONTRIBUTING.md#translators 
There aren't volunteers for the Korean language at this moment.

PR #4409, opened 2019-02-24
Changed base image to node:10-alpine

m-a-v:
@PieterGit I didn't get an answer (see #4392). I did the changes in the dev branch now. I hope this is what you expected.

### DIFF
--- a/Dockerfile.example
+++ b/Dockerfile.example
@@ -1,17 +1,17 @@
-FROM node:8.9.1
+FROM node:10-alpine
 
 MAINTAINER Nightscout Contributors
-
-RUN apt-get update && \
-  apt-get -y dist-upgrade
 
 RUN mkdir -p /opt/app
 ADD . /opt/app
 WORKDIR /opt/app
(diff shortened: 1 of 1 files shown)


PieterGit:
@m-a-v thanks for the PR. merged.

philipgo:
@m-a-v Could you maybe add some instructions, maybe as a comment in the dockerfile? For example, how do I add my config vars when running Nightscout with Docker?

m-a-v:
Normally, you will use Docker Compose to start a Nightscout instance. There are several ways to define to pass the variables:

https://docs.docker.com/compose/environment-variables

For an (old) example you can look here:
https://github.com/nightscout/nightscout-docker/blob/master/docker-compose.yml

I don't think that this should be added to Dockerfile, because it doesn't belong there.

@PieterGit: Probably we could add a sample docker-compose.yml file. On the other hand there is also the following project, which doesn't seem to be maintained anymore.

https://github.com/nightscout/nightscout-docker

Issue #4442, opened 2019-03-04
Basal plugin erroneously required for Profile Editor link


unsoluble:
When the `basal` plugin is not included in `ENABLE`, the "Profile Editor" link does not appear in the main NS menu. This should not be the case.

![screen shot 2019-03-04 at 10 24 18 am](https://user-images.githubusercontent.com/801582/53754121-d9509000-3e67-11e9-8138-0b2b234cc6f9.png)

Have tested on three different sites running 0.11.1 (although I think this behaviour predates 0.11, as I've seen the symptom earlier — just never tied it to a direct cause until now).

PieterGit:
I think this is caused by 
https://github.com/nightscout/cgm-remote-monitor/blob/48c0f59088f42ff767b401390f8aa857a1376ca1/lib/client/index.js#L1073

unsoluble:
Could you elaborate? I'm not seeing how the presence of `basal` would affect that line either way.

PieterGit:
@unsoluble Sorry to be unclear. The link to the Profile Editor is enabled with the `.profilecontrol` toggle.  See https://github.com/nightscout/cgm-remote-monitor/search?q=profilecontrol&unscoped_q=profilecontrol or specific:
https://github.com/nightscout/cgm-remote-monitor/blob/48c0f59088f42ff767b401390f8aa857a1376ca1/views/index.html#L170

I don't know yet how not including `basal` in the `ENABLE` causes that the link disappears. 
I just wanted to give you a pointer in the code.

unsoluble:
Ahhh okay, thanks for clarifying. I thought maybe you had found the link to `basal` in that line, was hurting my brain trying to see it. :)

unsoluble:
Haven't verified this yet, but possible answer: The `profile` plugin is in the defaults now, and I'm guessing this `indexOf` test isn't checking against those, but only the manually-enabled plugins. The `basal` plugin is labelled "Basal Profile" — so the test is confirming "profile" from that.

Regardless: Since `profile` isn't an optional plugin any more, I don't see any reason for this toggle check to exist; instead of fixing this line, let's just get rid of it, yeah?

unsoluble:
Fixed in #4448.

PR #4448, opened 2019-03-05
Remove profile toggle

unsoluble:
`profile` isn't an optional plugin anymore; doesn't need this display toggle. (Plus it was erroneously toggling off in the absence of the `basal` plugin — see #4442.)


PieterGit:
Please also remove `id=profilecontrol ` from 
https://github.com/nightscout/cgm-remote-monitor/blob/48c0f59088f42ff767b401390f8aa857a1376ca1/views/index.html#L170

I'm ok with removing it. @sulkaharo can you also confirm?

sulkaharo:
LGTM

unsoluble:
Hmm. Just happened to notice that the side-effect fix that this was supposed to do — to stop erroneously hiding the Profile Editor link in the main menu when `basal` isn't in ENABLE — doesn't actually work. I've scanned through every part of the code I can think of, and can't see what else might be hiding that item.
Its HTML is there, but it's being CSS/JS-disabled by something. (And it becomes visible if you add `basal` to your ENABLE string.) @PieterGit @sulkaharo Mind taking a look at some point?

PR #4483, opened 2019-03-17
cover non-Heroku secure requests in HTTPS redirect exemption

### LINKED ISSUE #4427: Infinite redirects on 0.11 upgrade
**Description**
On upgrade to 0.11, a setup stopped working as a web endpoint.

**To Reproduce**
1. Create a 0.10.3 setup using a https URL and custom port
e.g.
PORT=1234
BASE_URL=https://my.cloud.server:${PORT}
2. Upgrade to Node 10.15.1 and version 0.11
3. Reload base URL
4. Page displays "Connecting to nightscout server failed..."
5. Previously-loaded pages continue to work

**Expected behavior**
Nightscout loading page

**Your setup information**
Fedora 27
Node 10.15,1
Nightscout 0.11 (pulled master)

CGM Dexcom G5, + iOS Loop (nothing indicating those do not function)
Tried on:
- Chrome
- Brave
- Direct cURL shows infinite redirects
```
$ curl -L https://my.cloud.server:1234/
curl: (47) Maximum (50) redirects followed
```
**Additional context**
I've tried removing the port from BASE_URL and it does not resolve the issue.
I've rewound the release to the `release/0.10.3` branch and it returned to function in the same setup

thecubic:
resolves #4427 by flipping conditional and testing for `req[uest].secure` which is set on pure-HTTPS-no-middleware setups (i.e. not Heroku)

This resolves the issue for me; with this commit I do not need to set `INSECURE_USE_HTTP` to `true` which is an accidental and strange workaround being that I'm HTTPS-only full-stop.

### DIFF
--- a/app.js
+++ b/app.js
@@ -19,10 +19,11 @@ function create(env, ctx) {
     console.info('Security settings: INSECURE_USE_HTTP=',insecureUseHttp,', SECURE_HSTS_HEADER=',secureHstsHeader);
     if (!insecureUseHttp) {
         app.use((req, res, next) => {
-        if (req.header('x-forwarded-proto') !== 'https')
+        if (req.header('x-forwarded-proto') == 'https' || req.secure) {
+            next();
+        } else {
             res.redirect(`https://${req.header('host')}${req.url}`);
-        else
-            next()
+        }
         })
         if (secureHstsHeader) { // Add HSTS (HTTP Strict Transport Security) header
           const helmet = require('helmet');


PieterGit:
@thecubic What kind of reverse proxy setup do you use? 

Do you known when `req.secure` evaluates to true? For which types of reverse proxies?

I tried to integrate , https://www.npmjs.com/package/express-sslify into Nightscout, see (unfinished, not working branch) https://github.com/PieterGit/cgm-remote-monitor/commits/wip/http-redirect
specifically https://github.com/PieterGit/cgm-remote-monitor/commit/82cd05af0933ba5d19f3647fa26e793fa741fa54

We must make this unsafe connection detection work on Heroku, Azure and reverse proxies setups (like  Nginx, Apache or Letsencrypt).

According to https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_response_fields `X-Forwarded-Proto` is a de facto standard for identifying the originating protocol of an HTTP request, but it's superseded by the Forwarded header, which is https://tools.ietf.org/html/rfc7239#section-5.4

@thecubic Can you do more research on the best way that works for all hosting scenario's? I would like to have a solution that works (and is tested) for all.

thecubic:
I don't use a reverse proxy.  None is necessary for my setup, because `expressjs` supports HTTPS out-of-the-box.  Part of not using Heroku for me is the non-support of TLS for free tiers.

here's the API documentation for that variable: https://expressjs.com/en/api.html#req.secure
> A Boolean property that is true if a TLS connection is established

In that situation, a redirect doesn't make sense, hence the bug and this pull request.  If a reverse proxy is using TLS to talk to the node server (I mean, why have a reverse proxy then but ok), it will be set, but that is again a situation where you would not redirect.

> We must make this unsafe connection detection work on Heroku,

That is not affected by this pull request, since direct-to-node-server would still not have secure set, and double-fail the OR conditional

> Can you do more research

That represents major scope creep from fixing a direct bug.  I cannot.

> I would like to have a solution that works (and is tested) for all possible use cases.

In the meantime, is this PR blocked?  It passes CI and I don't see a code problem called out.

sulkaharo:
LGTM, I think this would help in many of the cases where SSL is used in self-made hosting environments. (FWIW vast majority of Nightscout users are on Azure or Heroku.)

PieterGit:
Merged. Note this is similar as the `express-sslify` module logic, see https://github.com/florianheinemann/express-sslify/blob/master/index.js

jweismann:
As to the name of the config variable `INSECURE_USE_HTTP`. I believe it would be a good idea to let the name reflect the action, e.g. `REDIRECT_2_HTTPS` to avoid confusion in places like this where `INSECURE_USE_HTTP` really means the opposite, namely stick to the HTTPS already used and don't try to add additional redirections.

Issue #4496, opened 2019-03-27
Reporting fails to render for Libre users pushing 1 CGM entry / minute into Nightscout


sulkaharo:
If a user has set a Libre / MiaoMiao to send in 1 reading / minute into Nightscout, the reporting fails to render anything. On first look, this seems to be caused by a bug in the reporting REST client, where the large amount of data from the 1 min users causes the REST call to do a chunked load, but the script expects the data to be loaded in one go, so the multiple calls to the REST loaders from the chunks resets the BG data along the way. Sadly the code is a bit of a jumble so refactoring is needed...

rob-andy:
I posted this in  #3688   . It's a pretty simple fix to get most of the reporting working.

The GVI calculation will be wrong but that's not overly difficult to fix either.

rob-andy:
I've created a reporting-fixes branch in my github repo with my current work.  Happy to hear thoughts/feedback...

sulkaharo:
@rob-andy wanna add the GVI fix too?

rob-andy:
It's in there as well.. Currently my calculated value is very different to what was originally being calculated so need some more testing to make me feel confident it's right

rob-andy:
Added some fixes to minor errors in the existing code tonight.  After unit testing using some basic sample data, I'm happy with the results.

rob-andy:
@sulkaharo no thoughts?

Issue #4490, opened 2019-03-21
glucosedistribution.js seems to have an error in the code


zreptil:
**Describe the bug**
line 410 doesn't seem to make sense:
var GVIDelta = Math.floor(glucose_data[0].bgValue,glucose_data[glucose_data.length-1].bgValue);

Math.floor calculates the next lower integer number, but doesn't take multiple values. The result of this is the value of the first glucose value in the array. Is this the expected behavior?

This has influence at the value GVI on the page distributio in the reports.

rob-andy:
This is a good pick up.  I agree 100% the correct code should be 
```javascript
var GVIDelta = Math.floor(glucose_data[0].bgValue,glucose_data[glucose_data.length-1].bgValue);
```

The delta should be the difference between the first and last reading

jakobsandberg:
We can close this issue as it was resolved with #4502

Issue #3688, opened 2018-07-12
After using Glimp / MiaoMiao all entries in nightscout (MLAB entries are existing) reports aren't shown anymore.


thomasmenges:
I'm using Freestyle Libre (**Glimp**). The CGM-Information are automatically transfered to nightscout. (MLAB entries are existing).
Since yesterday I'm using **MiaoMiao** to get the CGM-information of freestyle libre automatically every 5 minutes.
Everything is working. Even in Nightscout the is shown correctly.
But if I'm starting any report for today or yesterday no result is  shown. No entries for the day are exisiting in the report,

This behavior can be found in MASTER as well in DEV.

What can I do to resolve the problem. 
![nightscout 1](https://user-images.githubusercontent.com/37514390/42634636-059d8650-85e4-11e8-9d14-94cb9589a54f.jpg)
Here you can see the Nightscout Overview --> everything is OK

![nightscout reports](https://user-images.githubusercontent.com/37514390/42634676-1ba67bfa-85e4-11e8-9f6c-421dc278363b.jpg)
Here you can see that the entries are missing

![mongodb entry](https://user-images.githubusercontent.com/37514390/42634705-3caa9caa-85e4-11e8-80c9-f4ac6fe1a842.jpg)
Here you can see that in MongoDB an entry exists

thomasmenges:
The reason, why some values aren't shown is found (I think).
When I'm using GLIMP without connection to MiaoMiao (the CGM Reader for freestyleLibre)  everything is transfered correctly to nightscout.
Do I use Glimp connected to MiaoMiao then the values are transfered to nightscout but these entries seem to be different like the "normal" ones. 

But I'm still confused. On the overwiew of NS all values are shown correctly (screenshot 1). Only in the reports these GLIMP/MiaoMiao entries are missing

PieterGit:
@thomasmenges Can you tell me which type of hosting (Azure, Heroku, self which OS), and what MongoDB version /hosting type you use.

Can you please check  if you're not hitting one of these issues:
- http://openaps.readthedocs.io/en/latest/docs/Troubleshooting/Rig-NS-communications-troubleshooting.html
- http://openaps.readthedocs.io/en/latest/docs/Troubleshooting/Rig-NS-communications-troubleshooting.html#mlab-maintenance
[...]

PR #4502, opened 2019-03-30
Wip/libre reporting fixes

rob-andy:
Fixes to nightscout reporting module to handle libre/miaomiao data that is captured at a 1 minute per sgv rate.  This fix also corrects two issues in the GVI calculation

glucosedistribution.js
  * Data cleaning pass 1; Off by one error in for loop.  Additionally, last value in array should be retained.  Original code was losing 2 values
  * Data cleaning pass 2; Off by one error in for loop.  Additionally, last value in array should be retained.  Original code was losing 2 values
  * GVI calculation loop;  
    - Off by one error in for loop.
    - Calculate the true delta between each record
    - Calculate current rate of rise per 5min for each record
    - Calculate distance travelled based on rise and time elapsed for each record
  * Fix error in GVIDelta.  GVI Delta is the difference between last and first value

report.js
  * Allow sgvs that are exactly 1 minute from the last sgv value


rob-andy:
Adding all the issues I think this addresses;
#3688 #4490 #4496

PieterGit:
@rob-andy thanks for having a look at this. Can you:
- decrease the complexity and make Codacy happy?
- add a unit test for the edge cases we missed?

Is it possible another  libre/miaomiao user can confirm this works?

rob-andy:
Can you be more specific on decrease the complexity?  I haven't actually changed any of the existing algorithms, so would need a little more guidance on what you are expecting?

Can you also point me to the existing unit tests for these two files?

GrantGam:
...

PieterGit:
@rob-andy Some quick remarks. Didn't have a close look at the changes and because of the reformat it's hard to spot the differences:

- please change the base of the PR to `dev` instead of `master`. Please let me know if you don't have the permission to do so
- please add tests to https://github.com/nightscout/cgm-remote-monitor/blob/dev/tests/reports.test.js (we might also split the test to a test per report)
- please confirm that you've tested after these changes
- after that I think @sulkaharo or me will check if it does not have regressions for non-libre users and merge it.

josziglimp:
I would be happy to test the reports generated with this version. I'm currently using nightscout
version 0.11.1 on Heroku.
How can I merge this into my NS? Should there be any problems, is it possible to undo the merge?
Thanks

josziglimp:
I merged this and seems to be working well. Thank you!!

rob-andy:
@PieterGit adding tests for the glucose distribution report is a whole different problem.  If you are going to test it reliably, I think all the data calculation/ processing needs to be broken out so they can be tested individually (and separately to the report page UI) and then you can write a test that verifies the rUI eport page works (as per the other tests in report.test.js).

Currently nothing like this exists for any of the other reports.

rob-andy:
Additionally, the reformatting is broken out as a seperate commit. So you can easily see the logic changes, and then the formatting as per project guidelines applied to the file.  If you don't like the reformat, we can easily drop that

sulkaharo:
LGTM & tested, merging

PR #4557, opened 2019-04-25
Remove Papertrail add-on from defaults

unsoluble:
Just PRing this for consideration/discussion — not sure yet if this is the only code that would need to be modified, or if this is even a great idea overall.

That said, every few days there's a post or two on the Facebook groups from someone who's gotten an email from the Papertrail service saying they've gone over their free log limit, which causes alarm and confusion.

As far as I can tell, the benefits of having this add-on installed are very specific to developers (the standard Heroku logs are typically more than enough to diagnose common end-user problems), and it can always be added manually by anyone who wants it. Should we perhaps remove it from the set of default add-ons?


PieterGit:
@unsoluble I think I would prefer to decrease the amount of logging, so that we fit in the papertrail limits and keep papertrail enabled by default. Disabling it would require a semver minor update, so I don't want to include in a 0.11.x. release.

scottleibrand:
@PieterGit Do you have a specific proposal for which logs are unnecessary and could be removed?

sulkaharo:
I've again recently used Papertrail for debugging, so if it's removed, it'll probably make it much harder to find out why random Nightscout instances are having issues. FWIW the logs going over the quota seem to be often caused by issues (which log additional error data) so it's probably impossible to have both of 1) log errors when they happen, and 2) don't go over the limits on free logging quota

unsoluble:
Fair points. I was expecting the response to this hypothetical PR to be either a) a few reasons why it'd be nice to keep it; or b) nobody cares either way. Since it's not b), let's leave it in for now. :)

PR #4587, opened 2019-05-07
feat: Override mode plugin

jasoncalabrese:
image

This is a simple plugin that displays a pill with any overrides.  To test add `override` to the `ENABLE` setting and show the plugin from the client.


Kdisimone:
This is great stuff! very helpful!!

Kdisimone:
tested this with a couple different Loop builds, and both MDT and Pod pumps, works great. With or without the override enabled in Heroku settings. seems smooth. Thanks!

PieterGit:
@jasoncalabrese LGTM. Can you add it in the `README` as well. Please let me know if it's ready to merge.

sulkaharo:
Hmm, wonder what's happened to the automated builds - the Loop plugin test broke here

PR #4590, opened 2019-05-09
bug fix: override targets don't need to be converted

jasoncalabrese:


jasoncalabrese:
don't need this, there was a bug in the Loop code

PR #4601, opened 2019-05-10
fix loop test to expect evBG in the pill value

jasoncalabrese:
for some reason this wasn't failing on Travis, but does fail locally...


jasoncalabrese:
I see that the test did fail on travis, when merging my override-mode PR, but the build was still marked as a success

https://travis-ci.org/nightscout/cgm-remote-monitor/jobs/530079224
image

Issue #3832, opened 2018-08-12
Clock view ignores Time format


herzogmedia:
The Clock view (bgclock.html) does not show the current time in the 24 hour date format, even when it is selected in the preferences.

Is this a new bug with the latest release, have you seen this earlier?
was also there in 0.10.2

Which device / browser version was used to reproduce the bug?
Firefox 61.0.2 on Win10

Does this happen every time you launch Nightscout, or sometimes?
happens always ;)

... I can try to fix it myself but don't know if and when I'll find the time.

unsoluble:
I'm digging into this one. The issue is clear — `window.serverSettings.settings.timeFormat` is pulling straight from the config variable defaults instead of from wherever the time setting is stored in the active profile — but I haven't been able to figure out the data path yet. If someone can point me in the right direction, could fix this one.

sulkaharo:
@unsoluble by "active profile", do you mean the local client settings that can override the server setting?

sulkaharo:
Looks like the client isn't doing a full init on the clock faces, so not all variables are available, need to look at that later

unsoluble:
@herzogmedia We'll be able to fix the time thing now — will incorporate into #4639.

unsoluble:
Fixed in #4641

PR #4639, opened 2019-05-19
feat: clock view code refactor

sulkaharo:
Work in progress, do not merge yet


herzogmedia:
hey there, could you perhaps look into #3832 while you are on it?

views/index.html
```diff
@@ -129,8 +129,26 @@
             <div id="notification" role="region" aria-label="Notification">
                 <a href="#"  aria-label="cancel" ><span></span><i class="icon-cancel-circled"></i></a>
             </div>
-            <div class="primary" >
-                <div class="bgStatus current">
+
```
unsoluble: Something in the new CSS here seems to be crushing the BG section over to the left, at least in my Chrome/macOS.

Screen Shot 2019-05-19 at 11 11 55 AM

sulkaharo: Hmm, the only change to this file was supposed to be the links. Checking

sulkaharo: This should now be reverted, good catch

bundle/bundle.clocks.source.js
```diff
@@ -0,0 +1,11 @@
+
+$ = require("jquery");
+
+window.Nightscout = window.Nightscout || {};
```
jasoncalabrese: don't need this, gets overridden on line 6

unsoluble:
@sulkaharo BG positioning fix worked. Note that there's now some padding to the left of the clock section, which imho looks decent? but is a change from `master`:
Screen Shot 2019-05-19 at 12 48 14 PM
vs
Screen Shot 2019-05-19 at 12 48 21 PM

sulkaharo:
AFAIK the new padding is due to Jason fixing pill wrapping on the page, where in master if you have a lot of plugins enabled, the pills can break the layout due to not wrapping. New layout is tiny bit different but should not have this issue.

unsoluble:
Ahhhh gotcha, okay.

package.json
```diff
@@ -27,9 +27,9 @@
     "start": "node server.js",
     "test": "env-cmd ./test.env mocha --exit tests/*.test.js",
     "env": "env",
-    "postinstall": "webpack --mode production --config webpack.config.js && npm run-script update-buster",
-    "bundle": "webpack --mode production --config webpack.config.js && npm run-script update-buster",
-    "bundle-dev": "webpack --mode development --config webpack.config.js && npm run-script update-buster",
+    "postinstall": "webpack --mode production --config webpack.config.js && webpack --mode production --config webpack-clock.config.js && npm run-script update-buster",
```
jasoncalabrese: instead of doing webpack 2 times with 2 configs we could probably add an extra `entry` to the main webpack

something like this:

```
entry: {
    app: './bundle/bundle.source.js'
    clocks: './bundle/bundle.clocks.source.js'
}
...
output: {
    ...
    filename: 'js/bundle.[name].js'
}
```

PR #4644, opened 2019-05-20

Meinosaurus:
NS Update yo


unsoluble:
Gotta make sure you're logged into GitHub _before_ initiating the update process. Go ahead and close this PR, and try starting from the update tool again — should get the merge the right way around that way.

PR #4658, opened 2019-05-28
Normalize Treatment, Entry and Device Status object dates to be all in UTC Strings

sulkaharo:
Some clients enter data in ISO string format with time zone offset. This changes those dates from treatments and device statuses to comply with the UTC ISO string format, so the REST API will return data consistently for queries that search data with dates.


tests/api.treatments.test.js
```diff
@@ -60,6 +61,43 @@ describe('Treatment API', function ( ) {
     });
   });
 
+
+  it('post single treatments in zoned time format', function (done) {
+   
+    var current_time = Date.now();
+    console.log('Testing date with local format: ', _moment(current_time).format("YYYY-MM-DDTHH:mm:ss.SSSZZ"));
```
jasoncalabrese: use `moment.toISOString()` instead of the format, in a few places

lib/server/devicestatus.js
```diff
@@ -6,9 +6,23 @@ function storage (collection, ctx) {
   var ObjectID = require('mongodb').ObjectID;
 
   function create(obj, fn) {
-    if (! obj.hasOwnProperty('created_at')){
-      obj.created_at = (new Date()).toISOString();
+    
+    let d = new Date();
+
+    try {
+      if (obj.created_at) d = new Date(obj.created_at);
```
jasoncalabrese: wondering if we should use moment here, I think it can handle more date formats than the built in `new Date`

also would be nice to share this code with treatments

tests/api.treatments.test.js
sulkaharo: @jasoncalabrese Note the formats are intentionally different to test inserting a zoned time, so changing those to toISOString() would break the test

lib/server/devicestatus.js
sulkaharo: Treatments has _almost_ the same code. I hadn't known about the preBolus treatment creation before reading the test and the treatment code. Wonder if it's actually in use anywhere...

lib/api/devicestatus/index.js
```diff
@@ -24,6 +24,9 @@ function configure (app, wares, ctx) {
       q.count = 10;
     }
     ctx.devicestatus.list(q, function (err, results) {
+      results.sort(function(a,b){
+        return new Date(b.created_at) - new Date(a.created_at);
```
jasoncalabrese: I think this will override the sort, by default we already sort by `created_at`, but that can be changed with a sort param

https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/server/devicestatus.js#L44

lib/server/devicestatus.js
jasoncalabrese: Not sure how often it's used, but it's displayed on the careportal form as Carb Time for Snack and Meal Bolus

lib/server/treatments.js
```diff
@@ -143,12 +143,31 @@ function storage (env, ctx) {
...
 function prepareData(obj) {
 
+  // Convert all dates to UTC dates
+
+  let d = new Date();
+
+  try {
+    if (obj.created_at) d = new Date(obj.created_at);
+  } catch (error) {
+    console.error(error);
```
PetrOndrusek: Is catching and "consuming" a format error really what we want? In case of bad format the current datetime is going to be used. That means that original event timestamp is forgotten and suddenly the event looks like it has occured right now. Wouldn't it be better in this case interrupt the action and finish the API command with 401 HTTP status error code?

PetrOndrusek: I would also agree with your suggestion in discussion and make created_at a required field. In general I think that being more strict about incoming data is better, because then you protect all other client apps that uses this data subsequently. Being too benevolent can lead to unpredictable weird issues like we are facing now. 
So I suggest to reply with 401 every time when created_at is missing or it is malformed.

sulkaharo: No, entries with bogus data should cause an error response from the server. Sadly we have so many layers of abstraction between this and the HTTP response that this is non-trivial to implement, unless I move this code to the HTTP API layer. Maybe that's the right thing to do...

lib/api/devicestatus/index.js
jasoncalabrese: I really don't think we should be sorting the data after it's returned from mongo, the sort should be done in mongo if there's bad data, that needs to be fixed or removed from the db

sulkaharo: The date change causes the sorting to not work for queries that get mixed data. Wondering if we should implement a data migration that fixes all the data in existing databases so all the dates would in in the UTC format (along with the utcOffset field).

jasoncalabrese: but thats bad old data, this will be a performance hit for everyone that has good data, maybe it could be an option, but should still be able to handle the other sorts

sulkaharo: Yup, so sounds like a data migration on upgrade is the right way to go

jasoncalabrese: an admin tool instead of something automatic might be safer, would give people a chance to make a backup, see if they have data that needs to be migrated, etc

sulkaharo:
On a whim I realized I should check how the API is queried and the date field queries use a random assortment of formats as well. Adding features to also normalize the query parameters to the UTC format, as the query results will be inconsistent without this.

PetrOndrusek:
Yes, I have noticed this, too. Input filtering parameters also has to be normalized to UTC before filtering in mongo takes place.

PetrOndrusek:
Maybe all places in the project where mongo is queried should be checked for this, not just API...

jasoncalabrese:
I've been thinking about this some, if we're going to have to do some kind of migration, maybe we should use a mongo [`Date`](https://docs.mongodb.com/manual/reference/method/Date/) using the `ISODate()` constructor type instead of an ISO date string.  We could add a new field for that and the offset during the migration.

We'd have to test it some, but I think with that we can store as zoned or unzoned based on what we get, and then be query with either zoned or unzoned dates.

It's a big change, but it would be nice to get timestamps/timezones right.

sulkaharo:
I had the same idea, but that’d require rewriting a lot of the Mongo logic. Check how the rest api query parsing works - it seems all the logic assumes it’s strings. OTOH using dates would be the correct thing to do.

lib/server/treatments.js
sulkaharo: Changed

PR #4667, opened 2019-05-30
Fix BG & direction wrapping on Chrome

sulkaharo:
Before: Screen Shot 2019-05-30 at 21 08 22

After: Screen Shot 2019-05-30 at 21 08 35


jasoncalabrese:
@sulkaharo We should probably should move the rawbg pill into that div, could also add a class to div and move the style to the css file.  I can do that today or over the weekend if you don’t do it first.

Issue #4685, opened 2019-06-16
Open APS Bolus wizard treatments don't appear in NS


old-square-eyes:
When I use the bolus wizard on my 722, the carb and bolus treatment don't appear in NS. The same thing happens for Easy Bolus. Quick bolus DOES get displayed..

The odd thing is... they all DO appear in NSClient downstream.

Here is an example which does appear in mlab at ~Jun 14 3pm NZST...

```
"_id": {
        "$oid": "5d030edadab7d271c3267a16"
    },
    "duration": 0,
    "bolus": {
        "_type": "Bolus",
        "_description": "Bolus 2019-06-14T15:01:00 head[4], body[0] op[0x01]",
        "timestamp": "2019-06-14T15:01:00+12:00",
        "_body": "",
        "appended": [
            {
                "_type": "UnabsorbedInsulinBolus",
                "_description": "UnabsorbedInsulinBolus unknown head[5], body[0] op[0x5c]",
                "_body": "",
                "_head": "5c0528ef44",
                "data": [
                    {
                        "amount": 1,
                        "age": 239
                    }
                ],
                "_date": ""
            }
        ],
        "programmed": 0.5,
        "duration": 0,
        "amount": 0.5,
        "_head": "01050500",
        "type": "normal",
        "_date": "40814f0e13"
    },
    "timestamp": "2019-06-14T15:01:00+12:00",
    "created_at": "2019-06-14T03:01:00.000Z",
    "carbs": 4,
    "ratio": "8",
    "bg": "48",
    "glucose": 48,
    "glucoseType": "BolusWizard",
    "wizard": {
        "unknown_byte[8]": 0,
        "_type": "BolusWizard",
        "bg": 48,
        "_byte[5]": 0,
        "unknown_byte[10]": 0,
        "_description": "BolusWizard 2019-06-14T15:01:00 head[2], body[13] op[0x5b]",
        "timestamp": "2019-06-14T15:01:00+12:00",
        "_body": "04900814300005000002000582",
        "bg_target_high": 130,
        "sensitivity": 20,
        "carb_ratio": 8,
        "food_estimate": 0.5,
        "unabsorbed_insulin_total": 0.2,
        "correction_estimate": 0,
        "carb_input": 4,
        "_head": "5b30",
        "unabsorbed_insulin_count": "??",
        "_byte[7]": 0,
        "bolus_estimate": 0.5,
        "_date": "40810f0e13",
        "bg_target_low": 48
    },
[...]

PR #4697, opened 2019-06-20
Clamp treatments & load less data

sulkaharo:
* Reduces the amount of treatments loaded to the runtime signicantly (>50%)
* Clamp treatments during visualization to max BG of 22 so data with erroneous values is not drawn outside the screen


sulkaharo:
Fixes https://github.com/nightscout/cgm-remote-monitor/issues/4685

PR #4706, opened 2019-06-29
Update .nvmrc to use Node 10

jakobsandberg:
Now that we support Node 10 our `.nvmrc` should push consumers to use that as opposed to Node 8


PieterGit:
Can't test this since I'm developing on Windows. For the Nightscout 0.11.x release we currently support all active LTS releases. In the README it lists: "Latest Node 8 LTS (Node 8.15.1 or later) or Node 10 LTS (Node 10.15.2 or later; Node 10.14.1 works for Azure). Node versions that do not have the latest security patches will not work.". So there is no need to push consumers to Node 10 for Nightscout 0.11.x.
Node 8 should also work.  Is there a way to allow both Node 8 and Node 10?

I discussed with @sulkaharo and we need to choose whether the next Nightscout release will be labeled `0.11.2` or `0.12.0`. If we choose Nightscout 0.12.x, I'm fine with changing the policy to using the latest active LTS version (`^10.16.0`) for new major/minor releases. If I interpret https://github.com/nvm-sh/nvm#nvmrc correctly than I think 
`$ echo "lts/*" > .nvmrc # to default to the latest LTS version` is preferred in `.nvmrc`, so that we don't have to change anything in `.nvmrc` with future Nightscout releases.

Can you also extend the README? E.g. write about `nvm install --lts` if we choose to release the next version as `0.12.0.`.

@sulkaharo do you prefer that we release the next version as  `0.11.2` or `0.12.0.`? There are some (small) new features that would mean according to semver to release 0.12.0. I think I will prefer that. Thoughts?

sulkaharo:
Oops good catch! The right value here should be 10.15.2

sulkaharo:
Pushed in through another branch, thanks for the pointer! <3

PR #4717, opened 2019-07-04
Added sanity check for the bridge interval value

ahaarrestad:
Setting minimum all…owed value to 30 seconds and maximum to 5 minutes. This is to avoid input of very low values which might overload the dexcom servers.

### DIFF
--- a/lib/plugins/bridge.js
+++ b/lib/plugins/bridge.js
@@ -46,9 +46,17 @@ function options (env) {
     , minutes: env.extendedSettings.bridge.minutes || 1440
   };
 
+  var interval = env.extendedSettings.bridge.interval || 60000 * 2.5 // Default: 2.5 minutes
+
+  if (interval < 30000 || interval > 300000) {
+        // Invalid interval range. Revert to default
+        console.error("Invalid interval set: [" + interval + "ms]. Defaulting to 2.5 minutes.")
+        interval = 60000 * 2.5 // 2.4 minutes
+  }
+
   return {
     login: config
-    , interval: env.extendedSettings.bridge.interval || 60000 * 2.5
+    , interval: interval
     , fetch: fetch_config
     , nightscout: { }
     , maxFailures: env.extendedSettings.bridge.maxFailures || 3
--- a/lib/plugins/bridge.js
+++ b/lib/plugins/bridge.js
@@ -46,7 +46,7 @@ function options (env) {
     , minutes: env.extendedSettings.bridge.minutes || 1440
   };
 
-  var interval = env.extendedSettings.bridge.interval || 60000 * 2.5 // Default: 2.5 minutes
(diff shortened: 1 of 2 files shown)


sulkaharo:
:) https://github.com/nightscout/share2nightscout-bridge/commit/a93a79e9e9789b93313555ba5ebc4f1e43883623#diff-168726dbe96b3ce427e7fedce31bb0bcR322

lib/plugins/bridge.js
```diff
@@ -46,9 +46,17 @@ function options (env) {
     , minutes: env.extendedSettings.bridge.minutes || 1440
   };
 
+  var interval = env.extendedSettings.bridge.interval || 60000 * 2.5; // Default: 2.5 minutes
+
+  if (interval < 1000 || interval > 300000) {
```
sulkaharo: I bet 1000 would be too low, too

sulkaharo:
The ideal solution btw would be to change the Dexcom Bridge code so that it'd set the next poll based on the timestamp in the most recent record; there's no need for periodic polling when we know the next record will land in 5 minutes

ahaarrestad:
I am currently running with 10 seconds interval (dev branch). Does the code you linked to do the job? I havw verified in logs that it polls every 10 seconds.

Also agree that it should be rewritten to be smarter.

lib/plugins/bridge.js
ahaarrestad: Will not kill dexcom share to poll every second...

ahaarrestad:
> :) [nightscout/share2nightscout-bridge@a93a79e#diff-168726dbe96b3ce427e7fedce31bb0bcR322](https://github.com/nightscout/share2nightscout-bridge/commit/a93a79e9e9789b93313555ba5ebc4f1e43883623#diff-168726dbe96b3ce427e7fedce31bb0bcR322)

This fixes share_interval, but my pr fixes bridge_interval

PieterGit:
@ahaarrestad @unsoluble @sulkaharo I think I will want to create a smarter mechanism for the share2nightscout bridge with Nightscout 0.13.x (and also use Dexcom v2 api, sync calibrations, smart backfill option every hour, and only poll when we suspect a new blood glucose to arrive.

Since I would like to do a 0.13 quite soon after releasing 0.12.4 I think it's best to postpone this a bit. Agree?

lib/plugins/bridge.js
PieterGit: but will ask quite some computing resources on the nightscout instance and thus stress the cpu usage with Azure, Heroku or self hosted quite a lot. Want to look at smarter ways of implementing this.

unsoluble:
Sounds reasonable to me.

ahaarrestad:
Should i Just delete the pull request?

sulkaharo:
Merging this for now, feel free to PR a new smarter version :)

PR #4748, opened 2019-07-20
Release 0.12.2

sulkaharo:
* Adds X close buttons to all views on the site, to help unstuck iOS users who've added the site to springboard

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nightscout",
-  "version": "0.12.1",
+  "version": "0.12.2",
   "description": "Nightscout acts as a web-based CGM (Continuous Glucose Montinor) to allow multiple caregivers to remotely view a patients glucose data in realtime.",
   "license": "AGPL-3.0",
   "author": "Nightscout Team",
--- a/static/css/drawer.css
+++ b/static/css/drawer.css
@@ -227,6 +227,7 @@ h1, legend,
   overflow: hidden;
 }
 #buttonbar {
+  margin-right: 50px;
   padding-right: 15px;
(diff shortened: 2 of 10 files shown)


unsoluble:
I think this one should have more conversation before merging. The philosophy of the clock views has always been an edge-to-edge fullscreen experience with zero elements that aren't absolutely necessary — these are meant to be displayed on walls and in docks, and are as much decorative as they are functional.

And aside from all that, I don't actually see why this button helps in any way — when you're in the fullscreen view and want to get out, you just push the Home button (or swipe up on an iPhone X or later).

unsoluble:
(Also also, I think we could probably clean up the button in the Profile and Reports views while it's in dev, before pushing to release — the size and shape and centering are a little janky.)

unsoluble:
Also also also :)  The space provided for the close box appears to be nudging the whole set of toolbar icons on the main screen over to the left — these should remain tucked in the far right corner, especially on mobile.
IMG_3218

sulkaharo:
Ah, need to fix the main screen buttons. On iOS 12, the home screen apps do not restart when you “kill” the app and thus you cannot go back to the main view by killing the view. We can add a preference to make the button invisible to users who don’t want the see the button, but preferably in a manner that still allows closing the view.

unsoluble:
I still don't see the need for going back to the main view directly from a clock view — you'd simply open your site in Safari instead of launching the fullscreen pinned app.

unsoluble:
Anyway, can we revert this one for now, maybe work out the reasoning and layout stuff in dev?

PR #4771, opened 2019-07-22
[WIP] Hide unit suffix for SMBs issued by AndroidAPS

cascer1:


sulkaharo:
Great! Can you do a small refactor that’s make the code more generic in some manner, so we’d ideally have no Openaps or AAPS specific code in the renderer?

cascer1:
@sulkaharo I'd like to remove the indexOf check and rely instead only on the `treatment.isSMB` check. However, I'm not sure whether OpenAPS sets this property. I'll look into this first.

cascer1:
It seems like OpenAPS doesn't currently set the `isSMB` property. I've opened https://github.com/openaps/oref0/issues/1280 to implement this.

cascer1:
After discussing with @cluckj, it seems more difficult than I expected to have OpenAPS include the `isSMB` property when uploading bolus treatments to Nightscout. Hence, only relying on this property to render boluses as SMBs will be difficult and be against the stateless design of OpenAPS. 

Instead, we suggest adding a config var to Nightscout that lets users configure under which size boluses should be rendered without a unit (instead of the hardcoded 1U like is the case now). On top of this, we would still include the `isSMB` property since it explicitly tells us to render an SMB. The if-statement would become something like this:

```js
if (treatment.insulin < configuredSmbSize || treatment.isSMB)
```

sulkaharo:
@cascer1 That makes sense. Should we implement the check so that the isSMB would be added to treatment objects on load if the dose is under the threshold, so logic in other places could potentially use the same flag?

cascer1:
@sulkaharo I don't think adding the flag ourselves is a great idea. I think this might mislabel small boluses as SMBs which can be an issue in cases where we do more than simply render the treatment?

PieterGit:
@cascer1 Can you give an update? 
Postponing this to 0.13 release as the 0.12.4 patch release is imminent.

cascer1:
@PieterGit 	currently I can think of two options:

1. Implement the change as-is, which improves SMB rendering for AndroidAPS without changing how OpenAPS SMBs are rendered.
2. Add the `isSMB` property to all incoming bolus events under x units (where x is configured by the user), this will add the property to bolus events uploaded by OpenAPS.

I think it will be possible to pursue option 2, since I can't imagine a situation where this will have negative consequences. However, it might take a while before I have time to work on this again.

PR #4775, opened 2019-07-22
Update browser requirements

unsoluble:
0.12 significantly changes the minimum browser requirements; readme needs an update.


sulkaharo:
FWIW we didn't intentionally change this, but I guess some of the dependencies possibly broke things. Sadly I don't have a device with an old iOS to test.

PR #4788, opened 2019-07-23
Update clock view notes

unsoluble:


PieterGit:
@unsoluble Am I right this PR can be closed? I noticed @sulkaharo also changed the docs a bit.

unsoluble:
I like my wording better, and I put the views in the same order here as they are in the main menu. But sure.

PR #4793, opened 2019-07-24
feat: Less obtrusive clock close button

sulkaharo:
* Close button color is much dimmer and changes color appropriately in the color face
* HIDE_CLOCK_CLOSEBUTTON setting to hide the button altogether (set to TRUE or ON to hide buttons altogether)
* Fix CSS for toolbar positioning in the main view
* Fix documentation on the clock URLs and add mention of the setting

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,7 +309,7 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
   * `/clock/clock` - Shows current BG. Grey text on a black background.
   * `/clock/bgclock` - Shows current BG, trend arrow, and time of day. Grey text on a black background.
   * `/clock/clock-color` - Shows current BG and trend arrow. White text on a background that changes color to indicate current BG threshold (green = in range; blue = below range; yellow = above range; red = urgent below/above).
-  * Optional configuration: set `HIDE_CLOCK_CLOSEBUTTON` (`false`) to `true` to hide the small X button to close the views
+  * Optional configuration: set `SHOW_CLOCK_CLOSEBUTTON` (`true`) to `false` to hide the small X button to close the views
 
 ### Plugins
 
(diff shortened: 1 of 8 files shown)


PieterGit:
Hate to be nitty gritty. We don't have `HIDE` environment variables yet, but we do have `SHOW_`. 
Shouldn't we use `SHOW_CLOCK_CLOSEBUTTON` with a default of `true` to be consistent?

sulkaharo:
@PieterGit changed

unsoluble:
Still need to add for the record that the buttons on the clock views aren’t actually necessary in any situation, but I guess I’ve lost this one. :)

unsoluble:
Also though: This one should be pushed out now, not left for 0.13 — the toolbar spacing offset in the main view actually breaks functionality.

PieterGit:
@unsoluble can you confirm that this PR works as expected (although you had like a different default). Depending on users knowing the Apple gestures is not a good userinterface thing. Nightscout should be usable withouth Apple features. I noted clock stuff still is not token-based proof, so I can't test it.

unsoluble:
Users don't need to know Apple gestures to use the site in _nearly_ every single use case. In the _one_ specific situation where a user has pinned the main site to the home screen, and then ventures to a clock view from there, a gesture is required to get back to the main view. Honestly I'd wager this navigation path essentially never happens — adding a UI element with a controlling variable for a 0.00001% edge case like this seems like brutal overkill to me.

Will test when I get a chance, but probably can't be till later tonight.

unsoluble:
(#4788 should be included in this too.)

unsoluble:
Tested this branch — LGTM. If we default `SHOW_CLOCK_CLOSEBUTTON` to false, we can call it a truce. :)

unsoluble:
We’re going with a `false` default though, right?

PR #4795, opened 2019-07-24
Remove the Deploy To Azure button

sulkaharo:
The Deploy to Azure button was before the Heroku button; this doesn't make sense given we don't recommend Azure deploys. This PR removes the button and makes the wording for Azure deploy even more stern

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Nightscout Web Monitor (a.k.a. cgm-remote-monitor)
 [![Codacy Badge][codacy-img]][codacy-url]
 [![Gitter chat][gitter-img]][gitter-url]
 
-[![Deploy to Azure](http://azuredeploy.net/deploybutton.png)](https://azuredeploy.net/) [![Deploy to Heroku][heroku-img]][heroku-url] [![Update your site][update-img]][update-fork]
+[![Deploy to Heroku][heroku-img]][heroku-url] [![Update your site][update-img]][update-fork]
 
 This acts as a web-based CGM (Continuous Glucose Monitor) to allow
 multiple caregivers to remotely view a patient's glucose data in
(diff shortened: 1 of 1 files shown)


PieterGit:
If 0.12 does not work on Azure and nobody is willing to make it work, I would even go on and remove Azure from the supported configuration sections for Nightscout 0.13, and make it a note and clearly indicate it is not working, nor recommended and that nobody seems willing to make it work.

If we agree on this, we should also extend https://github.com/nightscout/cgm-remote-monitor/pull/4769 for 0.13 and only allow Node `^10.16.0` for Nightscout 0.13 and remove the Azure exceptions. That way it won't start on Azure (unless Microsoft updates their Node version).

sulkaharo:
I think the current Azure blocker is the Node version requirement and the fact that the free Azure tier runs out of resources due to the deploy consuming all or almost all of the resources for the day. The former is fixable, the latter isn't...

PieterGit:
@sulkaharo @unsoluble What do to with this issue? Should we change docs for 0.12.4 patch release or just wait until there is more clear about https://github.com/nightscout/cgm-remote-monitor/issues/4798

unsoluble:
Think we should remove this button in the next patch release — deploying to Azure doesn't work, so we shouldn't have a button that offers it. If/when changes happen that make it work again, we can always re-add it.

snowstorm0182:
If there is no use for the button and there is no Azure button related code somewhere else in the code then removing it should be ok. But as in #4798 sounded like few people might be using it to deploying in non free tiers, removal might not feel nice. Also removal should not lead to dead code (code debt).

I have no personal stakes here but marking the button as deprecated and moving it separate last item might be one option 

[heroku] [update] (supported)--[azure](:warning: deprecated see &lt;issue&gt;)

This way preference to heroku should be clear and people can still choose azure. In best case scenario some new Azure enthusiastic with proper skillset discovers a feasible challenge.

unsoluble:
Great points. I think demoting the button's position and adding a deprecation note is a great compromise at this point.

PieterGit:
Did not make 0.12.x. Postponing to 0.13.x
@sulkaharo Should we merge it for 0.13.x if there is no solution for a free tier for Azure yet?

PR #4799, opened 2019-07-25
feat: browser setting for openaps color lines

sulkaharo:
* Adds a method for plugins to register a boolean preference in the client
* OpenAPS now registers the color line preference for the UI using said method


sulkaharo:
Ping @Erin879 in case you want to have a look :)

sulkaharo:
Tests run:

* openaps plugin not enabled, client behaves as expected
* openaps enabled but not visible, client behaves as expected
* openaps enabled and visible, setting not set in backend, client behaves as expected
* openaps enabled and visible, setting set to both true/false in backend, client behaves as expected

PieterGit:
Test run with loop, to see if it has regressions.
- openaps plugin not enabled, client behaves as expected
- openaps enabled but not visible, client behaves as expected
- openaps enabled and visible, setting not set in backend, client behaves as expected
- openaps enabled and visible, setting set to both true/false in backend, client behaves as expected

I noticed in the frontend we call it "Show Loop Forecasts" and "Show AR2 Forecast". Shouldn't we rename Forecast to Predictions to be consistent, to be consistent?

PR #4804, opened 2019-07-26
Update Readme to reflect Azure changes

unsoluble:
Rewrote the requirements section to remove Azure support (for now).

I don't know how the `doctoc` thing works — didn't touch the TOC, but not sure if that script runs itself on new builds, or...?


unsoluble:
Just noticed @sulkaharo made a bunch of the same changes in #4802 — not sure why they weren't already in `dev` though, which is where I branched this from? Anyway, let's see how well they merge.

PR #4806, opened 2019-07-26
feat: API improvements

sulkaharo:
* change to using moment.js for API date validation
* Fix Swagger UI
* Add API documentation
* Re-enable heap dump API for debugging


lib/server/query.js
```diff
@@ -62,7 +63,15 @@ function enforceDateFilter (query, opts) {
       let dateString = dateValue[key];
       if (isNaN(dateString)) {
         dateString = dateString.replace(' ', '+'); // some clients don't excape the plus
-        dateValue[key] = new Date(dateString).toISOString();
+
+        const validDate = moment(dateString).isValid();
+
+        if (!valid) {
```
mastacheata: I think you meant to check if validDate is true.

mastacheata:
But why did you choose to include a huge library like moment.js if all you want to achieve is validating if a date conforms to the ISO format? (Note; moment() does NOT guess or convert datestrings from different formats)

sulkaharo:
@mastacheata we already use moment almost everywhere else in the codebase; this just makes the API date validation consistent with the rest of the codebase

lib/server/query.js
sulkaharo: Oops, had pushed the working code to only my personal fork, apologies

mastacheata:
sorry, didn't know that. I just started looking into the code of nightscout itself. Only ever worried about the infrastructure before.

PR #4822, opened 2019-07-28
New utils.toFixed() for cleaner labels

bassettb:
New utils.toFixedMin() rounds to the specified number of digits, omits any trailing zeros.  Allows cleaner text labels on daytoday report.  Added tests.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -25,12 +25,22 @@ function init(ctx) {
     return settings.units === 'mmol' ? Math.round(bg * 10) / 10 : Math.round(bg);
   };
 
-  utils.toFixed = function toFixed(value,digits) {
+  utils.toFixed = function toFixed(value) {
+    if (value === undefined || value === 0) {
+      return '0';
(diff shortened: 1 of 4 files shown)


bassettb:
The DayToDay report can get cluttered with labels.  Anyone doing MDI doesn't need 2 decimal places for all of their insulin doses.  Drop any trailing zeros so 3.00U is just 3U.  And even for pump users, 3.50U shows as 3.5U.  Just a little cleaner all around.

lib/utils.js
```diff
@@ -34,6 +34,15 @@ function init(ctx) {
     }
   };
 
+  utils.toFixedMin = function toFixedMin(value,digits) {
+    if (value == null || isNaN(value)) {
```
sulkaharo: `== null` checks aren't very reliable, this would probably be more robust as `if (!value || isNaN(value)) {`?

bassettb:
Thank you.

PR #4831, opened 2019-07-30
staleTime div is missing in the clockview

ireneusz-ptak:
It's implemented in the js library, but the placeholder is missing in the view.

Let me contribute! ;)

### DIFF
--- a/views/clockviews/shared.html
+++ b/views/clockviews/shared.html
@@ -33,6 +33,7 @@
         <div id="arrowDiv"><img id="arrow" src=""/></div>
       </div>
       <div id="clock"></div>
+      <div id="staleTime"></div>
     </div>
   </main>
   


unsoluble:
Good catch, thanks. Can you test to see if it needs a `display:none` in the two non-color views, or if it doesn't impede the layout without it? When we split the clocks into modular parts recently, didn't get a chance to test all of the states fully.

ireneusz-ptak:
It looks ok. I'm forcing the last update time to be displayed all the time, but that's my local change (however a candidate for a nice feature). 



Screenshot_20190731-093322_Opera

unsoluble:
Nice, thanks. I'll get @sulkaharo to take a look and then hopefully push into the next patch release.

PR #4833, opened 2019-07-30
Don't show calculated insulin next to carbs

bassettb:
In the DayToDay report, when showing a carb event, the label above the bar shows the carbs followed by an insulin amount: "30 g (2.5U)". The insulin amount is simply a calculated value using the IC ratio. But the user may not have given themselves any insulin, or they may have bolused a different amount.  This PR removes the calculated bolus.


sulkaharo:
Agree this logic has a smaller chance of mis-interpreting the reports

PR #4838, opened 2019-07-31
Rewording of the "wrong profile" dialog

unsoluble:
This dialog as originally worded often confuses/worries people when it pops on fresh installs. Reworded to be clearer about what's happening.


unsoluble:
Screen Shot 2019-07-31 at 10 55 42 AM

sulkaharo:
@unsoluble Can you also remove the old text from language.js and update this in?

unsoluble:
Ah right, will do.

unsoluble:
Done. How do we flag that passage as needing tweaks in the translated languages?

PR #4846, opened 2019-08-03
Allow start nightscout from other directories but cgm-remote-monitor

GermanBluefox:
If I start nightscout from other path, e.g. from `/opt` and nightscout is in `/opt/cgm-remote-monitor`, it tries to find `/opt/tmp` what does not exist.


sulkaharo:
Thank you!

PR #4847, opened 2019-08-03
Allow use cgm-remote-monitor as npm package.

GermanBluefox:
To use cgm-remote-monitor as npm package, the main attribute is required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "dev": "env-cmd ./my.env nodemon server.js 0.0.0.0",
     "prod": "env-cmd ./my.prod.env node server.js 0.0.0.0"
   },
+  "main": "server.js",
   "config": {
     "blanket": {
       "pattern": [


sulkaharo:
I'll totally merge this, but can you please explain the node package use case so we can potentially benefit across the board? :)

GermanBluefox:
Hi. I am a developer of ioBroker and I have integrated nightscout into it.

https://www.iobroker.net/#en/intro

Now you can just click on install in ioBroker and get nightscout installed (without azure, clouds, git and so on). Only mongodb is required.

image

The idea is to use ioBroker as an orchestration for installation of nightscout and it will be possible to send the alarms not only via pushover, but telegram, pushbullet, pushsafer, email and do automations with this values. 

image

NodeRed is integrated into nightscout too.

So to make it possible to use nightscout server easy it must be a npm package.

I have additional suggestion too for that:  
- The web pages must be checked into github and not be generated every time by installation.
- It will save install time (on raspi it is especially slow) and size of final installation, because webpack is not required.

charlenewoollard:
Can you please tell me how to stop the emails please as I’m not doing Nightscout anymore.

GermanBluefox:
> Can you please tell me how to stop the emails please as I’m not doing Nightscout anymore.

image

PR #4850, opened 2019-08-04
Add Sensor Restart Treatment type

efidoman:
Logger / Lookout will use this treatment type to enable the user to specifically specify that they are doing a restart of a sensor vs. a new insert or start. This will allow the users to explicitly denote that they do not want to clear the current local Lookout/Logger calibration cache files since the sensor is the same before/after the restart.


sulkaharo:
Hmm, adding this would be problematic for non-Logger users, as this would look like it's the right action for a restart, but no other system other than Logger would actually understand the event so logic in those systems would behave unexpectedly. So if we want this, the PR should probably also include the associated logic changes for code in Nightscout that users Sensor Start / Stop events _and_ probably some analysis of other systems that use the event (AAPS / OpenAPS without Logger, others?) so we don't shoot people using those in the foot with this.

efidoman:
Good points. I do not think that NS needs to do any further logic changes based on Sensor Restart. Are there other areas within NS that utilize Sensor Start besides The Sensor Age calculations?

https://github.com/nightscout/cgm-remote-monitor/blob/f81298fb287e4a93c4e05ef23bca5b2ac2bf6b78/README.md#sage-sensor-age

I looked through the code and that's all I could find. I do not think that Sensor Restart should change the sensor age at all. For example, if the user restarts the sensor at 7 days, then the age still should show 7 days at that point because the sensor is actually 7 days old.

I also couldn't find where Sensor Stop has any further logic in NS as well.

I agree that it for non-users of Logger/Lookout, it may look like the right thing to do, but essentially it will have no effect until those systems implement a NS "Sensor Restart" treatment response.

@scottleibrand - for his comments about adding Sensor Restart to NS and the impact to OpenAPS users that don't use Lookout/Logger.

Who can we check with about AAPS?

scottleibrand:
@sulkaharo I’m unclear what logic you’re referring to. Can you be specific? I agree that sage should not be affected by a restart.

sulkaharo:
@scottleibrand Ah apologies for unclear wording - I wasn't sure if latest OpenAPS or some component thereof monitors these events and if it'd break. Tried to make the point that if we do change the events in Nightscout in a manner that might break software depending on the events, we should try to find out if there is something depending on the semantics not changing.

What might be better than introducing a new event is introducing a new parameter to the Start event that indicates if it's a restart or a new sensor, so any logic that uses the event could continue to work as is.

scottleibrand:
Part of the reason for introducing a new restart treatment is to avoid that: most everything you’d do on a start treatment, from resetting sage to clearing calibrations etc., is a bad idea if it’s actually a restart. AFAIK the only systems that would manage the Dexcom transmitter’s view of sensor start (to handle a restart) are xdrip-J.S., xDrip+ (Logger/Lookout), and Spike, and AFAIK the latter two expect you to trigger actions through their UI, not through NS.

sulkaharo:
Yup agree 100% the current setup is not ideal - my sole point is, before we add the event, we should do exactly what you're doing above: list apps that read and write the current events and 1) check if something could break, and 2) check if some apps should be changed to accommodate the new event type (say, xDrip gaining the ability to send restarts and not just stop/starts).

scottleibrand:
Fair enough. Do you know who could do those two things for xDrip+ and Spike?

efidoman:
I added a post on both the Spike and the Xdrip+ gitter boards asking these questions.

jotomo:
@efidoman What is the gain of creating a specific event for it over a note saying _Sensor restarted_ (as xDrip does now)? As mentioned by others, no (automatic) actions need to be taken.

efidoman:
@jamorham - for comments on impact/utility to XdripPlus related to adding a new NS Treatment "Restart Sensor"

@MilosKozak - for the same regarding AAPS

@miguelkennedy_twitter - for the same regarding Spike

PieterGit:
Unless information from the other APS'es comes very quick, I postpone this to a Nightscout 0.13 release. We are about to release 0.12.4 patch release.

sulkaharo:
@efidoman ping! If you want this merged, I'd still like to see the PR have meaningful visualisation changes in Nightscout that uses this data such as, SAGE plugin showing the last restart date & data loading being changed to actually allow this. I don't see how it'd make sense to add the treatment type to the UI if entering one doesn't actually do anything in Nightscout itself.

efidoman:
@sulkaharo - Agreed. However, with the new g6 firmware not allowing "seamless" sensor resets, I'm not sure this command is going to be valuable going forward. That being the case, I suggest we consider closing this PR without merging.

Additional info ... The new g6 firmware has several issues with sensor restarts. It doesn't work some of the time and it requires physically removing the transmitter from the sensor, waiting for a bit, and reinserting. Doing a Restart in NS wouldn't work in this case, as time is needed between the Stop and the Start Sensors for this to have a chance to work.

sulkaharo:
Closing this for now - if you want the treatment merged, please reopen with logic changes to at least the sensor age plugin so it takes the treatment type into account correctly.

PR #4860, opened 2019-08-06
Fix clock if authentication is enabled

GermanBluefox:
Add .npmignore file


sulkaharo:
Not sure this needs a .npmignore? Also, see https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

unsoluble:
There are tons of changes in here — can you elaborate please on exactly what functionality has been added and/or altered?

lib/client/index.js
```diff
@@ -76,7 +76,20 @@ client.init = function init (callback) {
       console.log('Already tried to get settings after auth, but failed');
     } else {
       client.settingsFailed = true;
-      language.set('en');
+
+      // detect browser language
+      var lang = (navigator.language || navigator.userLanguage).toLowerCase();
```
sulkaharo: For language, you should check the client preferences on the language picked in the UI in the main view. Nightscout explicitly does not use the browser language detection

sulkaharo:
@unsoluble I read through the changes and everything except for the language auto-detection and the .npmignore look like they're good to go. This fixes the clock views not working on instances that don't allow API reads without authentication.

lib/client/index.js
```diff
@@ -76,7 +76,20 @@ client.init = function init (callback) {
       console.log('Already tried to get settings after auth, but failed');
     } else {
       client.settingsFailed = true;
-      language.set('en');
+
+      // detect browser language
+      var lang = Storages.localStorage.get('language') || (navigator.language || navigator.userLanguage).toLowerCase();
```
GermanBluefox: So I changed it. Please don't forget that for the VERY FIRST authentication dialog it is better to show it in the browser language, than in english.
E.g. my browser language is german and I would like to see the auth dialog in german.

lib/client/index.js
GermanBluefox: Please see https://github.com/nightscout/cgm-remote-monitor/pull/4860/files#diff-5f3d9018a260bb33cee03e0c67fdf95bR81

GermanBluefox:
Will it be merged or I should split it? (what will costs me time :) )

GermanBluefox:
> Not sure this needs a .npmignore? Also, see https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

The author forgot to write ".npmrc" into .npmignore and wrote the whole article about it on medium. Don't believe in everything what is written in internet. The authors there are just like you and me. :)

app.js
```diff
@@ -262,7 +262,7 @@ function create (env, ctx) {
   if (fs.existsSync(process.cwd() + '/tmp/cacheBusterToken')) {
     tmpFiles = express.static('tmp', {
       maxAge: maxAge
-    }); 
+    });
   } else {
```
GermanBluefox: Just formatting.

lib/client/clock-client.js
```diff
@@ -11,10 +11,27 @@ client.settings = browserSettings(client, window.serverSettings, $);
...
+  var secret = localStorage.getItem('apisecrethash');
+  var src = '/api/v1/entries.json?count=3&t=' + new Date().getTime();
+
+  if (secret) {
+    src += '&secret=' + secret;
+  } else if (token) {
+    src += '&token=' + token;
+  }
+
+  $.ajax(src, {
```
GermanBluefox: Add token or secret-hash to request to use the clock with authentication.

lib/client/clock-client.js
```diff
@@ -32,7 +49,7 @@ client.render = function render (xhr) {
   let now = new Date();
 
   // Convert BG to mmol/L if necessary.
-  if (window.serverSettings.settings.units == 'mmol') {
+  if (window.serverSettings.settings.units === 'mmol') {
     var displayValue = window.Nightscout.units.mgdlToMMOL(rec.sgv);
```
GermanBluefox: Formatting

lib/client/clock-client.js
```diff
@@ -11,10 +11,27 @@ client.settings = browserSettings(client, window.serverSettings, $);
...
+    src += '&secret=' + secret;
+  } else if (token) {
+    src += '&token=' + token;
+  }
+
+  $.ajax(src, {
     success: client.render
   });
-}
+};
```
GermanBluefox: Formatting

lib/client/clock-client.js
```diff
@@ -42,7 +59,7 @@ client.render = function render (xhr) {
   $('#bgnow').html(displayValue);
 
   // Insert the trend arrow.
-  $('#arrow').attr('src', '/images/' + rec.direction + '.svg');
+  $('#arrow').attr('src', '/images/' + (!rec.direction || rec.direction === 'NOT COMPUTABLE' ? 'NONE' : rec.direction) + '.svg');
```
GermanBluefox: If calculation cannot be done, so invalid icon was shown. This will fix the icon.

lib/client/clock-client.js
```diff
@@ -52,13 +69,13 @@ client.render = function render (xhr) {
   $('#bgnow').toggleClass('stale', (now - last > threshold));
 
   // Generate and insert the clock.
-  let timeDivisor = (client.settings.timeFormat) ? client.settings.timeFormat : 12;
+  let timeDivisor = parseInt(client.settings.timeFormat ? client.settings.timeFormat : 12, 10);
```
GermanBluefox: Just number validation.

lib/client/clock-client.js
```diff
@@ -52,13 +69,13 @@ client.render = function render (xhr) {
...
   let today = new Date()
     , h = today.getHours() % timeDivisor;
-  if (timeDivisor == 12) {
-    h = (h == 0) ? 12 : h; // In the case of 00:xx, change to 12:xx for 12h time
+  if (timeDivisor === 12) {
+    h = h || 12; // In the case of 00:xx, change to 12:xx for 12h time
```
GermanBluefox: Formatting

lib/client/clock-client.js
```diff
@@ -67,14 +84,14 @@ client.render = function render (xhr) {
...
-  
+
   if (!window.serverSettings.settings.showClockClosebutton || !queryDict['showClockClosebutton']) {
     $('#close').css('display', 'none');
   }
 
   // defined in the template this is loaded into
   // eslint-disable-next-line no-undef
-  if (clockFace == 'clock-color') {
+  if (clockFace === 'clock-color') {
```
GermanBluefox: Formatting

lib/hashauth.js
```diff
@@ -74,16 +74,18 @@ hashauth.init = function init(client, $) {
...
+
   hashauth.requestAuthentication = function requestAuthentication (eventOrNext) {
     var translate = client.translate;
     hashauth.injectHtml();
     $( '#requestauthenticationdialog' ).dialog({
-        width: 350
-      , height: 240
+        width: 400
+      , height: 270
+      , closeText: ''
```
GermanBluefox: Make a auth dialog a little bit bigger, because in other languages the text is longer and scroll bars appear.

lib/hashauth.js
```diff
@@ -74,16 +74,18 @@ hashauth.init = function init(client, $) {
...
       , buttons: [
         {
-          text: translate('Update')
+          id: 'requestauthenticationdialog-btn'
```
GermanBluefox: Search is better by ID that `$(this).parent().find('button.ui-button-text-only')`

lib/hashauth.js
```diff
@@ -102,9 +104,9 @@ hashauth.init = function init(client, $) {
         }
       ]
       , open: function open ( ) {
-        $('#requestauthenticationdialog').keypress(function pressed (e) {
+        $('#apisecret').off('keyup').on('keyup' ,function pressed (e) {
```
GermanBluefox: Install handler only on INPUT element and not on the whole dialog.

lib/hashauth.js
```diff
@@ -170,7 +172,7 @@ hashauth.init = function init(client, $) {
     var html =
       '<div id="requestauthenticationdialog" style="display:none" title="'+translate('Device authentication')+'">'+
         '<label for="apisecret">'+translate('Your API secret')+': </label>'+
-        '<input type="password" id="apisecret" size="20" />'+
+        '<input type="password" id="apisecret" size="20" style="width: 100%;"/>'+
```
GermanBluefox: make a secret input element to be a full width, because the secret could be longer than 20 chars.

lib/language.js
```diff
@@ -14143,7 +14143,7 @@ function init() {
     language.lang = newlang;
 
     language.languages.forEach(function (l) {
-      if (l.code == language.lang && l.speechCode) language.speechCode = l.speechCode;
+      if (l.code === language.lang && l.speechCode) language.speechCode = l.speechCode;
```
GermanBluefox: Formatting

views/clockviews/shared.html
```diff
@@ -30,19 +30,41 @@
     <div class="inner">
       <div id="trend">
         <div id="bgnow"></div>
-        <div id="arrowDiv"><img id="arrow" src=""/></div>
+        <div id="arrowDiv"><img id="arrow" src="" alt="arrow"/></div>
```
GermanBluefox: Just a HTML w3c requirement. Alt should be in every "img"

views/clockviews/shared.html
```diff
@@ -30,19 +30,41 @@
...
+      src += '&token=' + token;
+    }
+
+    var script = document.createElement('script');
+    script.onload = function () {
+      window.Nightscout.client.init( );
+    };
+    script.src = src;
+
+    document.head.appendChild(script); //or something of the likes
```
GermanBluefox: Load JSON with token or secret for the case that authentication is enabled.


There is still an open case if the user starts directly to work with clock view without calling the index.html. In this case no authentication dialog will be shown. But I don't know such a use case.  In this case the user can use token.

unsoluble:
Thanks for all the notes. Was really just wanting to know if any other functionality was affected, but appreciate the details. :)

lib/client/clock-client.js
PieterGit: This is not only formatting. For h == 4 this will be changed to h==12. Please restore old behaviour, perhaps with === instead of ==

views/clockviews/shared.html
```diff
@@ -30,19 +30,41 @@
...
       </div>
       <div id="clock"></div>
       <div id="staleTime"></div>
     </div>
   </main>
-  
-  <script src="/api/v1/status.js"></script>
-  <script src="<%= locals.bundle %>/js/bundle.clock.js?v=<%= locals.cachebuster %>"></script>
-  
+  <script src="/<%= locals.bundle %>/js/bundle.clock.js?v=<%= locals.cachebuster %>"></script>
```
PieterGit: locals.bundle is `/bundle` by default, so I get error:
image
remove `/` in front of it to make it work

PieterGit:
@GermanBluefox Thanks for your contributions. I tested this and did a code review of the changes. I don't understand how you have managed to test this ok, without removing the extra `/` because with the `/` it didn't work for me. I'm fine with merging this if you cherry pick or remove the `/`. I'm now able to use the clock and the implementation LGTM.
 
I'm sure you and @sulkaharo can find a solution on whether to keep or remove the `.npmignore`. I think I would advice not to merge it because I don't see the benefits of it If the discussion about that is persistent, I would suggest to open an extra PR for that and merge this without `.npmignore`.

lib/client/clock-client.js
PieterGit: fixed on my branch

sulkaharo:
Weird, I thought I posted a long piece here about the .npmignore. The main reason I'd prefer not having it is, given it's a file blacklist, the package maintainers will have to remember to consider the blacklist for every commit to the repository, while a whitelist approach would mean we detect something breaks if a mistake is made. As someone who's been looking after this repo for some years, I _really_ want to reduce the maintenance effort as much as possible, so any new code that comes in should do everything possible to the least additional effort on maintenance.

(Related - we should talk about how / where / when / why to release nightscout as a npm package. If it's done, it should be done from the original repo using the Nightscout org.

PieterGit:
merged with https://github.com/nightscout/cgm-remote-monitor/pull/4914 (without npmignore)

PieterGit:
@sulkaharo Your reaction on npmignore is https://github.com/nightscout/cgm-remote-monitor/pull/4860#issuecomment-519266463 here. Based on the article I wrote https://github.com/nightscout/cgm-remote-monitor/pull/4860#issuecomment-523182455


@GermanBluefox @sulkaharo agree to close this issue?

GermanBluefox:
It is Ok.

Issue #4859, opened 2019-08-06
Unable to submit Temporary Target Cancel - requests valid top and bottom range values


Nightfoxy:
**Describe the bug**
I try to set a Temporary Target Cancel action in the care portal, but when I submit, I am prompted to "Please enter a valid value for both top and bottom target to save a Temporary Target" and the submission is not completed. For the Temporary Target Cancel, there are no top and bottom value fields nor should there be.

**To Reproduce**
1. Go to '+' (top right, with careportal enabled)
2. Select 'Temporary Target Cancel' as the Event Type
3. Click 'Submit Form'
4. See error: 'Please enter a valid value for both top and bottom target to save a Temporary Target'

**Expected behavior**
Select Temporary Target Cancel and submit with no errors.

**Screenshots**
If applicable, add screenshots to help explain your problem
![image](https://user-images.githubusercontent.com/46848705/62557728-61c6c880-b82c-11e9-8ec1-bcb1f69e9b12.png)

**Your setup information**
- What version of Nightscout 0.12.3
- What type of CGM, and how do you get your data there? Dexcom G6 using the Dexcom app and Brdige
- Is your issue specific to a browser (Firefox/Safari/Chrome?) or a device (Android phone, etc.)? No. Replicated on Chrome Desktop and on Android Chrome as well as MS Edge

**Additional context**

unsoluble:
Is looking like this was probably an unintended side-effect of #4812.

PR #4861, opened 2019-08-06
Fixed issue with incorrect temporary target validation

Erin879:
This fixes the issue described in #4859, by only running the validation code if the temporary target event is actually a target rather than a cancel event.

### DIFF
--- a/lib/client/careportal.js
+++ b/lib/client/careportal.js
@@ -265,7 +265,7 @@ function init (client, $) {
 
     console.log('Validating careportal entry: ', data.eventType);
 
-    if (data.eventType == 'Temporary Target') {
+    if (data.duration !== 0 && data.eventType == 'Temporary Target') {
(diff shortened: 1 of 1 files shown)


Erin879:
Perhaps a better solution would be to remove the code [here](https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/client/careportal.js#L242) which turns `Temporary Target Cancel` events into `Temporary Target` events, and then that'd avoid the issue entirely. However, I'm unsure on why this code was implemented so I'd rather not include that change in a commit without other people's input.

sulkaharo:
The reason we set a new target with 0 duration is simply that the way the data is interpreted is, a new target automatically overrides the previous, and thus a 0 length target cancels the previous target without adding a new target. This is how the data is also interpreted by systems that read Nightscout, so there's no easy way to provide backward compatibility if we add a new event type

mhaeberli:
When can we expect this to be merged back into master?

PR #4865, opened 2019-08-07
Fixes #4852 (Dockerfile.example)

### LINKED ISSUE #4852: Dockerfile does not build with 0.12.3
To be able to build the Dockerfile with 0.12.3 Python and build tools need to be added.

Here is a possible solution:
https://github.com/nightscout/cgm-remote-monitor/issues/4843#issuecomment-517712085

See also here:
https://github.com/martinachtman/cgm-remote-monitor/commit/81418a2f039ce9b139ef00b4db9c81f9ce4a3c12?_pjax=%23js-repo-pjax-container

With Docker version 17.05 (release date 2017-05-04) onwards, we can use multi-stage builds. See also here:
https://github.com/mhart/alpine-node#example-dockerfile-for-your-own-nodejs-project

I will create a PR.

m-a-v:
With these changes the Dockerfile.example will becompilable again with the newest version (0.12.3).

### DIFF
--- a/Dockerfile.example
+++ b/Dockerfile.example
@@ -5,13 +5,16 @@ MAINTAINER Nightscout Contributors
 RUN mkdir -p /opt/app
 ADD . /opt/app
 WORKDIR /opt/app
-RUN chown -R node:node /opt/app
-USER node
 
-RUN npm install && \
-  npm run postinstall && \
-  npm run env && \
-  npm audit fix
+RUN chown -R node:node /opt/app && \
+    apk add --no-cache --virtual build-dependencies python make g++ && \
+    npm install --no-cache && \
+    npm run postinstall && \
+    npm run env && \
+    npm audit fix && \
+    apk del build-dependencies
+
+USER node
 
 EXPOSE 1337
 


PieterGit:
@m-a-v can you explain why you need `python`  to install it?
I don't think we should use `npm audit fix`. We will test that `npm audit ` does not have security issue when we release Nightscout. `npm audit fix` is not recommended to issue in an automated script.

m-a-v:
You should be able to explain to me better why Python is needed ;-). I know a lot about Docker but not much about the internals of cgm-remote-monitor. Without Python the Dockerfile cannot be built. 

Here is the output of the build process:

python

I can also omit `npm audit fix`. I thought sometimes it might take a few months before you release a new version and that this would be an advantage against new security vulnerabilities that might appear. That was already part of my last patch. The bug with Python appears independently.

m-a-v:
The Dockerfile.example also works with 13.0.1. I think it should be updated since the current example didn't compile.

jeromecc:
* node-heapdump requires node-gyp
* node-gyp  is a cross-platform command-line tool written in Node.js for compiling native addon modules for Node.js that requires python (v2.7, v3.5, v3.6, or v3.7)
* python is not present on node:10-alpine
* so you must install python on top of node:10-alpine

I would install python3 because Python 2 has reached end of life (EOL) on January 1st, 2020.
During my test Dockerfile built with these added packages:
```
python3 make g++
```

jeromecc:
_g++_ and _make_ can be installed with a single Alpine metapackage:
_build-base_
https://pkgs.alpinelinux.org/package/v3.9/main/x86/build-base

m-a-v:
> _g++_ and _make_ can be installed with a single Alpine metapackage:
> _build-base_
> https://pkgs.alpinelinux.org/package/v3.9/main/x86/build-base

This doesn't really matter since we only need g++ and make and both packets are deleted at the end. I didn't check if build-base installs addiontal packets/dependencies.

m-a-v:
> * node-heapdump requires node-gyp
> * node-gyp  is a cross-platform command-line tool written in Node.js for compiling native addon modules for Node.js that requires python (v2.7, v3.5, v3.6, or v3.7)
> * python is not present on node:10-alpine
> * so you must install python on top of node:10-alpine
> 
> I would install python3 because Python 2 has reached end of life (EOL) on January 1st, 2020.
> During my test Dockerfile built with these added packages:
> 
> ```
> python3 make g++
> ```

It makes sense to install python3. Did you test it with python3? I can change it for this PR.

@PieterGit Anything else that should be changed? Shall I remove `npm audit fix`? I thought there was no harm in adding this.

sulkaharo:
Apologies for the long latency - DEV now has working Docker setup, coupled with Github Actions Docker build. More improvements to land soon. https://hub.docker.com/r/nightscout/cgm-remote-monitor/tags?page=1&ordering=last_updated

PR #4810, opened 2019-07-27
Assume any string in DISPLAY_UNITS that contains "mmol" indicates mmo…

sulkaharo:
Resolves #4807

PR #4866, opened 2019-08-08
Wording tweaks to go with #4810

unsoluble:
plus a few general cleanups in the Readme


README.md
```diff
@@ -221,15 +215,15 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
 
 ### Required
 
-  * `MONGO_CONNECTION` - Your mongo uri, for example: `mongodb://sally:sallypass@ds099999.mongolab.com:99999/nightscout`
```
PieterGit: did you delete MONGO_CONNECTION on purpose?

README.md
```diff
@@ -274,26 +267,26 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs.htm
...
   * `DE_NORMALIZE_DATES`(`true`) - The Nightscout REST API normalizes all entered dates to UTC zone. Some Nightscout clients have broken date deserialization logic and expect to received back dates in zoned formats. Setting this variable to `true` causes the REST API to serialize dates sent to Nightscout in zoned format back to zoned format when served to clients over REST.
 
-
 ### Predefined values for your browser settings (optional)
+
   * `TIME_FORMAT` (`12`)- possible values `12` or `24`
   * `NIGHT_MODE` (`off`) - possible values `on` or `off`
   * `SHOW_RAWBG` (`never`) - possible values `always`, `never` or `noise`
-  * `CUSTOM_TITLE` (`Nightscout`) - Usually name of T1
-  * `THEME` (`default`) - possible values `default`, `colors`, or `colorblindfriendly`
```
PieterGit: according to source `default` is still the default, see https://github.com/nightscout/cgm-remote-monitor/blob/f81298fb287e4a93c4e05ef23bca5b2ac2bf6b78/views/index.html#L259

README.md
unsoluble: Far as I could tell, `MONGO_CONNECTION` has been deprecated; we use `MONGODB_URI` now. No?

README.md
unsoluble: Ah, yes. `colors` has been the _de facto_ default for quite a while, as it's pre-populated in the Heroku deployment template. Should update the index.html to reflect this.

PieterGit: Changing defaults should not be done in a patch release. @sulkaharo: can you reflect your view on this?

PieterGit:
@unsoluble @sulkaharo setting this for the 0.12.4 release. Let's have a look at these doc updates in the next coming days.

README.md
unsoluble: Was changed as the pre-populated default in the template back in early February after data gathering and consultation in the FB groups — vast majority of end users switch to `colors`, so made sense to default it.

README.md
sulkaharo: MONGODB_URI should be listed here. Also the MONGODB_COLLECTION is not a required variable?

unsoluble: It is required, per the Heroku template.

unsoluble: Added MONGODB_URI to the Required section.

unsoluble:
I put it there because `MONGO_COLLECTION` has had the `required` flag in the Heroku template since before I was here; figured it was set that way for a reason, so was just updating the Readme to match. If it's not required, then we should remove the flag from the template as well.

unsoluble:
These changes really should have gone in with 0.12.4, but I forgot to mention it at the time. Can we resolve the outstanding concerns? Happy to make tweaks.

PR #4884, opened 2019-08-11
Release 0.12.3

Surfside1975:
This is a bug fix / improvement point release of Nightscout Jellybean

**New features**

* Adds option to colorize the OpenAPS prediction lines, along with a new mechanism for plugins to expose their extended settings to the client UI, so this setting can be switched on and off in the client
* CarePortal now validates Temporary Target values and prevents saving entries with broken data

**Changes**

* Lowers the Node version requirement to allow Azure deploys. Azure deploys are still likely to fail on free Azure tiers due to running out of hosting resources to run the site. Microsoft is looking into this and hoping to provide us a solution to hosting Nightscout in Azure, but for now if you want to upgrade to this release, you will need to use a paid Azure tier or an alternate hosting solution like Heroku (instructions on how to migrate to Heroku are to be found at nightscout.info)
* Updated documentation regarding Azure and above mentioned configuration
* **OpenAPS 0.6 users** OpenAPS 0.6 has issues with date comparison that cause the Nightscout treatment synchronization to fail for users in time zones with the relative time earlier to UTC (UTC minus something). This release adds a workaround to the system that converts dates back to the originating form in the REST API, causing the OpenAPS date comparison logic to work as expected. You can enable this workaround by setting  DE_NORMALIZE_DATES=true in Heroku / Azure.* Fixes the button positioning the toolbar
* The clock view close X button is now controlled by a parameter in the URL. If you want to bookmark a clock view without the X, open the clock view from Nightscout navigation and remove the `showClockClosebutton=true` parameter from the page address before adding the bookmark
* Latest Stop Sensor event from last 32 days is now loaded to fix some plugin issues for users who have a sensor that was started and stopped recently but more than 2 days ago

**Developers**

* Allow Node 12 in the node version checks
* Swagger REST API documentation how works over /api-docs and developer docs in CONTRIBUTING.md have been updated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,6 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-
 # Contributing to cgm-remote-monitor
 
 [![Build Status][build-img]][build-url]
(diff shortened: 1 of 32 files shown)


sulkaharo:

PR #4895, opened 2019-08-12
German language updates

herzogmedia:
I updated the german language strings according to the change in #4851 and some new strings in the latest release


PieterGit:
@herzogmedia Thanks for the fixes. I will merge it. It will be released with Nightscout 0.12.4

PR #4943, opened 2019-08-28
Tap-to-show close buttons for clock views

unsoluble:
Decided to run with @sulkaharo 's idea of having close buttons that only appear after a tap — the JS/CSS here detects a touch event, shows the close button for a few seconds, then hides it again if it isn't used. (Buttons won't appear in desktop browsers without touch events; this is fine, they're not required there.)


sulkaharo:
Committed #4944 with a bit better state management to the event listener. The close button X tap/click registration area is weirdly shifted to the upper right corner of the screen in just the color view, ideas what in the CSS is causing this?

sulkaharo:
Merged through #4944

PR #4959, opened 2019-09-01
Rename xdrip-js to xdripjs

jpcunningh:
During the Heroku and mLab outage of August 31, 2019, I stood up a instance on a VM at home to get us through the night.

Standing it up in the VM revealed that `xdrip-js` should not have a `-` in the name due to the difficulty of setting environment variables with a `-` in them.  There is no good way to set the configuration parameters for `xdrip-js`.

This PR renames `xdrip-js` to `xdripjs` to resolve the issue.

This is a breaking change for `xdrip-js` users.


straykatz:
Will this mean a reinstall Jeremy?

jpcunningh:
No. It only requires a settings change on Nightscout.

sulkaharo:
Right, if we do renamings, my wish would be to also change the pill label to XDRIPJS, as the current "CGM" label is a bit confusing as it doesn't match the name of the plugin..?

jpcunningh:
@sulkaharo, I updated the labels to take a look at them.  I still prefer CGM for the label since it matches better with the `pump` device label.

Would it be better to rename the plugin to `cgm` instead of `xdripjs`?

unsoluble:
Feel like in terms of the wider project namespace that "CGM" is pretty broad to be using for this one plugin.

PieterGit:
@jpcunningh I think I prefer the CGM for the label as well, since it's much clearer to the user that it has CGM status/details. The UI should avoid technical sensor implementation names in it. Perhaps use `CGM status`

Do you want this in the 0.13.x release? Since it's a breaking change, we should add it into a major/minor release (and not a patch release). If @jpcunningh and @sulkaharo agree, I think this can be merged. 

I agree with @unsoluble that renaming the plugin to `cgm` is too broad, because this only affects the `xdripjs ` implementation.

Issue #5014, opened 2019-09-22
Cross on "Color" clock view is not clickable


PJHNS:
In 0.12.4, on the "Color" clock view page, the cross in the top right of the page isn't clickable (whereas the same crosses on the other clock view pages are clickable). On all three clock view pages the space immediately above the cross is clickable, so on the "Color" clock view page I need to use that area to close the page.

unsoluble:
Yep, known CSS bug. Those close buttons are primarily useful on mobile (where navigation isn't as obvious), and the targets work properly there. Something wonky in the desktop rendering of the click target, haven't had a chance to dig into it yet. If you've got any ideas, love to hear 'em.

unsoluble:
Fixed in #5018 (with a z-index on the close button).

PR #5018, opened 2019-09-22
Clock cleanup

unsoluble:
Fixes close button CSS bug (#5014), refactors CSS into less redundant files, removes `SHOW_CLOCK_CLOSEBUTTON`-related stuff (not needed any more).


sulkaharo:
@unsoluble do I read the code right that the watchface specific CSS files are actually not used at all anymore? If yes, can you update this to delete the files and I'll merge (just tested and looks good)

unsoluble:
They’re still used, just removed all redundancy; specific files now only contain specific overrides necessary for that clock.

PR #5019, opened 2019-09-23
Parse target top and bottom as integers

### LINKED ISSUE #4993: Temporary Target Top and Bottom validation problem
**Describe the bug**
Setting a temporary target with a Top value 100 or above and a Bottom value of 99 or below results in an error that says "The low target must be lower than the high target and high target must be higher than the low target." and the temporary target cannot be submitted.

If I enter the bottom value with a "0" in front like "080" and a top value over 100, then the error message is not triggered.

**To Reproduce**
1. Go to '+'
2. Choose Event Type of 'Temporary Target'
3. Enter a top value of 100 and bottom value of 90
4. Press Submit Form button and see the error.
5. Repeat step 3 with "090" and you'll see the confirmation message instead of an error.

**Expected behavior**
Should be able to enter a smaller number in the "bottom" field without a leading 0 under 100 and a top number 100 or over without an error.

**Screenshots**
image
image

**Your setup information**
- What version of Nightscout - 0.12.4
- What type of CGM, and how do you get your data there? G6 via Share
- Is your issue specific to a browser (Firefox/Safari/Chrome?) or a device (Android phone, etc.)? Both Windows and Android Chrome is what I tested in.

**Additional context**

p5nbTgip0r:
Fixes #4993.
The input variables `data.targetTop` and `data.targetBottom` are strings, so the `targetTop` and `targetBottom` variables were being assigned as strings too. This caused the verification checks to give incorrect results because those variables were not treated as numbers. Parsing the input variables as integers will convert them to the number type and allow these checks to work as intended.

### DIFF
--- a/lib/client/careportal.js
+++ b/lib/client/careportal.js
@@ -272,8 +272,8 @@ function init (client, $) {
         messages.push("Please enter a valid value for both top and bottom target to save a Temporary Target");
       } else {
 
-        let targetTop = data.targetTop;
-        let targetBottom = data.targetBottom;
+        let targetTop = parseInt(data.targetTop);
+        let targetBottom = parseInt(data.targetBottom);
 
         let minTarget = 4 * 18;
         let maxTarget = 18 * 18;


unsoluble:
Can't believe my eyes scanned right over those string assignments — knew it had to be a type issue somewhere, good catch. :)

PR #5038, opened 2019-09-30
Fix mg/dl /U sensitivity which does not need to by multiplied by 18

WhatWouldHoneyBadgerDo:
For example a sensitivity of 84 mg/dl /U is showing as 1512 mg/dl /U in the basal pill.


WhatWouldHoneyBadgerDo:
Hmmm, let me investigate a little further.

PR #5043, opened 2019-09-30
Remote Loop overrides

ps2:
This PR adds support to Nightscout for a new Loop remote override feature. Users can use careportal (or IFTTT) to trigger a Loop override remotely.  The override will be enacted immediately, via push notification, if the Loop app is online, or if Loop gains network connectivity within 5 minutes, and the device running Loop will receive a user notification that the override has been enabled.
careportal_entry
range_and_popup
override_notification

To test this feature, check out https://github.com/LoopKit/Loop/pull/1131


bewest:
Very cool feature.

muttech:
Treatments that don't have a duration (CGM Sensor Insert and BG Check are the two I've seen thus far) are being displayed with an Indefinite Duration on the graph:
image

lib/client/renderer.js line 312 looks like it's filtering specific types:
!_.includes(['Temp Basal', 'Profile Switch', 'Combo Bolus', 'Temporary Target'], treatment.eventType);

ps2:
Uses a separate representation of indefinite duration for overrides, rather than overloading '0'. @muttech, would appreciate testing.  Needs the latest `remote-overrides` branches in Loop for Loop to post new indefinite duration overrides properly. Indefinite duration overrides previously set with earlier versions of the `remote-overrides` branches in Loop will not display correctly.

Kdisimone:
Tested this out last night and today...everything is displaying properly for Loop, overrides and old style treatments (cage, sage, etc) too. 

(Also helped two others test this out and their sites are also functioning properly.) 👍

tiaberend:
I have tested this as of last night. The overrides work great. Also the indefinite duration of the sage and cage is no longer an issue. Thank you!

jbwittmer:
I'll try to test tonight.  WE had encountered same problem prior and reverted to prior version.

Kdisimone:
Just testing new updates on two sites and two loops...all working great. Push to NS from loop is quick and displays are correctly done. Indefinite overrides are properly displayed while active, and end correctly when canceled (display bars are also ended visually well). Shortcuts are working great too, no problems with those. Also had a loop failure to upload to ns from a network connection error and during the next loop run the upload went fine without any intervention, so failure recovery was great.

Kdisimone:
also tested sage, cage and bage entries...all displayed properly

PieterGit:
Tested this and the Loop update. Works like a charm. Let's merge the Nightscout part to `dev`, once @sulkaharo had a quick look and @ps2 says it's good to merge.

ps2:
I have one more change coming, that configures the apn server for dev or production based on the provisioning profile included in the Loop app.

PieterGit:
Just created a `wip/remote-loop-overrides` branch on the cgm-remote-monitor repo, see https://github.com/nightscout/cgm-remote-monitor/tree/wip/remote-loop-overrides . I already got 3 PR's of people wanting to update their own fork. 

@ardillen @shabanaj25 @Groob3210 Please report back here if you have succes or problems with @ps2 pull request for remote loop overrides.

shabanaj25:
> Just created a `wip/remote-loop-overrides` branch on the cgm-remote-monitor repo for convenience, see https://github.com/nightscout/cgm-remote-monitor/tree/wip/remote-loop-overrides . I already got 3 PR's of people wanting to update their own fork.
> 
> @ardillen @shabanaj25 @Groob3210 Please report back here if you have succes or problems with @ps2 pull request for remote loop overrides.

I am up and running! Thank you.

ps2:
Ok, I was unable to get automatic detection of aps-environment working. There is no official api to get this on iOS, and the hacks to get it are unreliable.  So there is a new environment variable `LOOP_PUSH_SERVER_ENVIRONMENT` that can be set to `production` if you are one of the (rare) people that use TestFlight or build your Loop with a production aps-environment provisioning profile via some other mechanism.

ps2:
I believe this PR is ready.

ardillen:
I updated repo and deployed from heroku and still getting same loop error “could not find device token in loop settings”.  Not sure what I’m missing…

Thanks,
Amy Dillenbeck (@ardillen)

francescaneri:
same thing. Thanks

ps2:
The device token will be uploaded to Loop the first time you run a version of Loop with Remote Overrides in it, and changing something in settings. Turning on an override locally is enough to trigger a settings upload.

lib/server/loop.js
```diff
@@ -52,12 +52,16 @@ function init (env, ctx) {
         , keyId: env.extendedSettings.loop.apnsKeyId
         , teamId: env.extendedSettings.loop.developerTeamId
       },
-      production: false
+      production: env.extendedSettings.loop.pushServerEnvironment === "production"
```
PieterGit: please be lenient with case errors and use a `toLowerCase()`, similar to https://github.com/nightscout/cgm-remote-monitor/blob/f81298fb287e4a93c4e05ef23bca5b2ac2bf6b78/env.js#L26

ps2:
@PieterGit thanks for the review comments.

It seems fine to me to be strict about this kind of settings input. Also would need to cover undefined separately to avoid calling .toLowerCase() on undefined.

As for the docs changes, the section is already presuming people know how to add entries to heroku app settings, so I'm not sure why there are specific instruction telling them how to do that for these variables. Also, there is much more than what you've covered that a user needs to know about setting these up. I think loopdocs will will be the primary guide for setting up Remote Overrides in Loop. Do you really want to add more details here that risk becoming out of date?  I know loopdocs will be updated with the latest full info.  It seems sufficient here to document what these settings are for, as a comprehensive list of what is configurable in NS.

For both of these requests, this isn't a hill I'll die on, so if they're needed, I'll comply. Just thought I'd share my thoughts on the matter.

PieterGit:
@ps2 No problem with adding extra info to the loopdocs instead of the Nightscout docs. I noticed the text in your PR was better and more readable than the Nightscout documentation. Since the `LOOP_PUSH_SERVER_ENVIRONMENT` is only used by a few people, I can understand your point. We had quite some issues with users using uppercase instead of lowercase with other environment variables, so I just wanted to be consistent with that. It's just a minor thing in this case.

Let's wait for @sulkaharo to agree on this PR. Since it's a new (cool) feature I would like to get a bit more core developer feedback before merging it.

sulkaharo:
LGTM. Can't test with Loop personally. Merging to dev for testing with a larger group

Issue #4, opened 2014-06-01
pebble endpoint does not support trend


bewest:
As a result the pebble watchface does not draw arrows, and people do love the arrows.

Issue #5075, opened 2019-10-12
Upgrade D3 to the latest version


sulkaharo:
Nightscout currently uses the no longer maintained D3 version 3.5.17. The graph renderer should be refactored to be easier to maintain, not leak memory and to use a newer version of D3 that is still maintained.

jpcunningh:
@sulkaharo, I started looking at upgrading D3 last night.  I'm willing to give it a shot this week and see how far I can get.

alphacentauri82:
On it too

PR #5081, opened 2019-10-14
D3 upgrade (library for visualizing data such as main view, graphs, etc.)

jpcunningh:
This is in reference to issue #5075.

### DIFF
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -4,8 +4,6 @@
 var times = require('../times');
 var d3locales = require('./d3locales');
 
-var focusMin = 30;
-
 var PADDING_BOTTOM = 30
   , CONTEXT_MAX = 420
   , CONTEXT_MIN = 36
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -50,16 +50,16 @@ function init (client, d3, $) {
   
   function brushStarted ( ) {
     // update the opacity of the context data points to brush extent
-    chart.context.selectAll('circle')
-      .data(client.entries)
-      .style('opacity', 1);
+    //chart.context.selectAll('circle')
+    //  .data(client.entries)
+    //  .style('opacity', 1);
   }
 
   function brushEnded ( ) {
     // update the opacity of the context data points to brush extent
-    chart.context.selectAll('circle')
-      .data(client.entries)
-      .style('opacity', function (d) { return renderer.highlightBrushPoints(d) });
+    //chart.context.selectAll('circle')
+    //  .data(client.entries)
+    //  .style('opacity', function (d) { return renderer.highlightBrushPoints(d) });
   }
 
   var extent = client.dataExtent();
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -50,16 +50,16 @@ function init (client, d3, $) {
   
   function brushStarted ( ) {
     // update the opacity of the context data points to brush extent
-    //chart.context.selectAll('circle')
-    //  .data(client.entries)
-    //  .style('opacity', 1);
+    chart.context.selectAll('circle')
+      .data(client.entries)
+      .style('opacity', 1);
   }
 
   function brushEnded ( ) {
     // update the opacity of the context data points to brush extent
-    //chart.context.selectAll('circle')
-    //  .data(client.entries)
-    //  .style('opacity', function (d) { return renderer.highlightBrushPoints(d) });
+    chart.context.selectAll('circle')
+      .data(client.entries)
+      .style('opacity', function (d) { return renderer.highlightBrushPoints(d) });
   }
 
   var extent = client.dataExtent();
@@ -200,7 +200,7 @@ function init (client, d3, $) {
   chart.context.append('g')
     .attr('class', 'y axis');
 
-  chart.createRange = function () {
+  chart.createBrushedRange = function () {
     var brushedRange = chart.theBrush && d3.brushSelection(chart.theBrush.node()) || null;
     var range = brushedRange && brushedRange.map(chart.xScale2.invert);
     var dataExtent = client.dataExtent();
@@ -222,7 +222,7 @@ function init (client, d3, $) {
   }
 
   chart.createAdjustedRange = function () {
-    var adjustedRange = chart.createRange();
+    var adjustedRange = chart.createBrushedRange();
 
     adjustedRange[1] = new Date(adjustedRange[1].getTime() + client.forecastTime);
 
@@ -270,7 +270,7 @@ function init (client, d3, $) {
 
     // get current brush extent
     var currentRange = chart.createAdjustedRange();
-    var currentBrushExtent = chart.createRange();
+    var currentBrushExtent = chart.createBrushedRange();
 
     // only redraw chart if chart size has changed
     var widthChanged = (chart.prevChartWidth !== chartWidth);
@@ -532,6 +532,11 @@ function init (client, d3, $) {
       }
     }
 
+    // Refresh currentBrushExtent to catch when we are not in Retro mode
+    // and an update comes in during execution of the code above that could
+    // drive us into retro mode.
+    currentBrushExtent = chart.createBrushedRange();
+
     // update domain
     chart.xScale2.domain(dataRange);
     chart.xScaleBasals.domain(dataRange);
@@ -548,7 +553,7 @@ function init (client, d3, $) {
   };
 
   chart.scroll = function scroll (nowDate) {
-    var currentBrushExtent = chart.createRange();
+    var currentBrushExtent = chart.createBrushedRange();
     var currentRange = chart.createAdjustedRange();
 
     chart.xScale.domain(currentRange);
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -210,18 +210,20 @@ function init (client, d3, $) {
       range[0] = new Date(range[1].getTime() - client.focusRangeMS);
     }
 
+    var end = range[1].getTime()
+    if (!chart.inRetroMode()) {
+      var lastSGVMills = client.latestSGV ? client.latestSGV.mills : client.now;
+      end += (client.now - lastSGVMills);
+    }
+    range[1] = new Date(end);
+
     return range;
   }
 
   chart.createAdjustedRange = function () {
     var adjustedRange = chart.createRange();
 
-    var end = adjustedRange[1].getTime() + client.forecastTime;
-    if (!chart.inRetroMode()) {
-      var lastSGVMills = client.latestSGV ? client.latestSGV.mills : client.now;
-      end += (client.now - lastSGVMills);
-    }
-    adjustedRange[1] = new Date(end);
+    adjustedRange[1] = new Date(adjustedRange[1].getTime() + client.forecastTime);
 
     return adjustedRange;
   }
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -319,7 +319,9 @@ function init (client, d3, $) {
         // disable resizing of brush
         chart.context.select('.x.brush').select('.overlay').style('cursor', 'move');
         chart.context.select('.x.brush').selectAll('.handle')
-          .style('cursor', 'move')
+          .style('cursor', 'move');
+
+        chart.context.select('.x.brush').select('.selection')
           .style('visibility', 'hidden');
 
         // add a line that marks the current time
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -172,7 +172,7 @@ function init (client, d3, $) {
 
   chart.xAxis = d3.axisBottom(xScale)
     .tickFormat(tickFormat)
-    .ticks(4);
+    .ticks(6);
 
   chart.yAxis = d3.axisLeft(yScale)
     .tickFormat(d3.format('d'))
@@ -185,6 +185,9 @@ function init (client, d3, $) {
   chart.yAxis2 = d3.axisRight(yScale2)
     .tickFormat(d3.format('d'))
     .tickValues(tickValues);
+
+  d3.select('tick')
+    .style('z-index', '10000');
 
   // setup a brush
   chart.brush = d3.brushX()
@@ -223,22 +226,26 @@ function init (client, d3, $) {
 
   // create the x axis container
   chart.focus.append('g')
-    .attr('class', 'x axis');
-
+    .attr('class', 'x axis')
+    .style("font-size", "16px");
+    
   // create the y axis container
   chart.focus.append('g')
-    .attr('class', 'y axis');
+    .attr('class', 'y axis')
+    .style("font-size", "16px");
 
   chart.context = chart.charts.append('g')
     .attr('class', 'chart-context');
 
   // create the x axis container
   chart.context.append('g')
-    .attr('class', 'x axis');
+    .attr('class', 'x axis')
+    .style("font-size", "16px");
 
   // create the y axis container
   chart.context.append('g')
-    .attr('class', 'y axis');
+    .attr('class', 'y axis')
+    .style("font-size", "16px");
 
   chart.createBrushedRange = function () {
     var brushedRange = chart.theBrush && d3.brushSelection(chart.theBrush.node()) || null;
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -221,17 +221,14 @@ function init (client, d3, $) {
   }
 
   chart.inRetroMode = function inRetroMode() {
-    if (!chart.brush || !chart.xScale2) {
+    var brushedRange = chart.theBrush && d3.brushSelection(chart.theBrush.node()) || null;
+
+    if (!brushedRange || !chart.xScale2) {
       return false;
     }
 
-    var brushedRange = chart.theBrush && d3.brushSelection(chart.theBrush.node()) || null;
-    var maxTime;
-    var brushTime = maxTime = chart.xScale2.domain()[1].getTime();
-
-    if (brushedRange) {
-      brushTime = chart.xScale2.invert(brushedRange[1]).getTime();
-    }
+    var maxTime = chart.xScale2.domain()[1].getTime();
+    var brushTime = chart.xScale2.invert(brushedRange[1]).getTime();
 
     return brushTime < maxTime;
   };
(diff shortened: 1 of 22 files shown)


jpcunningh:
Definitely a work in progress!

sulkaharo:
Yay! Note that refactoring of the renderer to introduce a more generic API would be _awesome_. The renderer should have 0 lines of code that are specific to Nightscout clients / data sources. For example, a better structure would be to add new plugin API calls to give the renderer things like predictions & move code related to that into the relevant plugins.

unsoluble:
In Chrome here on macOS 10.15 I'm getting clipping of low-end data points by the scrubber viewport thing:

Screen Recording 2019-10-17 at 11 48 50 PM

unsoluble:
Also the grab box for that viewport is oddly small — previously, could grab anywhere in the 48hr view and have the viewport jump to where you clicked. Here, need to grab inside a hitbox that's only the top third or so of the inset viewport.

PieterGit:
@unsoluble thanks for testing.

@jpcunningh thanks for this great effort! I put it for 0.13.x, but we should see if that is feasible.

sulkaharo:
Great job! The old and new plots are practically identical. :)

* The font in the X and Y axis for time and CGM values has changed and is much smaller than before
* As noted above, the brush used to allow tapping anymore, doesn't anymore
* The graph has a new behaviour where it first renders so it uses maybe 20% of screen height, then scales to fit the screen. Wonder if there's a way to have the graph scale to right height from the get go?
* Looks like this branch for some reason doesn't default to the units set in the backend for the client, had to set it to mmol manually. Did something change there? The client is expected to default to units from DISPLAY_UNITS

sulkaharo:
Re: code - refactoring this is still very welcome. Things that could be useful:
* Break the mega long functions to smaller bits that deal with one part of the render at a time and name the functions descriptively
* Rename existing functions to describe the function better and remove typos
* Add jsdoc or some other form of comments to make it easier to get hold of what's going on in the  renderer

Busteren:
Can't find where the specific comments about the brush not working on touch are. But I just wanted to note that it won't work on the Chrome Developer Mobile Emulator: [d3 touchable](https://github.com/d3/d3-brush#brush_touchable)

So testing this on a device is recommended.

inventor96:
I have a similar issue as the one @unsoluble presented, but it appears to be a black bar on top of the points, not just making them disappear. Not sure if it's because they're prediction lines instead of real data, though?

image

Also, I get some weird jumping around as I scroll through my history, and sometimes the dotted vertical line on the bottom is outside the two solid lines (I don't think that's supposed to happen? I don't recall seeing it happen before).

Here's a screen recording that shows all of these: https://drive.google.com/file/d/1HKcO56rha_6j4uxJNOmgm-peeC6TsGi1/view?usp=sharing

This is in Chrome 77.0.3865.120 (Official Build) (64-bit) on MacOS 10.15 (Catalina).

unsoluble:
Tooltip popover when hovering over an SGV point is rendering on the far left (as if x=0), but y height is still correct.

bartsluimer:
Schermafbeelding 2019-10-18 om 21 19 53
Sometimes in safari the min and max dot lines are gone. When refreshing they are there again.

jpcunningh:
Thanks for all the great feedback!

I’m on the road today, but will take a look this evening or tomorrow to try
to fix the issues.

inventor96:
Here's an example of the tooltip popover getting stuck on the left.
image

PieterGit:
Report from @Kdisimone on Discord: Treatment reports appear to be stuck on "rendering".
https://media.discordapp.net/attachments/629952669967974410/634807701875982387/image0.png?width=388&height=754

Cause is in Javascript errors in `calibration.js` report.  It should use https://github.com/d3/d3-scale/blob/master/README.md#scaleLinear and https://github.com/d3/d3/blob/master/CHANGES.md#axes-d3-axis syntax
(and it should not influence another report 😟 )

I started fixing the D3 issues in the calibration reports. Please 
cherry-pick 3445eeeedd1c0d863ca5210d4cd32994a9da4d49  06a8a9283100a980ccb5bed7920bce96a5b138cc 1ac583ce2b1d0796f8d8f0d5f9b5ee29a051af1e on this PR from https://github.com/nightscout/cgm-remote-monitor/commits/wip/013alpha

I will stop now, and currently get stuck on the next error: 

image

Kdisimone:
Loop/pump pills disappeared when BG data went stale.

jpcunningh:
I cherry-picked the three commits.

I haven't made it over to the reports, yet. It looks like week2week,
day2day, and calibrations all need work.

I forced our BG data to go stale my pills are still visible. Can you
describe more about what happened with the pills disappeared, @Kdisimone?

Thanks!

jpcunningh:
@unsoluble, the tooltip location should be fixed now. Thanks!

@unsoluble and @inventor96, the solid vertical line, dashed vertical line, and the black bar at the top you noticed hiding glucose readings are being positioned by the code on every brush event.  The black bar should simply cover the x-axis of the focus chart to make it appear to disappear for the portion of the context chart at the bottom that is visible in the focus chart at the top.  The vertical bars should end at the x-axis of the focus chart.  They are definitely not being drawn with the correct y value for for the bottom and top, and maybe not the correct x value sometimes.

I've tested on Chrome, Firefox, and Edge on Windows and Chrome and Firefox on Ubuntu, and haven't been able to make it drawing the lines with the wrong y values like shown in your screenshots. 

Unfortunately, I don't have a Mac to test on.  Do they always extend past above the x-axis of the focus (top) chart? How does the window size effect how far past the x-axis it extents?

Thanks!

unsoluble:
@jpcunningh Oh I see what's happening there, gotcha. Yeah, the three vertical lines and the black crossmember are always drawn too high (in both Safari and Chrome, on macOS 10.15 here). Doesn't matter what viewport size, they always seem to be drawn with the same proportions.

PieterGit:
I pushed your changes to the https://github.com/nightscout/cgm-remote-monitor/tree/wip/013alpha as well so others can help testing. @jpcunningh can you also push to that branch. I gave you permissions for it.

PieterGit:
@jpcunningh I can confirm the tooltip is fixed for treatments and pills on the left side. With pills on the right side (under the bloodglucose, e.g. IOB and BASAL) are shown half way and not to the right with a wide screen (Chrome, Windows 10). When I use inspect i think it's with the `span.pills.iob`, `span.pills.basal`).

jpcunningh:
I pushed the the update for the header tooltips to 013alpha.

I should be able to get to the update for the reports today in a few hours.

I am on the lookout for a Mac so I can see what’s going on with why the
layout for the brush boundaries are not working properly on Macs. I’m not
sure when I’ll be able to find one, though, so any help would be
appreciated. Just looking at the screenshots, it looks like the scales
offset isn’t being applied properly on the Mac.

@PieterGit, for future updates relative to this PR, would you like me to
maintain them here and cherry pick them over to whip/013alpha?

Thanks!

unsoluble:
If you could briefly document the elements for us, could put some more eyes on the code — as it is, I can't see at a glance where/what anything relevant is. (What's `focus` vs `context`, for example? Trying to sort out where all the calls involving the bugged drawing are happening.)

jpcunningh:
@unsoluble, the context section contains the graph that holds the 48 hours of glucose readings and provides the brush surface.  The focus section holds the top graph.

The code for bugged drawing is in `chart.js`. The left line is called `open-left`, the right line is `open-right, and the top black bar that covers the focus area's x-axis is `open-top`.

jpcunningh:
Reports are updated now to d3 v5.

The updates are also pushed to `wip/013alpha`

lib/client/chart.js
```diff
@@ -235,15 +280,16 @@ function init (client, d3, $) {
     var dataRange = client.dataExtent();
     var chartContainerRect = chartContainer[0].getBoundingClientRect();
     var chartWidth = chartContainerRect.width;
-    var chartHeight = chartContainerRect.height - padding.bottom;
+    var chartHeight = chartContainerRect.height - PADDING_BOTTOM;
 
     // get the height of each chart based on its container size ratio
     var focusHeight = chart.focusHeight = chartHeight * .7;
-    var contextHeight = chart.contextHeight = chartHeight * .2;
+    var contextHeight = chart.contextHeight = chartHeight * .3;
```
unsoluble: Re the `open-` drawing height bugs: Was there a particular reason this was changed from `.2` to `.3`? Wondering if perhaps how it adds up here, plus padding, could be overflowing the height box, or something along those lines.

PieterGit:
@jpcunningh I just push my changes to my own branch and the `wip/013alpha` branch in such cases. It doesn't matter what branch they are. The 3 PR's don't have much merge issues. There is currently quite some people helping to test the `wip/013alpha` branch. Most people have troubles using PR or other peoples branches.

jpcunningh:
Yes, .2 was too small and not rendering correctly with D3 v5. With .2 the
entire context graph was too small. After seeing it was too small, and
adding up the numbers to see that it really should be too small when set to
.2, I changed it to .3 and the layout was corrected.

I wonder if there is a difference how the coordinates for the lines are
being interpreted between Windows and Linux vs. Mac...

Just for curiosity sake, any chance someone can see if it has the same
layout error on a Mac with Firefox?

I also might try translate instead of x, y coordinates to see if that
resolved the differences between the layout behaviors. I’ll let you know
when I push it if I can get translate to work on Windows and Linux.

unsoluble:
Exactly the same draw behaviour in Firefox/Mac. It's definitely odd that it's platform- and not browser-specific; that's very rare these days. (Also it's not a macOS Catalina thing — same results on a few older OS tests, too.)

unsoluble:
Also, not sure if this has been noted yet or not, but: Same draw errors on iOS, too. Starting to look like Windows might be the odd one out here. :)

jpcunningh:
Interesting... I’m on iOS 12.4 using Firefox and Safari, but haven’t seen
any draw issues.

unsoluble:
Oh well now we're into interesting territory. Maybe it's a config difference, in terms of displayed elements or something? Try trading URLs? Mine's at https://ns-testing.herokuapp.com/ — see how that looks on your end.

jpcunningh:
My site is locked down and I’m on my phone

I can see the issue you see, though! It’s a configuration difference! I bet
it is due to mmol vs mg/dL. I know what to look for when I get back to a
computer.

Thanks!

PieterGit:
@jpcunningh Thanks for fixing the reports.  I also use mmol and can't see any blockers with a quick scan of the latest version

@jpcunningh @unsoluble Can I merge it to dev soon, or do you see any blocking issues. We can fix potential minor D3 issues with future PR's.

unsoluble:
I have tried setting mine to mg/dL (both in the settings pane and in `DISPLAY_UNITS`, and got the same results. So it might not be that, but it’s definitely something along those lines.

unsoluble:
Another minor bug: On iOS, if you tap-and-hold somewhere on the screen that isn’t the draggable area (typically by accident), it’ll select some text (from an axis or something), but then after this point the context port becomes un-draggable until a full reload. (This is a regression from `master`.)

unsoluble:
Ah, found it: Renders properly when Scale is set to Logarithmic. Draws slightly too high on Log Dynamic, and way too high on Linear. I’m on mobile so can’t check code right now, but that’s where to look for the line-drawing one.

jpcunningh:
@unsoluble, I think I found it... can you update and test it?

unsoluble:
Nailed it, at least on iOS here. Will confirm on other platforms later tonight.

jpcunningh:
I finally got the click anywhere in on the bottom graph to scroll working.

This is the list of items remaining that I'm aware of:

   1. The forecast seems to be further out than the logic should allow (no
   more than half of the top graph should be forecast.
   2. The fonts are smaller than they used to be for the axes and values on
   the chart.
   3. Get the graph to scale correctly from the get go.
   4. Treatment reports appear to be stuck on "rendering".
   5. Loop / pump pills disappear when BG data went stale.
   6. More scrolling performance improvements.
   7. Refactor renderer.

#6 and #7 could probably wait for a follow on PR.

@PieterGit, I'm OK with merging to `dev` if you think the items above are
not show stoppers for the merge to `dev`.  I think I would rather fix #4
first, though.  I should be able to get to #4 Sunday.

PieterGit:
@jpcunningh I think we should fix #4 and #5 and we should know that there are no essential stuff missing for people. The treatments report of #4 worked for me yesterday (Chrome, Windows) and I checked the @Kdisimone page for #5 yesterday and here site worked for me on Chrome Windows, so perhaps it's only an issue on ios. 

We can also merge to dev later, until we a sure dev is able to be used for daily usage.

Nightfoxy:
@jpcunningh I tried to tackle #2 as an excuse to figure out how the code is laid out, but my web coding experience peaked at some basic bootstrap stuff. I couldn't quite figure out where the code is to be fixed.

I compared the rendered HTML to a v 0.12.x site I have and noticed this version has some inline styles for font size of 10 and sans serif declared. The CSS is the same and results in a 20 px font for the axis labels of the inline is removed, like in v 0.12. 

I'd love to know where the fix is if someone finds where to remove the inline styles.

bartsluimer:
Running latest on OSX 10.15 and IOS. Chrome and Safari both same experience. 
#1 Only half page is forecast
#2 Fonts are smaller yes. Not a big issue for me. Maybe even better for the overview.
#3 Graph does require a moment to scale
#4 Treatments rendering work correctly
#5 Haven't had the Loop disappearing yet.

What I did notice is in Dutch the reports page the different items require 2 lines. Half of the options are in front of the other. In English no issue. 
Schermafbeelding 2019-10-20 om 10 03 07

sulkaharo:
Tiny pointer in this thread in case someone doesn’t know: check CONTRIBUTING.md for instructions on how to run Nighscout in dev mode so it automatically detects code changes and packages the client without the need to manually restart/re-run any scripts.

sulkaharo:
PRd https://github.com/jpcunningh/cgm-remote-monitor/pull/2 that fixes the font size and fixes the z-ordering of the ticks so they render fully

sulkaharo:
Oops found a significant performance regression and fixed it: https://github.com/nightscout/cgm-remote-monitor/pull/5123 Once this is in dev, it might be good to merge dev here for a huge boost in the page load speed

unsoluble:
@bartsluimer Number of lines there is just a matter of window width. True that they shouldn't occlude each other when stacking, but this isn't a regression from `master`.

@jpcunningh Click anywhere to scroll works great on iOS here, but all scrolling is now super janky in Chrome/macOS — lots of pauses, lags, jumps. (Was butter smooth prior to this change.)

Cosmetic note: Any chance we could do the context graph transparency as a global overlay pass instead of per data point? When each point has partial transparency, they end up overlaying each other in ways that make the end result look blurry (and often essentially opaque anyway). Think if it's possible to just lay down a semi-trans black rect overtop of the whole context field (minus the focus area), would improve this.

unsoluble:
@jpcunningh Took some minutes to play around with the scrolling, to elaborate on "janky":

Behaviour is slightly different on iOS vs desktop, but this bit's the same on both: If you initiate the scroll from anywhere _outside_ the focus area, the scroll can move freely back and forth anywhere to the left of that point, but can't progress further to the right of where you started.

On iOS, initiating the scroll event from _within_ the focus area will let you freely move anywhere, no matter where it started (i.e. working as expected). On desktop, initiating from within focus doesn't do this, still works as per the previous paragraph (i.e. not as expected).

Let me know if you're not seeing any of this, I can do screen recordings or explain better.

PieterGit:
I've updated the wip/013alpha branch again, and included @sulkaharo fix for the COB plugin ( https://github.com/nightscout/cgm-remote-monitor/pull/5123 )

jpcunningh:
@unsoluble, thanks for the info! I think I have a fix for the jankiness. I saw some of what you were describing and found example of how to implement the "click anywhere" to begin scrolling that seems to be smoother than the what was there.

jpcunningh:
@sulkaharo, thanks for the font fix!

I can no longer replicate #4 above from @Kdisimone. I was able to make it happen yesterday evening, but haven't been able to make it happen again today.  If anybody is able to make Treatments hang at rendering and post a screen shot of the console, I would appreciate it!

jpcunningh:
This commit should fix the disappearing status pills @Kdisimone saw (assuming I was replicating the issue she saw).  It's also pushed to `wip/013alpha`.

Current Status:
   1. The forecast seems to be further out than the logic should allow (no more than half of the top graph should be forecast.
   2. The fonts are smaller than they used to be for the axes and values on the chart. -- Fixed
   3. Get the graph to scale correctly from the get go.
   4. Treatment reports appear to be stuck on "rendering".  -- Fixed (?) I haven't had it happen again in over 24 hours testing on Chrome/Firefox on Windows and Linux and Firefox/Safari on iOS.
   5. Loop / pump pills disappear when BG data went stale. --Fixed
   6. More scrolling performance improvements.
   7. Refactor renderer.

PieterGit:
Merged to `dev`. Please don't use `wip/013alpha` from now on, but switch to `dev` branch.
@jpcunningh Thanks for all the hard work.  All others: thanks for testing and doing small work.

Please file a new issue or PR for further improvements (polishing, etc).

PieterGit:
@jpcunningh Can you have a look with @sulkaharo on why the tests fail on travis. I will not busy the upcoming days, and will be back for Nightscout time friday or in the weekend.

Nightfoxy:
I leave NS open on a phone in our kitchen as well on my PC at work. The loop prediction (pink) seems to overlay the BG readings that come in unless I refresh. Then the pink starts eating the BG readings again. Android 7 running Chrome and Windows 10 running Chrome.
image

sulkaharo:
@Nightfoxy Yup looks like we're not clearing stale forecast dots correctly, haven't found the cause yet, looking for it

jpcunningh:
@Sulka, I have a thought on why the Loop predictions are eating the glucose
levels.  The time is used as a key to identify which circles are which, and
the circle's format is only set on the initial creation of the circle.

I'll push a branch to test shortly that separates the two circle types and
manages them separately so they can have exactly the same times between the
two and not have the problem of one eating the other.

jpcunningh:
@Nightfoxy, can you test the `forecast-eating-entries` branch on PR #5145?

PR #5102, opened 2019-10-16
fix: Support a wider range of unit strings for profiles

### LINKED ISSUE #5101: Basal pill sometimes converts mmol to mmol
Some users are reporting that after #4923 was merged, the basal pill is showing mmol value that's divided by 18. Looking at the recent code change that was made on this code, this makes me wonder if the uploaded profiles sometimes contain an unexpected units string, so rather than just checking string equality, we should do something like what Nightscout does on start and assume that any string that contains 'mmol' after lowercasing == mmol and any other string is mg/dl.

Ping @Erin879

sulkaharo:
Fixes #5101

### DIFF
--- a/lib/client/renderer.js
+++ b/lib/client/renderer.js
@@ -588,7 +588,7 @@ function init (client, d3) {
 
     function treatmentTooltip () {
       var glucose = treatment.glucose;
-      if (client.settings.units != client.ddata.profile.data[0].units) {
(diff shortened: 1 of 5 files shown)


sulkaharo:
@PieterGit can you find additional units checks that should just use the API in profile?

PieterGit:
This LGTM. Will do a search for more places, but at least let's merge this to dev for broader testing.

PR #5123, opened 2019-10-20
fix: Fix value caching in the profile functions. 

sulkaharo:
Looks like when Circadian profile support was added, the caching of value fetches broke, causing the calls to get profile data to be > 100x slower than before. Additionally, this changes the COB plugin to not calculate COB locally at all if a recent value is available from status uploads. When profiling the script execution speed using Chrome, the changes in this PR drop the script execution time of a full page reload from 3000ms to 330ms.


PieterGit:
LGTM. Merging to dev and also pushing to `wip/013alpha` branch

PR #5127, opened 2019-10-21
Fix tests failing in dev

sulkaharo:
* Fix reporting test to match the new output
* Fix number formatter to return the expected 0 when data is NaN
* Remove excessive logging during test runs from cob plugin


sulkaharo:
@aredridel Not sure why, but looks like tap-set-exit behaves differently locally vs on Travis, where in Travis we get exit code 2 when all tests succeed and the build is reported as failed.. :(

aredridel:
So do the tests properly fail in CI if they fail?

sulkaharo:
Not sure, I need to add a test that fails :D

aredridel:
Do! It's a bit terrifying to have potentially health-critical software that has a CI system that doesn't fail on failure. 😬

PR #5140, opened 2019-10-22
D3 Upgrade Residual Cleanup

jpcunningh:
This PR is in work with the following things to do:

- [x] fix forecast limit logic
- [x] fix minimum forecast time logic
- [ ] fix graph to scale correctly from the get go
- [ ] adjust comp z-index of focus y axis scale to be above `open-top`


unsoluble:
Tiny thing, but: Would it be possible to lift the main y axis scale up in the comp z-index, so that the lowest mark doesn't get partially obscured by the `open-top` bar?

Screen Shot 2019-10-22 at 10 52 52 AM

jpcunningh:
Will create a new PR for the 2 outstanding tasks.

PR #5145, opened 2019-10-24
Stop Focus Circles from Eating Entries

jpcunningh:


Nightfoxy:
Doesn't seem to have fixed the issue.
If it's any help, I didn't experience this in Pete's remote overrides branch for NS and Loop. 

I dont remember noticing it when I updated to Loop Dev after the remote overrides were merged in.

I thought it was from this alpha NS testing but someone on FB posted the same issue running NS 0.12.6-DEV and Loop Dev. 
https://m.facebook.com/groups/1782449781971680?view=permalink&id=2391723521044300

Cleared cache on W10 Chrome and refreshed on Android Chrome.

Could the issue be in Loop? Or introduced in NS 0.12.6-DEV? Will that help narrow down what to look at?

Nightfoxy:
@sulkaharo works perfectly now! Pink doesn't overlap the BG readings unless the are supposed to (loop hasn't updated but BG has). Nicely done.

Issue #5034, opened 2019-09-29
Extra info (delta and last measurement time) on clock view.


ireneusz-ptak:
The issue I'm trying to solve is that NS views are too complicated (graph) or too simple (clock views) for people that take care of our kids at school/kindergarten.

Two fields I'm currently missing are the delta and last measurement time (despite the stale status).

I fixed the problem by adding 2 parameters to the clock view URL (showDelta and showLastTime) so the features could be disabled for people using watches (we're using a tablet). It's working for me right now but maybe it makes sense to include it in the official NS release?

My changes are available here if you would like to take a look: https://github.com/ireneusz-ptak/cgm-remote-monitor/commit/07a0752f1b00b8f76291b3c2fd50cdf4606ad4b6

If you think it makes sense, I'm happy to contribute.

![image](https://user-images.githubusercontent.com/31506973/65832563-b2572100-e2c5-11e9-93eb-318c5ec13cd2.png)

unsoluble:
If you could also include user parameters for this so that those elements would default to off but be enabled with config vars, I'd back the PR. The current design of the color clock in particular was determined through extensive preference collection from NS users; the vast majority wanted it to only display the value and arrow, so that's what it's got.

Ultimately I'd love to see a fully modular clock system, where anyone can easily pick and choose all the elements they'd like to have on either of the two BG schemes.

ireneusz-ptak:
@unsoluble https://github.com/nightscout/cgm-remote-monitor/pull/5625 Before it is buried on PR list.

PR #5151, opened 2019-10-24
Extra info (delta and last measurement time) on clock view. 

ireneusz-ptak:
Idea described here:
https://github.com/nightscout/cgm-remote-monitor/issues/5034

PR adds 2 new parameters for "Color" view:
`SHOW_CLOCK_DELTA` to show BG change in the last 5 minutes, 
`SHOW_CLOCK_LAST_TIME` to always show BG age.


unsoluble:
I'll test when I've got a chance, but looks good in principle.

What I'd love to see (and have been aiming to do, just haven't gotten around to it) is getting us down to just _one_ actual clock view page, with all modules (color vs b&w, arrow, time, delta, and time-since) set via an in-site UI (as opposed to via config vars, so that you can run different viewsets in different browser instances).

But this'll be a decent midway step for now. :)

ireneusz-ptak:
One step at a time ;)

Nightfoxy:
Love this. I implemented something similar before 0.12.4 and also going to looping and haven't made time to do it again. Thanks!

PR #5170, opened 2019-10-30
Fix falsely suspended timeago alarms

fibbers:
This PR should fix that `timeago`-alerts aren't sent, because Hibernation was detected which silenced the alarms. At least, it does/did in my case 😄 (deployed this code to Heroku and verified the logs and tested the behaviour using Pushover)

When I looked at the logs in Papertrail, I saw a log-line each minute saying hibernation was detected, even though the app was still running fine in Heroku, preventing the alarm notifications when data isn't received for >15 minutes (or >30).

Looking at the code, I saw that the `delta` value checked for being > 15 seconds (which is always true). In Papertrail I saw the hibernation log-line each minute, so I searched for something like `setInterval` each minute. Eventually, I found `settings.heartbeat` which is 60secs by default (and matches the periodic log-lines), so I made the `delta`-check dependent on this value.

I tried adding a test and this is the best I could think of. I first tested with `timeago.checkNotifications(sbx)`, but found that when my test failed, other tests failed as well, so I switched to just testing `timeago.checkStatus(sbx)`. If there are better or more reliable ways, let me know.
All tests in `timeago.test.js` pass though on my machine, consistently now.

A more real-world test:
When I deployed this, I found that when I force-killed Spike, I received the `timeago` alerts again after 15 minutes (via Pushover).

Lastly, I want to express my gratitude for this project. I've been using it since a short while and I have nothing but good things to say. I'm impressed by the possibilities and integrations the project provides and I like it more and more every day, so thanks to all that have made this project possible!! 🥇 👍 👍

### DIFF
--- a/lib/plugins/timeago.js
+++ b/lib/plugins/timeago.js
@@ -3,10 +3,10 @@
 var levels = require('../levels');
 var times = require('../times');
 var lastChecked = new Date();
-var lastSuspendTime = new Date("1900-01-01");
 
 function init(ctx) {
   var translate = ctx.language.translate;
+  var heartbeatMs = ctx.settings.heartbeat * 1000;
 
   var timeago = {
     name: 'timeago',
@@ -64,21 +64,14 @@ function init(ctx) {
   };
 
   timeago.checkStatus = function checkStatus(sbx) {
-
     // Check if the app has been suspended; if yes, snooze data missing alarmn for 15 seconds
     var now = new Date();
     var delta = now.getTime() - lastChecked.getTime();
     lastChecked = now;
 
-    if (delta > 15 * 1000) { // Looks like we've been hibernating
-      lastSuspendTime = now;
-    }
-
-    var timeSinceLastSuspended = now.getTime() - lastSuspendTime.getTime();
-
-    if (timeSinceLastSuspended < (10 * 1000)) {
-
-      console.log('Hibernation detected, suspending timeago alarm');
+    // More than 2 heartbeats since last call: looks like we've been hibernating
+    if (delta > 2 * heartbeatMs) {
+      console.log('Hibernation detected because of 2 missed hearbeats, suspending timeago alarm. heartbeatMs:', heartbeatMs, 'delta:', delta);
       return 'current';
     }
 
--- a/lib/plugins/timeago.js
+++ b/lib/plugins/timeago.js
@@ -3,6 +3,7 @@
 var levels = require('../levels');
 var times = require('../times');
 var lastChecked = new Date();
+var lastSuspendTime = new Date("1900-01-01");
 
 function init(ctx) {
   var translate = ctx.language.translate;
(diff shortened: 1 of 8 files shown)


sulkaharo:
Yay thanks! Did you stumble onto the commit that broke this, by chance? Sounds like I maybe broke the server side when I fixed the client issuing false alarms (where the code is expected to not issue an alarm on the client side if the client has been suspended and hasn’t been loading data).

fibbers:
Hey, yep I saw 2 commits indeed. Wasn't sure whether it was about suspending the client or if the server on Heroku got suspended. You might probably make more sense out of it.
The commits are https://github.com/nightscout/cgm-remote-monitor/commit/3e4b45b2de1c8c34e2b111c0427492d49a83c56f and https://github.com/nightscout/cgm-remote-monitor/commit/7ae192d6a89801780727c3e5361c7305b7f0ff78.

(Maybe there's a different kind of heartbeat on the client that might be taken into account?)

sulkaharo:
Ya I need to make this run differently on the server and the client. On client, the issue is browsers stop the execution of JS if the window is not visible and the alarm is falsely triggered immediately when the execution is resumed, so we need to suspend the alarm for ~10 seconds after the execution has resumed to give the client time to update the data to prevent false alarms.

fibbers:
Ah I see and I think I understand now why there was a seperate `lastSuspendTime` variable, I assume because `checkStatus` might be called more often than `checkNotifications`. My change will then cause the false alarms again in the client if the first call is regular `checkStatus` and a second call is via `checkNotifications`... hence the suspending (just explaining to myself here:)).

So on the server I assume `checkStatus` is being called once per 60 seconds (default heartbeat), so the `delta` would be divergent (hibernated) if it is `>60seconds` instead of the stricter `>15seconds`... what if that's the only thing being changed?
```javascript
// Only change this check:
//if (delta > 15 * 1000) { // Looks like we've been hibernating
if (delta > 2 * heartbeatMs) {
  lastSuspendTime = now;
}

var timeSinceLastSuspended = now.getTime() - lastSuspendTime.getTime();
if (timeSinceLastSuspended < (10 * 1000)) {
  console.log('Hibernation detected, suspending timeago alarm');
  return 'current';
}
```
Then on the client, the alarms are only suspended if it was hibernated for 2 minutes instead of 15 seconds. The suspension-time stays 10 seconds though. I think this would also work on the server.

And taking the `timeago` settings into account (default 15 and 30 minutes), there would only be a false alarm if a user set `alarmTimeagoWarnMins` to a really low value like 1 or 2 minutes... the code might also take that into account with `Math.min` maybe.

What do you think?

sulkaharo:
The sandbox object doesn't currently identify if it's the client or the server, but I thought we could something like
```
index 3acb9a19..4bcee3b4 100644
--- a/lib/sandbox.js
+++ b/lib/sandbox.js
@@ -44,6 +44,7 @@ function init () {
   sbx.serverInit = function serverInit (env, ctx) {
     reset();
 
+    sbx.runtimeEnvironment = 'server';
     sbx.time = Date.now();
     sbx.settings = env.settings;
     sbx.data = ctx.ddata.clone();
@@ -83,6 +84,7 @@ function init () {
   sbx.clientInit = function clientInit (ctx, time, data) {
     reset();
 
+    sbx.runtimeEnvironment = 'client';
     sbx.settings = ctx.settings;
     sbx.showPlugins = ctx.settings.showPlugins;
     sbx.time = time;
```
and then have the alarms detect the runtime env and act accordingly

fibbers:
Ah, getting more familiar with the source now :-)
I modified the sandbox like you suggested and detect which environment is active and my test still passes. I find the github diff a bit hard to read, but what do you think about this?
The logic for the client should stay the same.

sulkaharo:
I’m looking into this and looks to me like the real source for no alarms in the server is the alarm enabling is broken for all plugins that uses the extended settings. Still digging into this

sulkaharo:
I get the following error when I try to run the test?

```
/Users/sulka/dev/cgm-remote-monitor/lib/plugins/timeago.js:10
  var heartbeatMs = ctx.settings.heartbeat * 1000;
                                 ^

TypeError: Cannot read property 'heartbeat' of undefined
    at init (/Users/sulka/dev/cgm-remote-monitor/lib/plugins/timeago.js:10:34)
```

fibbers:
My bad, I was testing it in isolation (`env-cmd ./test.env mocha --exit tests/timeago.test.js`).
I modified the tests so `npm test` should succeed again.

I'm not sure if I'm working in the right direction with this, so do let me know if you think the solution should be found in a different place.

Issue #5190, opened 2019-11-08
"Glucose Distribution" report doesn't finish rendering without data


steve8x8:
**Describe the bug**
With only BG measurements available (CGM temporarily unavailable), rendering the "glucose distribution" view doesn't finish

**To Reproduce**
1. Run out of CGM sensors for a few days.
2. Select a time span short enough in Reports
3. Select "Show" or "Distribution"
4. Wait for "Rendering..." to disappear

**Expected behavior**
There should be a detection whether (enough) data are available at all.

**Screenshots**

**Your setup information**
- What version of Nightscout - 0.12.6-dev
- What type of CGM, and how do you get your data there? - currently none
- Is your issue specific to a browser (Firefox/Safari/Chrome?) or a device (Android phone, etc.)? - No.

**Additional context**
Any previously drawn pie diagram isn't erased while (not) "rendering..."

jpcunningh:
@steve8x8, this should also be fixed by #5195. Please let us know if it works for you.

steve8x8:
Unfortunately I cannot test it right away (you're referring to the 2nd of the 3 commits, I presume) - but lgtm. Go ahead...

steve8x8:
Update: although the server claims to be running dev 534c443, I cannot see any change :(

sulkaharo:
@steve8x8 are you self-hosting? If yes, remember to run the packaging script between installs

steve8x8:
I'm not self-hosting (yet) so I must believe what the interface and the server admin tell me...

Issue #5189, opened 2019-11-08
No scale if no CGM data, only manual BG measurements


steve8x8:
**Describe the bug**
If there are no CGM data, only BG measurements (e.g. uploaded by xDrip), there's no Y axis scale, and values cannot be read off.

**To Reproduce**
1. Run out of CGM sensors, wait until the graph is "empty" (few days)
2. Use manual BG measurements instead, upload using xDrip.
3. Check display: Red circles are there, but all along a single horizontal line (0.12.5) or two horizontal lines (0.12.6-dev git 9c42fed). No scale (except some blurry stuff in the top-right corner?), no mouse-over. Looks like division by zero?

**Expected behavior**
If there's no CGM value range, either the BG range or a default should be used.

**Screenshots**
https://ns.10be.de:6089

**Your setup information**
- What version of Nightscout - 0.12.6-dev
- What type of CGM, and how do you get your data there? - was using Libre with Glimp, switched to xDrip also for uploading Contour meter data
- Is your issue specific to a browser (Firefox/Safari/Chrome?) or a device (Android phone, etc.)? - No

**Additional context**
May be related to issue #3025

jpcunningh:
@steve8x8, can you test #5195 to see if it corrects the issue?

jpcunningh:
I created a separate branch `fix-missing-scale-when-no-entries` so you can deploy it to your server more easily than from the PR.

steve8x8:
I cannot test (as I'm not having admin powers on that server), but the first commit lgtm - in particular as I discovered that in "pure logarithmic scale" the overview graph is OK (but the detailed one has all the values at the left corner so that might be yet another bug?)...
The other two commits are not related to this issue...

steve8x8:
OK, since the server instance has been upgraded to dev 534c443 I can now report...
... that using a logarithmic scale (non-dynamic) places all markers (BG readings, boluses, pump maintenance actions, etc.) at the left border of the main canvas while the overview graph shows the BG measurements of the last days.
... that switching to linear results in everything still in the top left corner (and no useful scale), so basically the behaviour is unchanged.
The URL is still the same.
Apparently also the main graph X axis needs some more attention...
Thanks so far -

jpcunningh:
Thanks! I’ll be able to check it out in a couple of hours when I get to a
computer.

jpcunningh:
Looks like your server is not running 5434c443.  One of the relevant lines in the browser developer tools look like this:
[...]

PR #5195, opened 2019-11-10
Don't use dynamic scale unless entries in database

jpcunningh:


jpcunningh:
@sulka, if you can do a code review, I think this is ready to merge.  I tested it on a local VM with no current data to address issues #5189 and #5190.

PR #5200, opened 2019-11-12
Ice

friedl2:


jpcunningh:
Hi! It seems this pull request was made in the wrong direction - you're trying to pull changes from your copy of Nightscout into the master version.  The correct direction is from master into your copy.

The correct direction for your update PR is created by this link: [https://github.com/friedl2/cgm-remote-monitor/compare/master...nightscout:master](url)

However, looking at your Nightscout fork, it is currently up to date with the master due to your successful PR at [https://github.com/friedl2/cgm-remote-monitor/pull/8](url), so an update is no longer required at this time.

PR #5218, opened 2019-11-18
Update

lukebrogan:
Update nightscout app


jpcunningh:
Hi! It seems this pull request was made in the wrong direction - you're trying to pull changes from your copy of Nightscout into the master version.  The correct direction is from master into your copy.

Looking at your fork, your master is current with the Nightscout master, so an update is not necessary at this time. For future updates, the correct direction for your update PR can be created by this link: https://github.com/lukebrogan/cgm-remote-monitor/compare/master...nightscout:master

PR #5236, opened 2019-11-23
D3 cleanup

jpcunningh:
* Trims the open-top rectangle to not extend into the focus area y axis.
* Handles clearing the timeago alarm styles a little better
* Don't set the direction if the SGV is not current

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2272,7 +2272,7 @@
     },
     "browserify-aes": {
       "version": "1.2.0",
-      "resolved": "http://registry.npmjs.org/browserify-aes/-/browserify-aes-1.2.0.tgz",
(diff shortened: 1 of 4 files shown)


npm-shrinkwrap.json
```diff
@@ -2272,7 +2272,7 @@
     },
     "browserify-aes": {
       "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/browserify-aes/-/browserify-aes-1.2.0.tgz",
```
jakobsandberg: Were these changes to `npm-shrinkwrap.json` intentional? Most of them point us to the npm registry via HTTP instead of HTTPS. I think we should remove them from the PR.

jpcunningh: Thanks, @jakobsandberg. I rolled them back as you suggested.

PR #5286, opened 2019-12-13
[Clock] Fix incorrect delta calculation

ireneusz-ptak:
Nothing better than example:
Last 3 BG values: 90->100->100 - delta was incorrectly calculated showing -5 (difference between 1st and 3rd value normalized to 5 min. period), it should be 0 instead.

Caused by incorrect !delta condition which returns true for delta=0.
Changed to string, works nicely now (including mmol conversion).

### DIFF
--- a/lib/client/clock-client.js
+++ b/lib/client/clock-client.js
@@ -44,7 +44,7 @@ client.render = function render (xhr) {
       rec = element;
     }
     else if (element.sgv && rec && !delta) {
-      delta = (rec.sgv - element.sgv)/((rec.date - element.date)/(5*60*1000));
+      delta = ((rec.sgv - element.sgv)/((rec.date - element.date)/(5*60*1000))).toFixed(0);
     }
   });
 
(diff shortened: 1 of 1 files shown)


lib/client/clock-client.js
```diff
@@ -44,7 +44,7 @@ client.render = function render (xhr) {
       rec = element;
     }
     else if (element.sgv && rec && !delta) {
-      delta = (rec.sgv - element.sgv)/((rec.date - element.date)/(5*60*1000));
+      delta = ((rec.sgv - element.sgv)/((rec.date - element.date)/(5*60*1000))).toFixed(0);
```
sulkaharo: toFixed returns a string. This should be cast back to a number, or you should use some other method to round the digits to produce a number

sulkaharo: Oops saved half-written comment - meant to say we should fix the !delta comparison and not rely on the number being cast to a string, as this is inherently fragile.

ireneusz-ptak: As a DB developer I'm allergic to null comparisons, but it should work.
Also removed the !delta condition from the previous if statement as it's not needed there.

sulkaharo:
👍

PR #5314, opened 2019-12-17
Fix auth dialog sizing error

jpcunningh:


jpcunningh:
I've tested on Android Firefox, iPhone FireFox, iPhone Safari, Windows 10 Chrome, Windows 10 Firefox, and Windows 10 Edge.

However, I found a latent bug that prevented the client from loading after a successful authentication. The correction is pushed to this PR now.

PR #5377, opened 2019-12-27
Fix Alexa Launch and SessionEnded Requests

Bartlomiejsz:
This fixes Alexa action when `Alexa, launch Nightscout` command is used and same later on calling `exit`.

Previously such command resulted in Alexa saying that there was issue with skill response caused by 500 error thrown by Nightscout API. Error itself was caused by API trying to find intent in request, but as in [Alexa documentation](https://developer.amazon.com/en-US/docs/alexa/custom-skills/request-types-reference.html#launchrequest) there is no `intent` field in LaunchRequest body.

This PR fixes this request to be correctly handled and includes adding open session to wait for next command.

For `exit` there was no 500 error because Alexa session is closed anyway, just this situation was incorrectly handled (response was not sent at all until timeout error).

Also test for LaunchRequest was added.


inventor96:
@Bartlomiejsz Previous to the changes released in v13, the code looked exactly as you have changed it in this PR. During my testing of those changes, I found that there were occasions where the intent was included in the `LaunchRequest` requests (even though the docs say it's not supposed to have it), and so I modified it to handle requests with the intent. Obviously I didn't encounter all situations, because I don't recall getting the error that you've uncovered.

Could you modify your PR so that it can handle requests with and without an intent parameter?

Bartlomiejsz:
Sure @inventor96, I'll work on that. Got small question, is there any specific reason why shouldEndSession is being set as string instead of boolean?

inventor96:
I'm not sure on that one. It was like that before I worked on this stuff, and nothing seemed broken because of it, so I just left it alone.

Bartlomiejsz:
Ok, I'll verify if it will work on Alexa device and change it to boolean if possible, in previous functionality `'true'` was always casted to `true`, but `'false'` in some situations might also be casted to `true`

Bartlomiejsz:
Tested @inventor96, works with boolean values 👍

Bartlomiejsz:
@inventor96 I added also fix for SessionEndedRequest. Previously it was never called because session was ended after each intent, but now with launch fixed it's possible to use `exit` to stop session.

PR #5381, opened 2019-12-28
Fix #5146 - display of treatment with only Fat and Protein set

### LINKED ISSUE #5146: Fat & Protein Display
Recently the feature to log Protein and Fat intake in g has been added. Great feature and the first step to may be move to an insulination model not purely based on carb or carb equivalent. However events in Nightscout do not display any of those input. Mongo DB has them though.

Any treatment containing input for those 2 variables should display them on NS.

Bartlomiejsz:
This is fix for #5146 - previously when Carb Correction was added without Carb (with only Fat/Protein or those both) it was visible differently. This PR fixes behavior of such treatment.

Also in tooltip for treatments BG was incorrectly calculated when set to mg/dl because of mismatch in https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/client/renderer.js#L635 - `mg/dl` being compared to `mgdl` returned false so was treated as `mmol` - this was also fixed.

### DIFF
--- a/lib/profilefunctions.js
+++ b/lib/profilefunctions.js
@@ -163,7 +163,7 @@ function init (profileData) {
   profile.getUnits = function getUnits (spec_profile) {
     var pu = profile.getCurrentProfile(null, spec_profile)['units'] + ' ';
     if (pu.toLowerCase().includes('mmol')) return 'mmol';
-    return 'mgdl';
+    return 'mg/dl';
   };
 
   profile.getTimezone = function getTimezone (spec_profile) {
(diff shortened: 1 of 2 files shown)


lib/profilefunctions.js
```diff
@@ -163,7 +163,7 @@ function init (profileData) {
   profile.getUnits = function getUnits (spec_profile) {
     var pu = profile.getCurrentProfile(null, spec_profile)['units'] + ' ';
     if (pu.toLowerCase().includes('mmol')) return 'mmol';
-    return 'mgdl';
+    return 'mg/dl';
```
sulkaharo: This might have consequences for systems using Nightscout data, but let's change it anyway

Bartlomiejsz:
@sulkaharo I only found few places in NS using this function and in each one of those same issue occurred, and changing it to mg/dl fixes it. So I don't think there should be any consequences apart from fixing those issues

lib/profilefunctions.js
Bartlomiejsz: @sulkaharo I only found few places in NS using this function and in each one of those same issue occurred, and changing it to mg/dl fixes it. So I don't think there should be any consequences apart from fixing those issues

sulkaharo:
@Bartlomiejsz Yup this will work in Nightscout, but the value also propagates to the profiles which are accessed through the API, which might in turn have consequences for external code accessing the data.

Bartlomiejsz:
@sulkaharo ahh, ok, thank you for clarification

PR #5393, opened 2020-01-01
Move toolbar and authentication status to partial and include on each page

Bartlomiejsz:
This PR creates reusable authentication-status and toolbar partials that is included on each page and styled in consistent way.

Preview of changes:
image
image

Header and authentication status is consistent on each page


sulkaharo:
This changes the index page layout on mobile - the menus switch from the right to left side

IMG_1020

sulkaharo:
(Looks really good, other than that change)

Bartlomiejsz:
@sulkaharo oh, didn't notice that, I'll fix that then :)

Bartlomiejsz:
@sulkaharo could you please let me know on what device is that? And are you sure those are not old styled cached?
image

sulkaharo:
🤔 Yup can't repro now. I did clear the cache for the previous test, but guess something went wrong with that. This would be an iPhone 11.

PR #5410, opened 2020-01-03
Add server teardown ability

PetrOndrusek:
Right now, when the Nightscout server is started, a number of timers remain hooked up and remain active during the tests. 
The purpose of this PR is to prepare a valid server shutdown by calling `ctx.bus.teardown`, which causes a `teardown` event to be sent into the bus. 
Any place in code that needs to clean up something when shutting down the server can now start listening to the `teardown` event from the bus.

### DIFF
--- a/lib/bus.js
+++ b/lib/bus.js
@@ -1,11 +1,11 @@
 'use strict';
-
 var Stream = require('stream');
 
 function init (settings) {
   var beats = 0;
   var started = new Date( );
   var interval = settings.heartbeat * 1000;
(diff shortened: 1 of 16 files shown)


sulkaharo:
😍

sulkaharo:
(Looks like this is based on a bit old version of dev, please update to the latest dev before issuing a new PR as that'll make testing a lot easier.)

PetrOndrusek:
Ok, I'll be careful about that next time, thanks.

PR #5419, opened 2020-01-05
update

Maddiecgm:


sulkaharo:

Maddiecgm:
Thank you, I will work on that tonight, I don’t use github or heroku often I tried the update tool and I had two errors that had to be resolved.

PR #5425, opened 2020-01-07
Api3 output renderers

PetrOndrusek:
This PR offers the possibility to use CSV or XML as an output content type in APIv3. 
CSV/XML output can be requested either by adding `.csv` or `.xml` suffix in URL or sending Accept HTTP request header (`text/csv` or `application/xml`).
Both new output renderers use lightweight npm packages (csv-stringify and easyxml) in order to minimize the impact on codebase growth. PR includes appropriate tests and documentation changes.


jasoncalabrese:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- lib/api3/shared/renderer.js  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Nightscout/cgm-remote-monitor/pullRequest?prid=4779305&bid=15921994)

lib/api3/swagger.js
```diff
@@ -10,7 +10,7 @@ function setupSwaggerUI (app) {
   const serveSwaggerDef = function serveSwaggerDef (req, res) {
     res.sendFile(__dirname + '/swagger.yaml');
   };
-  app.get('/swagger.yaml', serveSwaggerDef);
+  app.get('/swagger', serveSwaggerDef);
```
jakobsandberg: This is a breaking change which will result in a 404 for users who have bookmarked this page or scripts that programmatically request this route.

Do we need to change this? If so, what do you think about supporting both `/swagger.yaml` and `/swagger`?routes?

PetrOndrusek: You are right that it looks like `/swagger.yaml` URL has changed, but in fact it works both `/swagger.yaml`, `/swagger`. 
Actually `/swagger.xxx` returns JSON everytime, no matter which file extension you use.
Only READ, SEARCH, HISTORY operations (GETs for some collection) evaluate the `Accept`/extension, so it should be fine.

sulkaharo:
LGTM

PR #5442, opened 2020-01-11
Add sensor code and tx id fields to CGM Start treatment

c-robertson:
Added input fields to the treatment drawer to specify Sensor Code and Transmitter ID.  Updated Care Portal to show these fields for the CGM Start Sensor treatment and capture input values in specific properties on the treatment event.  The intent is to allow applications like Logger to deterministically locate and make use of the values entered when processing sensor starts.

### DIFF
--- a/lib/client/careportal.js
+++ b/lib/client/careportal.js
@@ -52,7 +52,7 @@ function init (client, $) {
     submitHooks = {};
 
     _.forEach(careportal.allEventTypes, function each (event) {
-      inputMatrix[event.val] = _.pick(event, ['bg', 'insulin', 'carbs', 'protein', 'fat', 'prebolus', 'duration', 'percent', 'absolute', 'profile', 'split', 'reasons', 'targets']);
+      inputMatrix[event.val] = _.pick(event, ['bg', 'insulin', 'carbs', 'protein', 'fat', 'prebolus', 'duration', 'percent', 'absolute', 'profile', 'split', 'sensor', 'reasons', 'targets']);
       submitHooks[event.val] = event.submitHook;
     });
   }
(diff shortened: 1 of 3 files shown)


efidoman:
LGTM - Once this merges, we can update Logger and Lookout to utilize these new fields.
@jpcunningh

sulkaharo:
What's the expected behaviour with these fields? I'm a bit hesitant to add fields to the codebase that would cause behaviour changes in external systems, without any logic changes in Nightscout itself

c-robertson:
As far as Nightscout is concerned, the intent is simply to capture sensor details in dedicated properties, rather than somewhat arbitrarily in the Notes field.  The immediate use case is for Logger to be able to conveniently capture Dexcom sensor changes, including those that entail pairing a new transmitter.  Any downstream app could be updated to use these fields in a backwards compatible way, falling back to the Notes field if the sensor specific properties are not found in an NS treatment.  From a user experience perspective, we could consider a radio selector to allow the user to specify their particular sensor type, with corresponding input fields for whatever info different systems capture (I am only aware of Dexcom G6 requirements...).

c-robertson:
Inadvertently closed...

inventor96 (approved):
I was just debating to myself whether I should do something similar. Glad someone else got to it first! The code looks good to me.

inventor96:
@c-robertson Could you rebase your branch on the latest release to resolve the merge conflicts?

sulkaharo:
To explain better why I was asking about what this is for: I _really_  would prefer to not merge in changes that don't provide a clear use case for end users immediately and don't include some developer docs, as there's almost no resources available for maintaining Nightscout and all code in the system adds to maintenance overhead. So if you want me to merge this, at least: 1) add the fields to the timeline user interface, so the data is shown to users if it's added and 2) add the fields to the swagger files so they're at least somehow documented.

sulkaharo:
Closing for now, please see above for the change request

PR #5454, opened 2020-01-14
Add mute button and move sound test button to settings section

### LINKED ISSUE #5451: Make sound button mute / un-mute all sounds
I assumed that the sound button (volume icon) would mute all sound. After inspecting the logic, I realized that the purpose of this button is to test the sound on your device (it simulates an alarm for 4 seconds). Once I tested the sound and verified it worked, this button no longer had a use for me. Given that this button takes up valuable real estate (next to primary navigation buttons), it mostly just gets in my way.

I'm concerned that other users are confused by this as well. I'm also worried that new users will find the alarm sounds jarring and intuitively click the volume icon to silence it. This will lead to frustration when the button doesn't silence the alarm. (e.g. see #5349)

I think it would be very useful to have a mute / un-mute button in a visible place, especially when you are in a quiet environment (e.g. a library). We can include a new icon to make it explicit that the audio is muted, and toggle this icon with the existing one as the sound is muted / unmuted.

I still see merits to a button to test the sound feature, so how about we move that to the settings section next to the alarm settings?

What do y'all think? If there is interest, I'll open a PR for this change

jakobsandberg:
Resolves #5451


sulkaharo:
Just to confirm - which browsers / platforms did you test this on? I'm wondering if this will be confusing to users, as unmuted sound doesn't mean any sound will ever play. In terms of UI, what could maybe make more sense is having the muting in the settings tray (default to on, allow users to disable sound for this client) and making the audio testing button stateful (with no persistence) so users would see if they've pressed the button during the currently open browser session and thus enabled it.

PR #5499, opened 2020-02-01
Fix Brushing Loop

jpcunningh:


sulkaharo:
LGTM, tested after merging latest dev into this

PR #5512, opened 2020-02-09
client - mobile - Axis label font is scaled on small screens

stephencmorton:
On phones or other small screens, the axis labels (especially x axis)
were previously an unreadable jumble of too-close letters

Developer note: I do not like using `important` but it seemed impossible otherwise. I do not think that in this case I'm violating any important principle by doing so.

IMG_6940

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -516,6 +516,11 @@ a, a:visited, a:link {
 }
 
 @media (max-width: 750px) {
+  .x.axis {
+    font-size: 2.5vmin !important;
+  }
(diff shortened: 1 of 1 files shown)


sulkaharo:
Hi! Can you re-submit against dev? We merge all changes through dev, so they appear in the next release. The changes LGTM

stephencmorton:
@sulkaharo  Will do tonight.

stephencmorton:
Base adjusted to `dev`.

PR #5518, opened 2020-02-10
Split view

sulkaharo:
Adds a 2, 3, 4 and 8 way split view option


Tornado-Tim:
Good feature, but IFrames are not good for use in this case due to loading the DOM with far too many elements. In the case of Nightscout there is a huge amount of elements alone on one screen.
Not trying to prevent the PR from going ahead, but this will make the web UI slow, and use a huge amount of ram.

sulkaharo:
@Tornado-Tim This setup is actually used in the wild. 😅 See for example [this thread in Twitter](https://twitter.com/csebastian/status/1226269051921600518?s=20). The rationale for the iframe setup is simply to allow easy full-screen view of multiple sites (vs having many windows open at the same time). With the recent performance optimisation, my laptop comfortably runs 8 sites in one window without perceived slowdown (using Safari), so it doesn't look like performance is an issue.

PR #5519, opened 2020-02-10
Trying to fix random fail of APIv3 tests

PetrOndrusek:
There are some random failures of APIv3 tests during CI on Github (Github Actions). This is probably caused by clashes of computed identifier of test documents. These clashes probably arise in very fast (or parallel) test execution, when test documents get the same timestamp.
This PR tries to solve this by isolating test documents, which is achieved by using different device name in different test file.


sulkaharo:
LGTM. And tests pass on this one 😅

Issue #5514, opened 2020-02-10
Option to switch off bolus amount outputs


andrew-dixon:
**Is your feature request related to a problem? Please describe.**
With the new "automatic bolus" feature currently available to test in Loop, the number of micro boluses can make the chart in NS very hard to read, as shown in this screenshot.

<img width="618" alt="Screenshot 2020-02-10 at 07 52 02" src="https://user-images.githubusercontent.com/7441801/74131311-f3e4ca80-4bdb-11ea-94fc-b46c03d112e4.png">

**Describe the solution you'd like**
Have an option to switch off the output of bolus values on the charge so you only see the bolus circle and have to hover over it to see the bolus amount.

**Additional context**
Relates to: 
- https://github.com/LoopKit/Loop/tree/automatic-bolus
- https://github.com/LoopKit/Loop/pull/1219

sulkaharo:
Which version of Nightscout do you have installed? We changed the rendering quite long time ago to help with this issue for OpenAPS SMB users (so wondering if you have an old build or whether the alternate rendering is only used for OpenAPS data).

andrew-dixon:
@sulkaharo Version 13.0.1, which I believe is the latest, so maybe that is the case, I wasn't aware of that. I'll see if I can track down the issue and take a look in the code.

andrew-dixon:
@sulkaharo I think it might be this line:

https://github.com/nightscout/cgm-remote-monitor/blob/41ac93f7217b1b7023ec6ad6fc35d29dcf2e4f88/lib/client/renderer.js#L572

Which is only for OpenAPS and SMB, so I will look into if there is a way we can use this for Loop AB as well, e.g. what conditional logic can be added to detect Loop AB.

sulkaharo:
Actually, instead of adding new logic that tries to detect a specific Nightscout client, I think I'd prefer removing the OpenAPS reference and adding a new generic setting that defines a threshold for a bolus size, under which the boluses are rendered in smaller size. This would be future proof and satisfy all users regardless of the system they use.

andrew-dixon:
That makes sense, I'll see if I can find time to look at that this week.

andrew-dixon:
@sulkaharo What do you think about this as an option in the menu?

<img width="237" alt="Screenshot 2020-02-11 at 23 20 57" src="https://user-images.githubusercontent.com/7441801/74288808-4a115500-4d25-11ea-81cc-2a129675a011.png">
[...]

PR #5522, opened 2020-02-12
Option to switch off bolus amount outputs

andrew-dixon:
Reference to #5514

Adds a new settings menu item "Render Bolus Amount" with 5 options:

- All
- All (without leading zero)
- All (just the number)
- 1 U and Over
- None

These options change how the bolus amount is output on the chart. 

Also removed code specifically for OpenAPS SMB as this would no longer be required as the user could select an option from the list.

Also removed an erroneous piece of code on line 573 in `renderer.js` that was setting a variable called `enteredBy` which is never used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,7 @@ To learn more about the Nightscout API, visit https://YOUR-SITE.com/api-docs/ or
     * The `linear` option has equidistant tick marks; the range used is dynamic so that space at the top of chart isn't wasted.
     * The `log-dynamic` is similar to the default `log` options, but uses the same dynamic range and the `linear` scale.
   * `EDIT_MODE` (`on`) - possible values `on` or `off`. Enables the icon allowing for editing of treatments in the main view.
+  * `BOLUS_RENDER_OVER` (1) - U value over which the bolus values are rendered on the chart if the 'x U and Over' option is selected. This value can be an integer or a float, e.g. 0.3, 1.5, 2, etc...
 
 ### Predefined values for your server settings (optional)
   * `INSECURE_USE_HTTP` (`false`) - Redirect unsafe http traffic to https. Possible values `false`, or `true`. Your site redirects to `https` by default. If you don't want that from Nightscout, but want to implement that with a Nginx or Apache proxy, set `INSECURE_USE_HTTP` to `true`. Note: This will allow (unsafe) http traffic to your Nightscout instance and is not recommended.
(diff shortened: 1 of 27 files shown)


sulkaharo:
Wonder if the 1U threshold should be configurable / is the right threshold. Nightscout is primarily used to monitor kids, for whom SMBs are all very small (I'd assume max 0.4 IU), so being able to set the threshold somewhere closer to the SMB limit could make sense?

andrew-dixon:
@sulkaharo I set it to 1U and over because that was what it was in the hardcoded OpenAPS SMB code that was already there. I'll take a look at making it configurable. Is there a preference with this sort of thing to where the configuration option should be, e.g. in the interface or in the environment vars? Just thinking it might be easier in the environment vars and also not clutter up the interface more.

andrew-dixon:
@sulkaharo I've added this as an environment variable, BOLUS_RENDER_OVER, with a default value of 1, to match the current OpenAPS SMB implementation but can be set to any value, e.g. 0.3, 1.5, 2, etc... Added description to the `README.md` file as well.

sulkaharo:
Sorry took a while - was doing final comparison between this and release and realised none of the options here match the current behaviour in the master release. If we change behaviour and make it configurable, it'd probably be good to default to a config option that matches the current behaviour, or at least include such option. Second, what'd possibly improve this quite a lot is a small additional CSS change so the boluses under the threshold used a bit smaller font, as that'd reduce the likelyhood of messy rendering a lot.

andrew-dixon:
@sulkaharo The "x U and Over" option, which defaults to 1 U if you don't set the environment variable, is the same as what it currently does for OpenAPS SMB, so that option is the same, except for it defaulting to that for OpenAPS SMB, which in the discussion on #5514 I thought you said you would like to see removed anyway, as having something hardcoded for a specific system isn't a particularly good idea.

With regards to the font size, I will look into that.

andrew-dixon:
@sulkaharo I've made some changes for the font size depending on the bolus amount, so the smaller the bolus amount, the smaller the font. This applies to any bolus under 1U. I've also resolved the merge conflict that was previously showing.

sulkaharo:
What I meant with the removal of OpenAPS code removal is, I'd prefer a solution that's not dependent on OpenAPS vs being generic.

When I enable "1U and over", the text labels below 1U are removed. This doesn't match the current behaviour - in master, the OpenAPS behaviour is that for <1U boluses the string is shortened to not have the leading zero and the U is dropped. With this PR, the options are to

1) Show everything,
2) Strip 0 but show the U,
3) Show the numbers without the U but with the 0,
4) Don't show the number / U at all for boluses under 1 IU,
5) Show nothing

The current Master strips the leading zero _and_ the U, so the dosing string is shortened as much as possible while still shown. I think the best way to have people not be surprised by the PR could be if we can merge the options 2 and 3 in the PR to behave like master (strip the leading 0 and the U), as neither of the options materially shorten the string.

Shot from current master:

Screenshot 2020-03-01 at 20 37 46

andrew-dixon:
Ok, got it, I'll make the requested adjustments.

sulkaharo:
@andrew-dixon if you have the time to finish this, I'd love to get this into the next release

andrew-dixon:
@sulkaharo Sorry for the delay, work has been a little manic. Will hopefully look at this later today.

andrew-dixon:
@sulkaharo I've made the requested change, it was actually a much smaller change to get it working than I expected.

PR #5526, opened 2020-02-16
Release 14.0.0

sulkaharo:


PieterGit:
@sulkaharo @jpcunningh I hope I will get more time for NS development than last months. Are you planning to release a 13.1.0 soon, or should we work on a 14.0.0 release as the next release.

If we drop Node v8 and v10 support and introduce Node v12 support, I think we should release the next release as 14.0.0 (because people need to upgrade their Node).  If we stick to our own policy of only maintaining Nightscout  with Active Node LTS we should release a new Nightscout with v12 support before 2020-04-30, since Node v10 maintaince is expected then. See https://nodejs.org/en/about/releases/

Thoughts?

Gracious-ann:
my dev

Gracious-ann:
hjgf

PR #5496, opened 2020-01-30
Plugin to show database size (% of available space or in MiB)

dlvoy:
This plugin shows pill with information about size of database,
can display both % of available (declared) max size, or absolute size in MiB

This is useful for long-time users on free tier mongoDB, to indicate when to clean-up DB and prevent issues when DB size quota is exceeded unnoticed.

When size reached default or configured warning/urgent threshold - it highlights yellow or red
Includes: icon (with sources), translation support, README entry & tests

dbsize_plugin_pill

for `DBSIZE_IN_MIB=true`

dbsize_pill_mib

PR #5529, opened 2020-02-17
[DEV][FIX] Fix dbsize plugin to make it hideable again

dlvoy:
Fix bug that forced dbsize plugin to be enabled and always checked on plugin list (= forced to be enabled).
After fix it is enabled by default, but user can uncheck it on plugin list to hide it.


andrew-dixon:
@dlvoy I've applied this patch to my local dev environment and can confirm the "Database Size" option now works, however, I'm still getting the lodash warning/error message in the console:

https://github.com/nightscout/cgm-remote-monitor/pull/5496/files/db0a3b3a076add2f4e7168ed296e816365e01ac7#diff-25d013e7c4f5605f78a97e7e3afa9865

dlvoy:
@andrew-dixon Thanks for that! Somehow i missed the warning, but `lodash` is unused dependency so i simply removed it :)

PR #5572, opened 2020-03-15
Fixes #4991 - Now selects basal profile based day by day

### LINKED ISSUE #4991: The  DayToDay Report only uses the last basal profile, not the actual profiles
**Describe the bug**
Where multiple basal profiles that apply over the period I want to report on - only one  basal profile is applied, and it is always the latest one

**To Reproduce**
Steps to reproduce the behaviour (in this example I am looking at one basal profile per day, starting on a Monday):
- Use Profile Editor to create a basal pattern with a valid from date time of Monday midnight, e.g. with a total basal of 6.0 units
- Then create a second profile, with valid from Tuesday midnight with a different basal total e.g. 11.0 units
- Create a third profile, e.g. for Friday midnight with a different basal, e.g. 4.0 units
- Run a "day to day" report from Sunday to Wednesday, and make sure "Insulin Distribution" is checked
- When the report is run the Friday night basal is the one shown on each days basal

**Expected behaviour**
Total basal units for Monday should be 6.0 units, Tuesday & Wednesday should be 11 units (unless temporary basals are in the dataset)

**Setup information**
- Nightscout (e.g. 0.12.3)
- CGM:  Dexcom G6 or Libre 
- Pump: Medtronic Mini-Med

**Proposed Solution**
Root cause appears to be that profile.getTempBasal() function does not pull back correct basal profile based on date & time
To fix that I have (so far) changed the following:
In lib\profilefunction.js:
- Added a new function profile.profileFromTime
- Changed to follow to use it and select the correct profile:  getCurrentProfile, activeProfileToTime, activeProfileTreatmentToTime

In lib\report-plugins\daytoday.js, I make changes to drawChart in the Basals data section to call 
profile.loadData to load the basal profiles, and the subsequent calls to profile.getTempBasal() return the correct basal profile.

fromorbonia:
Detailed list of changes:

In **lib\profilefunctions.js**:

- Added profile.profileFromTime

- Changed:
  - getCurrentProfile to use it
  - activeProfileToTime
  - activeProfileTreatmentToTime

In **static/report/js/report.js**:
- Add loadProfilesRange which loads only the profiles for the day range, and the profile(s) before and after based on the StartDate of the profile
- Changed loadProfileSwitch to use loadProfilesRange
- Note: loadProfilesRange uses a ajax $.when to chain three sync API calls - this doesn't feel right, but unless the profiles are all loaded the subsequent processing will fail. I also wanted to re-use as much existing code to speed up changes, so didn't implement the code server side
- Existing loadProfiles I think now becomes redundant

In **lib\report-plugins\daytoday.js**:
- Use profile.loadData to load daily basal profiles (which now has the correct profile range by using loadProfilesRange)
- This then means the subsequent call to profile.getTempBasal returns the correct value

In **lib/report_plugins/profiles.js**:
- added a check to make the routine more robust in error conditions

In **lib/server/profile.js** and **lib/api/profile/index.js**:
- added the same query model into profiles as the other APIs

In **profile.test.js**:
- Added new tests for multi-profile testing


hauraki:
Any news when this will be released?

Any more help needed? (I tested it successfully in https://github.com/nightscout/cgm-remote-monitor/issues/4991)

PR #5598, opened 2020-04-03
Add remote bolus/carbs + otp entry for loop

josep1972:
This branch implements components needed for adding remote bolus/carbs to Loop


francescaneri:
@sulkaharo @ps2 @scottleibrand @jasoncalabrese what do you think?

francescaneri:
@josep1972  big work!!

sulkaharo:
It's a bit hard to understand the full intent here by just reading the code. Could you write down a description of how the user and data flows are expected to work? Second, this doesn't add a timeline visualisation to the system nor checks on frequency and size of the doses, which would mean users might accidentally add multiple remote requests. Remote bolusing is inherently extremely risky, so some evaluation / implementation that mitigates user errors would make sense. Third, please check with the AndroidAPS team on their plans for similar features, where I'd prefer to implement features in a manner that's agnostic to the external systems and not have code that's specific to Loop, AndroidAPS or other systems

josep1972:
I apologize for the multiple commits. I was trying to get rid of the 'nutrition' fieldset when 'proteinGiven' fatGiven and carbsGiven where not selected, but I was not able to figure it out... Reverted back to the code that only adds the loop inputs and double check that the app builds correctly

josep1972:
@sulkaharo: Thanks for the feedback. There is already code specific to Loop  with respect to remote commands. The event types: Temporary Override and Temporary Override Cancel are sent to Loop via APNS, this code enhances the remote commands to Loop by adding the events Remote Carb Entry and Remote Bolus Entry with a required OTP field. 
Since remote dosing and carb entry is inherently serious operations, the OTP is checked at the Loop side, which implements RCF4226. This should mitigate any repeat bolus/carbs entries. Although not impossible, it is very difficult to enter the remote command more than once within 30 seconds of the timeout of the OTP.
I am not sure how AndroidAPS are intending to implement the feature, but I assume they will want some sort of OTP protection similar to this. The beauty of simply providing at passthrough of the OTP via the cgm-remote-manager is that it does not have to implement anything specific to security. The 'Looper' and remote provider exchange shared secret keys and "security method" without worrying about a compromised heroku,azure, etc app allowing a third party to send bolus/carb commands.

josep1972:
Here are some snapshots:

image

image

josep1972:
@sulkaharo Follow up  with the AndroidAPS team:  I communicated with @MilosKozak who commented that they are likely not ever going to use NS for remote commands to their system. I am not an android developer, but as I understand it, android apps can listen to SMS, so the solution to their system is for remote provider to send and sms to the device where the brain of the APS is running along with an OTP.

josep1972:
Is there anything else I need to do to have this PR considered? Thanks, Jose

Issue #5599, opened 2020-04-03
Announcement enters incorrect data types


ricksburnett:
**Describe the bug**
Adding a treatment log entry of Event Type Announcement fills protein and fat fields with empty strings.
Schema specifies these fields to be of type number

**To Reproduce**
1. Add Announcement event to log using web interface
2. Query API using api/v3/treatments?sort$desc=date&limit=1&eventType$eq=Announcement

**Expected behavior**
fat and protein fields should probably not even contain entries for an Announcement. If they need to be filled should be a number. Maybe 0 as default.

**Screenshots**
[{"enteredBy":"","eventType":"Announcement","reason":"","protein":"","fat":"","duration":0,"notes":"Test","created_at":"2020-04-01T20:40:33.534Z","utcOffset":0,"isAnnouncement":true,"identifier":"5e84fc4161b73397284379f4","srvModified":1585773633534,"srvCreated":1585773633534}]

**Your setup information**
Nightscout 13.0.1
- What version of Nightscout (e.g. 0.10.3)

**Additional context**
Causes problems in applications decoding json and expecting fields to have consistent types. Can workaround by getting all treatment not type Announcement and then requesting Announcements without those fields, which aren't really needed, but that's a bit of a pain.

PR #5619, opened 2020-04-14
Sanitize data from manual careportal entries

sulkaharo:
...so only fields with actual data are sent to the server


sulkaharo:
Should fix #5599

PR #5625, opened 2020-04-18
Configurable clock views

ireneusz-ptak:
This is a continuation of https://github.com/nightscout/cgm-remote-monitor/issues/5034

I was thinking about strategic solution for configurable clock views and my idea is to store config in the URL itself - easy to share, flexible and simple :)

EXAMPLE:
Current default "clock-color" could be written as
<NS URL>/clocks/cn13-sg40-nl-ar30-nl-ag6
which is parsed by NS as (split by '-', first chunk is generic config and the rest are clock parts):
c - color background
n - show time from last read only when threshold is breached
13 - 13 minutes threshold
"-" - next item
sg40 - sensor glucose (size 40)
nl - new line
ar30 - trend arrow (size 30)
nl - new line
ag6 - time since last read (size 6)

Current clock URLs are translated internally into this, but of course you may create your own watch like /clock/cy10-sg40-dt20-nl-ar30-nl-ag6 etc.

If this gets accepted my next step is to create a WYSIWYG config page available from NS menu so anyone can do this without checking the code for available items :)

Top views are new (maybe not pixel perfect, but I simplified CSS a lot):
clockviews


unsoluble:
Hmm! Lots to think about here. I like the novel approach to config store, but wondering how it fits... philosophically? with the rest of the way NS works. Would love to hear @sulkaharo weigh in.
Regardless, as soon as there's a clear method — not _necessarily_ a full WYSIWYG editor, but that'd be nice — for people to set things up, I'd be happy to ship this.

ireneusz-ptak:
@unsoluble Finally had some time to create the editor.
I also checked how those views behave in weird ratios and I think they look better than originals.

clock-config

unsoluble:
Looks great! I'll try running the PR here when I've got a chance over the next couple days, will let you know how it goes.

unsoluble:
Okay haven't had a chance to do a full barrage of testing, but spent a few minutes toying around. Initial thoughts:

- Overall super solid. Couldn't get anything to visibly break.
- Vertical padding between elements when there's only two on the screen is a *little* tight. Being allowed to insert multiple newlines instead of having them collapse down to one might be a reasonable workaround for people.
- Using the configurator on a mobile device is not good currently, as the first tap is always eaten by the show-the-close-button routine. Probably reasonable to override that while in config mode.
- Best if the original "Color" preset remains as it was before (as you've matched with the other two), so as not to pull the rug from under anyone.

Still waiting to hear from @sulkaharo re approval of the general concept, but I think you're close to a final product here.

ireneusz-ptak:
@unsoluble 
1. Great :)
2. That was the original idea, but I lost div height somewhere, fixed. I was also thinking about adding some kind of vertical separator, but wasn't sure if it's needed.
3. Can't reproduce on Chrome/Vivaldi on Android - the close button appears, but the form works as expected.
4. It should be already, check if you have SHOW_CLOCK_LAST_TIME and SHOW_CLOCK_DELTA set the same between instances.

unsoluble:
2. Great. :)
3. Okay, probably Safari-only then. Will still need a fix for this one, as that's going to be a very large proportion of the use case.
4. Aaahhh yeah, had those two set from a testing session a while back. All good now.

ireneusz-ptak:
3. Done.

Waiting for @sulkaharo then.

sulkaharo:
(Booking time for review in the next couple days, sorry for the delay!)

unsoluble:
Something's borked with this one after merge to dev — was working fine when I tested it as a pre-dev PR. But now when deployed it's producing empty clock views (reported by a user; I just confirmed).

ireneusz-ptak:
@unsoluble I have it deployed after merge: https://github.com/nightscout/cgm-remote-monitor/compare/dev...ireneusz-ptak:dev and it's working fine. Could you provide more details?

sulkaharo:
Working fine on my dev deployment as well. If you're seeing the + UI in the browser but the view is blank, that sounds like the browser cache is serving you an old file?

sulkaharo:
Ok I could replicate on one of the iPhones, where the configurable clock was all black. iOS seems to Really Want To Cache files; rebooting the phone fixed the issue on the phone.

sulkaharo:
Oh also - @ireneusz-ptak good work on this one! <3

unsoluble:
Wow, yep, full phone reboot required here to get the cache to clear for this one. Okay, guess that was all it is — will report back to the user. Thanks for verifying both!

Vanessa0022:
Is it possible to make the "Open my clock view" to not open upp in the whole webbrowser window when having the split screen setting?

I've created a request here:
https://github.com/nightscout/cgm-remote-monitor/issues/7425

ireneusz-ptak:
@Vanessa0022 You are doing it wrong. Create a clock view you want to use -> copy the page URL -> save it in your split screen configuration.

tiparraco:
No entiendo nada...puedes explicarme mejor, graciasEl 5 may. 2022 13:31, Vanessa0022 ***@***.***> escribió:
Is it possible to make the "Open my clock view" to not open upp in the whole webbrowser window when having the split screen setting?
I've created a request here:
#7425

PR #5631, opened 2020-04-24
Allow api3 behind reverse proxy

MilosKozak:


PieterGit:
@MilosKozak perhaps you can make this configurable based on `INSECURE_USE_HTTP`=`true`. This is used elsewhere in the code to redirect HTTP to HTTPS.

sulkaharo:
Yup @MilosKozak as Pieter mentions above, we have configuration settings for enabling HTTP use; rather than just commenting things out, this should enforce HTTPS by default and allow HTTP if the right config is set.

MilosKozak:
i know. this check is already done in app.js so i thought it would be just duplicated
@sulkaharo @PieterGit

sulkaharo:
Sorry this has been hanging - @PetrOndrusek would be the best person to evaluate this

bewest:
I can help test as well, but I'm a little foggy on intended behavior and intent.  I use nginx reverse proxy heavily.  I have a triple tiered nginx config with ssl ingress termination on nginx A, performing `auth_request` to nginx B for an app to inspect request.  A successful response indicates to nginx A to proxy to nginx C, which performs another `auth_request`, a successful response will cause final proxy to cgm-remote-monitor.  Looks like everything, including websockets works.  Would I notice if api3 was not working?  Issuing and authing tokens works fine.

PetrOndrusek:
Hi, APIv3 has been designed to keep high security from the start and therefore the possibility to run it using plain HTTP was not allowed. 
You can run it with `API3_SECURITY_ENABLE=false` env variable but this would turn off security completely (this is just for local development).
I would say that using proxy to redirect HTTPS to HTTPS is possible (as @bewest said) and it's just matter of configuration.
What is the actual problematic scenario to test?
In case of need we surely can incorporate `INSECURE_USE_HTTP=true` exception for APIv3, too.

MilosKozak:
there is no issue. i just told that more checking in api v3 code is not necessary because it's already done at app level. thus my patch -> remove extra check only

MilosKozak:
> Hi, APIv3 has been designed to keep high security from the start and therefore the possibility to run it using plain HTTP was not allowed.
> You can run it with `API3_SECURITY_ENABLE=false` env variable but this would turn off security completely (this is just for local development).
> I would say that using proxy to redirect HTTPS to HTTPS is possible (as @bewest said) and it's just matter of configuration.
> What is the actual problematic scenario to test?
> In case of need we surely can incorporate `INSECURE_USE_HTTP=true` exception for APIv3, too.

HTTPS to HTTPS only brings troubles with certificates. I think it's pretty rare config

PetrOndrusek:
OK, looks good to me.

sulkaharo:
Merging - @PetrOndrusek can you please PR in a code clean that removes the comment blocks.

PR #5638, opened 2020-04-28
Fix d3 portrait to landscape brush failure

jpcunningh:
* Upgrade to latest d3
* Optimize renderer's highlightBrushPoints
* Reset brush extent in updateBrushToNow (keeps brush extent in sync with screen changes)


lib/client/index.js
```diff
@@ -388,9 +388,11 @@ client.load = function load (serverSettings, callback) {
     // update brush and focus chart with recent data
     var brushExtent = client.dataExtent();
 
+    chart.theBrush.call(chart.brush)
```
jasoncalabrese, line 391: based on `chart.theBrush && ` a few lines below it looks like we should check that `chart.theBrush` isn't null here, grouping this with the call bellow in a single if would be better than using `&&` in both places

jpcunningh: Thanks @jasoncalabrese!  Requested changes are incorporated.

PR #5651, opened 2020-05-08
Fix Issue #5486 - Device Status Days Feature

### LINKED ISSUE #5486: Loop Pill data unavailable after 24 hours
**Describe the bug**
When scrubbing back in NS home screen graph, the Loop Pill shows no data and a warning symbol when you exceed 24 hours exactly from your current time. 

**To Reproduce**
1. Go to lower NS mini-graph
2. Click on graph and scroll back 24 hours and 1 minute or more
3. The loop pill will not show any data and have a warning symbol in place of the data.

**Expected behavior**
The Loop pill should remain intact, with its data and over data available like it does in the 23 hours since your current time.

**Screenshots**
24+ hours
image

Less than 24 hours
image

**Your setup information**
- What version of Nightscout (e.g. 13.0.2-dev)
- What type of CGM, and how do you get your data there? ShareBridge

**Additional context**
Running Loop Dev current as of 1/23/2020

mtfoley:
I'm attempting to fix #5486 , by allowing the number of days in Retro mode to be either a default of 1 or 2 days.  The value is set in `env.js`, used in `lib/data/dataloader.js`, and documented in `README.md`. I don't have a site with live data so I don't have a way to test this myself. I had started to run js-beautify on the source files, but there were a lot more changes than just the lines I was working with.

### DIFF
--- a/lib/data/dataloader.js
+++ b/lib/data/dataloader.js
@@ -361,8 +361,10 @@ function loadFood(ddata, ctx, callback) {
 }
 
 function loadDeviceStatus(ddata, env, ctx, callback) {
+    var retroDays = 1;
+    if(env.extendedSettings.devicestatus && env.extendedSettings.devicestatus.days) retroDays = env.extendedSettings.devicestatus.days;
     var dateRange = {
-        $gte: new Date(ddata.lastUpdated - ONE_DAY).toISOString()
+        $gte: new Date( ddata.lastUpdated - (ONE_DAY * retroDays) ).toISOString()
     };
     if (ddata.page && ddata.page.frame) {
         dateRange['$lte'] = new Date(ddata.lastUpdated).toISOString();
--- a/lib/data/dataloader.js
+++ b/lib/data/dataloader.js
@@ -361,10 +361,10 @@ function loadFood(ddata, ctx, callback) {
 }
 
 function loadDeviceStatus(ddata, env, ctx, callback) {
-    var retroDays = 1;
-    if(env.extendedSettings.devicestatus && env.extendedSettings.devicestatus.days) retroDays = env.extendedSettings.devicestatus.days;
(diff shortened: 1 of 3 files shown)


sulkaharo:
Couple considerations: 1) this is likely to cause rendering performance issues for loopers, so I'd prefer this is tested by opens and Loop users before merging, 2) can you add a mention into readme that this will roughly double the bandwidth usage of Nightscout, so those on a data cap might not want to enable this.

lib/data/dataloader.js
```diff
@@ -361,8 +361,10 @@ function loadFood(ddata, ctx, callback) {
 }
 
 function loadDeviceStatus(ddata, env, ctx, callback) {
+    var retroDays = 1;
+    if(env.extendedSettings.devicestatus && env.extendedSettings.devicestatus.days) retroDays = env.extendedSettings.devicestatus.days;
     var dateRange = {
-        $gte: new Date(ddata.lastUpdated - ONE_DAY).toISOString()
+        $gte: new Date( ddata.lastUpdated - (ONE_DAY * retroDays) ).toISOString()
```
sulkaharo, line 367: I think this would be much cleaner if instead of multiplying the constant you'd add a new TWO_DAYS constant and assign a variable to the constant based on the config

mtfoley: I think there may be a TWO_DAYS constant in that file already - if so I'll use that. I had originally been intending to use a multiplier but decided to limit it to two days for the reasons you mentioned on performance.

mtfoley:
I don't have a way to test it and I don't know anyone that I can ask to test it. I'll make the changes you asked for and I suppose we'll go from there.

mtfoley:
@sulkaharo, I made the requested changes.

PR #5671, opened 2020-05-21
do not redirect to profile editor by default

bewest:
In a default configuration, there is no treatment data.  The code to redirect
the UI to the profile editor is buried deep within the chart rendering code for
basals.  This plugin is only supposed to go into action when enabled via
ENABLE=basal.  This commit fixes first-use experience for the default
configuration intended to draw real-time CGM traces and no basal information is
expected.  Since no basal information is expected unless plugin is enabled via
ENABLE=basal, this allows skipping instead of redirecting to the profile editor.

### DIFF
--- a/lib/client/renderer.js
+++ b/lib/client/renderer.js
@@ -1034,6 +1034,9 @@ function init (client, d3) {
 
   renderer.addBasals = function addBasals (client) {
 
+    if (!client.settings.isEnabled('basal')) {
+      return;
+    }
     var mode = client.settings.extendedSettings.basal.render;
     var profile = client.sbx.data.profile;
     var linedata = [];


sulkaharo:
I know this is a long ask but... would you be willing to look at moving the profile redirect check to somewhere better? I general, we don't currently have a good mechanism during the site startup to check if everything is ok and report issues to users and having something in this vein would be great.

bewest:
I'm intrigued.  Can anyone help me understand what kind of checks, and maybe I can make some better arrangements.  IIRC, basal plugin is not one of the ones included by default.

I'd be motivated to improve the startup experience in general.

sulkaharo:
First things that come to mind:

* If database connection fails, show a user friendly message that tells what's going on
* If database connection succeeds but DB is empty, tell user this is the case and instruct to check the uploader setup (users can't distinguish between uploading failing and UI not loading data)
* If database connection succeeds and there is data but it's clearly malformed, tell user what's going on (some uploaders send in junk and should be avoided / there should be more data validation in the REST API or uploaders should switch to the v3 api)
* If required config variables are missing, tell user how to add those
* If site is in MMOL mode and targets are outside the MMOL range, tell user / if site is mg/dl mode and targets are single digits, tell user

Issue #5640, opened 2020-04-28
Azure deploy fail


CaioBauab:
Azure deploy fail

I follow the wiki http://www.nightscout.info/wiki/welcome/set-up-your-nightscout-website-deploy-to-azure but the deploy fail with error: 

Submitting Deployment
Deployment Started
Creating Web Hosting Plan
Creating Website
Setting up Source Control
Deployment Failed (View Error)

Deployment Error
[...]

Issue #5503, opened 2020-02-02
build official docker image for each new release


niepi:
**Is your feature request related to a problem? Please describe.**
I run NS in docker/kubernetes and build the image myself when a new version is released.

**Describe the solution you'd like**
Provide an official NS docker image on Docker Hub. This could be achieved with the following steps:

- provide a dockerfile for the official image
- integrate docker hub to build this image automatically when a new tag is push to the repo

DOFfactory:
I fully support this, I would love to run NS on my own server accessible via VPN. I am somewhat hesitant running it on cloud.

niepi:
Is there anything I can do? 

The repo has a Dockerfile.example that basically works, but most of the work is the configuration on the docker hub side.

DOFfactory:
Well, as far as I can see one has to `git clone` instead of downloading a ready made docker container everything with 1 command, and a `docker-compose` would be lovely (and much simpler than cloning and building and debugging for build errors)

niepi:
this is my current docker-compose setup:

```
version: "2"
networks:
    nginx-proxy:
    nightscout:
services:
  nginx-proxy:
  ...
  letsencrypt:
  ...
  database:
    restart: unless-stopped
    image: 'mongo:latest'
    environment:
      AUTH: 'true'
    volumes:
      - ./data/mongo:/data/db:cached
      - ./import:/import
    ports:
      - '27017'
    networks:
      - nightscout

  nightscout:
    restart: unless-stopped
    image: 'niepi/cgm-remote-monitor:13.0.1'
    environment:
      VIRTUAL_HOST: ns.example.com
      VIRTUAL_PORT: 1337
      LETSENCRYPT_HOST: ns.example.com
      LETSENCRYPT_EMAIL: example@example.com
      API_SECRET: 'secret'
      AUTH_DEFAULT_ROLES: 'denied'
      MONGO_CONNECTION: 'mongodb://nightscout:secret@database/nightscout'
      PORT: 1337
      MONGO_COLLECTION: 'entries'
      NIGHT_MODE: 'off'
      SHOW_RAWBG: 'never'
      THEME: 'default'
      TIME_FORMAT: '24'
      DISPLAY_UNITS: 'mg/dl'
      ... # more NS env vars
    networks:
      - nightscout
      - nginx-proxy
    ports:
      - '1337'
```

niepi:
how can we help to get this issue forward? 

docker build works fine for me. 

[...]

PR #5716, opened 2020-06-27
FIX issue #4852 Dockerfile does not build with 0.12.3 

skalahonza:
I edited the Dockerfile so it can be acutally built. I also added a docker-compose file so you can run nightscout and mongodb locally using docker-compose up. 

This is also related to 
* Azure deploy fail #5640
* build official docker image for each new release #5503

Because it is now possible to run Docker images pushed to DockerHub in the Azure AppService.

## Building official Docker image
It is now possible to let DockerHub build image for you. All you need to do is to link it with the official repository.
https://docs.docker.com/docker-hub/builds/

You can build images with tag dev from dev branch and official latest/production images from master branch with ease.


docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    environment:
+      API_SECRET: "${API_SECRET:-123456789abc}"
+      MONGO_CONNECTION: 'mongodb://${MONGODB_USERNAME:-nightscout}:${MONGODB_PASSWORD:-secret}@database/nightscout?retryWrites=true&w=majority&authSource=admin'
+      MONGO_COLLECTION: 'entries'
+      INSECURE_USE_HTTP: "true"
+      SECURE_HSTS_HEADER: "false"
+      PORT: 1337
+      TIME_FORMAT: "24"
+      DISPLAY_UNITS: "mmol"
+      WAIT_HOSTS: "database:27017"
```
m-a-v, line 21: This doesn't work for me. I had to use something like

`command: sh -c "sleep 5; node server.js"`

skalahonza: Can you pls post the std out output in the comments? It works fine on my machine.

m-a-v: Actually I am not using your branch, but my docker compose file is very similar. I've also added `WAIT_HOSTS`. I think it is a timing problem. As you can see in the startup sequence the nightscout-app is not waiting long enough. Although nightscout-app makes afterward a successful connection, Nightscout doesn't work. I didn't have this problem with the same docker-compose.yml file before upgrading to 14.0.7.

```
nightscout-db | 2020-11-27T16:01:00.060+0000 I CONTROL  [initandlisten] MongoDB starting : pid=1 port=27017 dbpath=/data/db 64-bit host=a6370c7791ae

...

nightscout-app | Error connecting to MongoDB: {"name":"MongoNetworkError"} - retrying in 3 sec
nightscout-app | ERROR CONNECTING TO MONGO Error: MongoDB connection failed! Double check the MONGODB_URI setting in Heroku.
nightscout-app |     at MongoClient.connect.then.catch.err (/opt/app/lib/storage/mongo-storage.js:75:30)
nightscout-app |     at process._tickCallback (internal/process/next_tick.js:68:7)
nightscout-app | Mongo Storage system ready

...

nightscout-db | 2020-11-27T16:01:01.177+0000 I NETWORK  [initandlisten] waiting for connections on port 27017

...

nightscout-app | Successfully established a connected to MongoDB
nightscout-app | Mongo user role seems ok: [ { role: 'root', db: 'admin' } ]
nightscout-app | Mongo Storage system ready

```

skalahonza: WAIT_HOSTS is not a built in functionality, I had to add support for it in the Dockerfile
```
## Add the wait script to the image
ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.7.3/wait /wait
RUN chmod +x /wait
```

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3.4"
+
+volumes:
+  mongodb_data_container:
+
+services:
+  nightscout:
+    image: "skalahonza/nightscout"
```
m-a-v, line 8: This seems to be obsolete and should be discarded.

skalahonza: Good point. Discarded.

sulkaharo:
Apologies for the long latency - DEV now has working Docker setup, coupled with Github Actions Docker build. More improvements to land soon. https://hub.docker.com/r/nightscout/cgm-remote-monitor/tags?page=1&ordering=last_updated

m-a-v:
@sulkaharo No worries for the long latency. I have running a similar setup like in this PR for several months. I think it would be important also to add a docker compose solution for the people. If you want to I could review and test @skalahonza in more detail.

sulkaharo:
@pazaan is working on the compose - we were discussing supporting both compositions that only contain the Nightscout Node component but also one that has Mongo as well, which would be a fully self contained image.

m-a-v:
@pazaan Is there some progress for docker compose?

pazaan:
See https://github.com/nightscout/cgm-remote-monitor/pull/6903

PR #5717, opened 2020-06-27
Suggested changes for Remote Bolus/Carbs

jasoncalabrese:
@bewest @sulkaharo @josep1972
cc @ps2

This PR is mostly to point out some issues I saw, and start some discussion, feel free to replace with a different/better PR.

I was looking at the recent merge for #5598, and saw a few things that I think need to be cleaned up and improved.

The main issue is that I don't think the Remote Bolus or Carbs options should be included in the careportal unless they are enabled in Loop.  I think that could work in a similar way to how Loop uploads the available `overridePresets`.  I'm not sure when the Loop PR will be merged, but I assume it will take some time.

### DIFF
--- a/lib/plugins/loop.js
+++ b/lib/plugins/loop.js
@@ -231,7 +231,7 @@ function init (ctx) {
         , split: false
         , targets: false
         , reasons: reasonconf
-        , otp: true 
+        , otp: false //TODO: was why was this set to true? I don't think we need that, maybe an option?
         , submitHook: postLoopNotification
       },
       {
@@ -249,19 +249,21 @@ function init (ctx) {
         , targets: false
         , submitHook: postLoopNotification
       },
+      //TODO: Remote Carbs Entry should only be available if enabled in Loop
       {
         val: 'Remote Carbs Entry'
         , name: 'Remote Carbs Entry'
-        , remoteCarbs: true 
-        , remoteAbsorption: true 
-        , otp: true 
+        , remoteCarbs: true
+        , remoteAbsorption: true
+        , otp: true
         , submitHook: postLoopNotification
       },
+      //TODO: Remote Bolus Entry should only be available if enabled in Loop
       {
         val: 'Remote Bolus Entry'
         , name: 'Remote Bolus Entry'
-        , remoteBolus: true 
-        , otp: true 
+        , remoteBolus: true
+        , otp: true
         , submitHook: postLoopNotification
       }
     ];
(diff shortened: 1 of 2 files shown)


lib/server/loop.js
```diff
@@ -1,4 +1,4 @@
-//'use strict';
+'use strict';
```
jasoncalabrese, line 1: @josep1972 why was this commented out?

bewest:
Thanks @jasoncalabrese, I agree.  Plugins shouldn't do anything unless they become enabled, especially for features that require additional configuration or change liability and risk created by using or publishing the software.

lib/server/loop.js
josep1972: Hm...to be honest I don't remember...

josep1972:
I am not sure I agree @jasoncalabrese that loop needs to do anything other than have an OTP shared with the remote provider. For temp overrides, the plugin is available even if overridePresets is not received from loop, except no overrides can be chosen. The same reasoning is used with these new plugins. Yiou can decide to make them available, but if you don't have a valid OTP the plugin will not work, thus not "enabled". In other words, there is a difference between available and enabled. By editing the environment variables you make the plugins available. By adding presets and sharing the OTP with a remote provider you make the plugins enabled. There is no current clear "enable remotes" setting in loop, so I would have to defer to @ps2 on this if he thinks there should be one now. But, you can argue that creating the actual override presets and a shared OTP are themselves the enable settings.

lib/plugins/loop.js
```diff
@@ -231,7 +231,7 @@ function init (ctx) {
         , split: false
         , targets: false
         , reasons: reasonconf
-        , otp: true 
+        , otp: false //TODO: was why was this set to true? I don't think we need that, maybe an option?
```
josep1972, line 234: The intention is for loop to have the option to require an OTP even for remote overrides. It was set to true on purpose. The current loop code ignores it for now, but this makes the OTP available if the looper desires to use it in the future. 
In fact, I firmly believe that the remote overrides should have used an extra level of authentication when it was first implemented via an OTP. Loopers that are already using the feature have come to expect no extra OTP, though. But, this code is meant to move into a more secure direction wrt doing things remotely.

josep1972: Here is a long thread about the feature: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Remote.20bolus.20.2F.20remote.20carbs   
On my March 30 posting I talk about OTP for overrides too...

josep1972: I don't agree that this should be changed. Having the ability to enable more security is highly desirable.

bewest:
Where do we sit with this and what needs to be done for merge and release?

Issue #5726, opened 2020-06-30
Broken Bolus wizard in Dev


bjornoleh:
The bolus wizard is not working for me in NS-Dev as of June 24. The BG field turns red, no bolus/carbs advice is given. Changing decimal separator makes no difference. 

The Bolus wizard was working on master before, and on Dev from a couple of weeks back.  Tested in Chrome Win10, and Android Chrome. I did clear browser cache, this made no difference,

I have successfully I rolled back to Master, and got the Bolus wizard working again.

And then I reverted my repository to get back to Dev as of May 23. This version also has a working Bolus wizard (this was before the configurable clock view and other changes were added)

![Merknad 2020-06-29 112221](https://user-images.githubusercontent.com/63544115/86112011-bc61b680-bac7-11ea-8191-fc1107a663e2.png)

Thanks!

ireneusz-ptak:
I have never used this feature but it seems it accepts mg/dl only (checked on master branch, could you confirm it @bjornoleh?), so integer validation should be fine in that case.

So far I prepared a fix that brings back proper bolus calculations (https://github.com/nightscout/cgm-remote-monitor/pull/5729).

bjornoleh:
Hi, it worked with mmol/L with 0,1 mmol/L precision just fine in both Master and earlier Dev builds. So something happened very recently in Dev.

But now I just merged your fix with my repo, and deployed to Heroku. And it works! Thank you!

ireneusz-ptak:
![image](https://user-images.githubusercontent.com/31506973/86181838-121d7980-bb2f-11ea-8042-956f8fc4b898.png)

That's my screenshot from 13.0.1 - BG value is red and numbers make no sense (at least for BG provided in mmol/l).
I'll have to check that later.

bjornoleh:
Hm, I just learned something here.. It turns out the BG cell do only expect integer values, and there is even a small popup text when you mouse-over this cell. Pink is indicating this error, but the calculations actually seem ok, and the entered value is used in my case (mmol/L).

It was actually the insulin correction dose that turned red in my case, see my initial screenshot. Sorry for beeing sloppy!

But I agree that the value in your pic does not make sense. It should be positive for BGs above your target, and negative below. Some other problem in Master then?

Using the ireneusz-ptak:dev that I just merged with my repo, it seems fine.

![image](https://user-images.githubusercontent.com/63544115/86183306-01bace00-bb32-11ea-8f11-5272c0d309a8.png)

[...]

PR #5728, opened 2020-06-30
Fix boluscalc mmol issue

ireneusz-ptak:
Fix for https://github.com/nightscout/cgm-remote-monitor/issues/5726

Also a master branch problem.


ireneusz-ptak:
Ah, it fixes only BG validation - need to check the rest.

PR #5729, opened 2020-06-30
Fix for boluscalc issue

ireneusz-ptak:
The PR fixes https://github.com/nightscout/cgm-remote-monitor/issues/5726

It seems the changes in getCurrentProfile in lib/profilefunctions.js were made by accident (@fromorbonia). Returning spec_profile doesn't make sense here.

The open question is why boluscalc works in mg/dl even if NS is set to mmol.


sulkaharo:
LGTM, thanks

fromorbonia:
@ireneusz-ptak - thanks for spotting that, I was trying to split out the original spec_profile handling, but I didn't put in the data.store retrieval against the spec_profile. 
`    var timeprofile = spec_profile || profile.activeProfileToTime(time);
    return data && data.store[timeprofile] ? data.store[timeprofile] : {};`

PR #5725, opened 2020-06-29
Hotfix/2020 06 28/carelink

bewest:
Hotfix for master.  Similar to https://github.com/nightscout/cgm-remote-monitor/pull/5721

PR #5735, opened 2020-07-03
Switch to using package-lock.json

Grynn:
Package-lock:
* newer (introduced in npm 5)
* Not published (except for top level packages/web apps)
* Inspired by yarn.lock

If shrink-wrap is present, package{,-lock}.json is ignored. 
If package-lock.json is present, and newer than package.json then it is used. 

This gives all the benefits of repeatable installs while making it harder to forget to checkin the latest package-lock.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4734,22 +4734,26 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "bundled": true,
+          "resolved": false,
+          "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
           "optional": true
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
+          "resolved": false,
+          "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
           "optional": true
         },
         "aproba": {
           "version": "1.2.0",
-          "bundled": true,
+          "resolved": false,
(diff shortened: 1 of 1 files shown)


Grynn:
For example: package.json refers to `minimed-connect-to-nightscout": "nightscout/minimed-connect-to-nightscout#hotfix-2020-06-28-EU-01`, but shrinkwrap refers to `github:nightscout/minimed-connect-to-nightscout#hotfix-2020-06-28-EU` 

whereas package-lock.json correctly refers to `github:nightscout/minimed-connect-to-nightscout#hotfix-2020-06-28-EU-01`

bewest:
FWIW, I believe https://github.com/nightscout/cgm-remote-monitor/pull/5725 targets the new release v1.4.0 of minimed-connect-to-nightscout.

Grynn:
Absolutely; the idea is for package-lock.json to always reflect the current package.json.
My understanding is that `npm install` auto generates package-lock.json if it is missing; also if package.json is newer than package-lock.json it gets updated to stay in sync with package.json

## No package-lock and no shrinkwrap
`npm install` - generates package-lock.json using the latest version of dependencies allowed by package.json. Future invocations of `npm install` will use the package-lock.json file, effectively caching the resolution of dependency versions (until package.json is modified). For example if package.json has `foo: ~1.2` and when package-lock was generated latest published "foo" was 1.2.4 then that is written to package-lock.json. If 1.2.5 is released, `npm install` will not care (unless package.json is also updated). This is just the right behaviour IMHO.

## Shrinkwrap
Once generated, it has to be manually re-created, adding to maintenance burdens.

## Why the Pull Request
If merged, this will switch from shrinkwrap to package-lock going forwards. As soon as #5725 is merged into dev, package-lock will update to reflect this (since package.json would be modified). This would not happen automatically with Shrinkwrap.

sulkaharo:
Merged through https://github.com/nightscout/cgm-remote-monitor/commit/14872695ba1ae8d9477b610fefc06993c6e81bbe

Issue #4971, opened 2019-09-07
Loop Basal Rate Actual Delivery vs Nightscout Day to Day Report


marionbarker:
EDITED: Anything in {} should be disregarded.  Please see my later post (9/7/2019). I did not delete anything.  Just added newline{ and }newline around incorrect assumptions.

Please include the following information with your bug report:

Is this a new bug with the latest release, have you seen this earlier? 
The reported basal values in Day-to-Day Report differ significantly for Omnipod Loop from the actual values.
I believe this is an inconsistency in method for estimating basal from rate*time on NS vs Loop (actual insulin delivered).  
{
Significant differences are noted for Omnipod which must cancel TB before initiating the next TB, so the period for the TB is typically just under a minute mark and NS rounds down the minute.
} 
[Sep 12, 2019: EDITED to add examples from database records for "eventType": "Temp Basal",: old Loop code had "duration": 4, new Loop code (after Sep 9, 2019, dev) has "duration": 4.767875816424688, and new field: "amount": 0.1,
END EDIT ]

See discussion here: 
https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Nightscout.20daily.20total.20temp.20basal.20delivery/near/174760280

{
I think that Mark Yeager is correct that the primary problem is rounding down the minutes for TB. He said:
"It seems the "revert to programmed basal" that Loop issues between temp basals (as @ps2 explained) results in a 4 minute increment in NS vs the 4min:57sec ? in Loop.."
}

The other part of the problem is something I've noticed before - if you change your profile for scheduled basal, the day-to-day report changes because the intervals between TB are modified to new basal pattern.

Which device / browser version was used to reproduce the bug?
Running dev branch from https://github.com/LoopKit/LoopWorkspace
Using Omnipod, iPhone 7
Running dev on Nightscout (0.12.4-dev)

Does this happen every time you launch Nightscout, or sometimes?
yes

Steps how to reproduce (if you can’t reproduce the issue, please don’t report the issue / if we can't reproduce the bug, we can't fix)

I am attempting to reproduce bug by taking the treatment report and doing rate * time and summing up to match the reported basal.  

{ 
If I use an extra minute of Scheduled Basal between each TB reported by Loop plus the actual scheduled basal, I get close to the overestimate provided by NS. 
}

 I want to clean up the spreadsheet some to make it more understandable before posting it.

Please include 1 or more screenshots of the issue appearing, ideally with an annotation on what's wrong

[...]

PR #5740, opened 2020-07-05
effected temp basal amounts

bewest:
https://github.com/nightscout/cgm-remote-monitor/issues/4971

* For loop/omnipod users, nightscout does not correctly calculate effected temp basal amounts for omnipod
* Loop users as of September 2019 updates have new data in the form of `.amount` contains effected rate 

Impacts nightscout reports for loop/omnipod users and becomes detectable comparing reports or even adding things up manually.
It looks to me like the issue in the daytoday reports traces through to [line 369](https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/profilefunctions.js#L351-L383) in `getTempBasal`.

There are 18 *other* places that similarly use `.absolute` and conceivably some of them will need a similar patch or update to use a method to calculate effected rates.  Thoughts?

```
lib/client/careportal.js:279:      data.absolute = Number(absolute);
lib/client/careportal.js:415:    pushIf('absolute' in data, translate('Basal value') + ': ' + data.absolute);
lib/client/renderer.js:1171:          .text((t.percent ? (t.percent > 0 ? '+' : '') + t.percent + '%' : '') + (isNaN(t.absolute) ? '' : Number(t.absolute).toFixed(2) + 'U') + (t.relative ? 'C: +' + t.relative + 'U' : ''));
lib/plugins/basalprofile.js:86:        !isNaN(basalValue.treatment.absolute) ? basalValue.treatment.absolute + 'U/h' : '';
lib/plugins/bgnow.js:166:      recent.mean - delta.absolute / delta.elapsedMins * 5 : recent.mean - delta.absolute;
lib/plugins/bgnow.js:191:      info.push({label: translate('Absolute Delta'), value: sbx.roundBGToDisplayFormat(sbx.scaleMgdl(delta.absolute)) + ' ' + sbx.unitsLabel});
lib/plugins/treatmentnotify.js:148:      (!isNaN(lastTreatment.absolute) ? '\n' + translate('Value') + ': ' + lastTreatment.absolute + 'U' : '')+
lib/profilefunctions.js:368:    if (treatment && !isNaN(treatment.absolute) && treatment.duration > 0) {
lib/profilefunctions.js:369:      tempbasal = Number(treatment.amount || treatment.absolute);
lib/report_plugins/daytoday.js:674:          //            .text((t.percent ? (t.percent > 0 ? '+' : '') + t.percent + '%' : '') + (t.absolute ? Number(t.absolute).toFixed(2) + 'U' : ''));
lib/report_plugins/treatments.js:133:    pushIf(!isNaN(data.absolute), translate('Basal value') + ': ' + data.absolute);
lib/report_plugins/treatments.js:205:            data.absolute = $('#rped_absolute').val();
lib/report_plugins/treatments.js:213:            if (data.absolute === '') {
lib/report_plugins/treatments.js:214:              delete data.absolute;
lib/report_plugins/treatments.js:241:        $('#rped_absolute').val('absolute' in data ? data.absolute : '');
lib/report_plugins/treatments.js:334:        .append($('<td>').attr('align','center').append('absolute' in tr ? tr.absolute.toFixed(2) : ''))
lib/server/treatments.js:169:  obj.absolute = Number(obj.absolute);
lib/server/websocket.js:315:             if (data.data.absolute) {
lib/server/websocket.js:316:              query_similiar.absolute = data.data.absolute;

```

### DIFF
--- a/lib/client/renderer.js
+++ b/lib/client/renderer.js
@@ -1168,7 +1168,7 @@ function init (client, d3) {
           .attr('dy', '.35em')
           .attr('x', chart().xScaleBasals((Math.max(t.mills, from) + Math.min(t.mills + times.mins(t.duration).msecs, to)) / 2))
           .attr('y', 10)
-          .text((t.percent ? (t.percent > 0 ? '+' : '') + t.percent + '%' : '') + (isNaN(t.absolute) ? '' : Number(t.absolute).toFixed(2) + 'U') + (t.relative ? 'C: +' + t.relative + 'U' : ''));
+          .text((t.percent ? (t.percent > 0 ? '+' : '') + t.percent + '%' : '') + (isNaN(t.absolute) ? '' : Number(t.amount || t.absolute).toFixed(2) + 'U') + (t.relative ? 'C: +' + t.relative + 'U' : ''));
         // better hide if not fit
         if (text.node().getBBox().width > chart().xScaleBasals(t.mills + times.mins(t.duration).msecs) - chart().xScaleBasals(t.mills)) {
           text.attr('display', 'none');
--- a/lib/plugins/treatmentnotify.js
+++ b/lib/plugins/treatmentnotify.js
@@ -145,7 +145,7 @@ function init() {
       (lastTreatment.insulin ? '\n' + translate('Insulin') + ': ' + sbx.roundInsulinForDisplayFormat(lastTreatment.insulin) + 'U' : '')+
       (lastTreatment.duration ? '\n' + translate('Duration') + ': ' + lastTreatment.duration + ' min' : '')+
       (lastTreatment.percent ? '\n' + translate('Percent') + ': ' + (lastTreatment.percent > 0 ? '+' : '') + lastTreatment.percent + '%' : '')+
-      (!isNaN(lastTreatment.absolute) ? '\n' + translate('Value') + ': ' + lastTreatment.absolute + 'U' : '')+
+      (!isNaN(lastTreatment.absolute) ? '\n' + translate('Value') + ': ' + (lastTreatment.amount || lastTreatment.absolute) + 'U' : '')+
       (lastTreatment.enteredBy ? '\n' + translate('Entered By') + ': ' + lastTreatment.enteredBy : '') +
 
       (lastTreatment.notes ? '\n' + translate('Notes') + ': ' + lastTreatment.notes : '');
--- a/lib/profilefunctions.js
+++ b/lib/profilefunctions.js
@@ -366,7 +366,7 @@ function init (profileData) {
 
     //special handling for absolute to support temp to 0
     if (treatment && !isNaN(treatment.absolute) && treatment.duration > 0) {
-      tempbasal = Number(treatment.absolute);
+      tempbasal = Number(treatment.amount || treatment.absolute);
     } else if (treatment && treatment.percent) {
       tempbasal = basal * (100 + treatment.percent) / 100;
     }
--- a/lib/report_plugins/treatments.js
+++ b/lib/report_plugins/treatments.js
@@ -130,7 +130,7 @@ treatments.report = function report_treatments(datastorage, sorteddaystoshow, op
     pushIf(data.insulin, translate('Insulin Given') + ': ' + data.insulin);
     pushIf(data.duration, translate('Duration') + ': ' + data.duration);
     pushIf(data.percent, translate('Percent') + ': ' + data.percent);
-    pushIf(!isNaN(data.absolute), translate('Basal value') + ': ' + data.absolute);
+    pushIf(!isNaN(data.amount || data.absolute), translate('Basal value') + ': ' + (data.amount || data.absolute));
     pushIf(data.preBolus, translate('Carb Time') + ': ' + data.preBolus + ' ' + translate('mins'));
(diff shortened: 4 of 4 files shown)


bewest:
I've audited all the places mentioned above and have questions about these two:

## Treatments
Should default treatments that are uploaded be forced to have an `amount` field?  I think no: it would defeat the patches added so far that will only use it when available.

https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/server/treatments.js#L169


## Websocket Query
Should websocket be able to query by `amount`?  Maybe?

https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/server/websocket.js#L315

sulkaharo:
Sorry for comment lag - I'd very much prefer an approach where 1) we try to get the data formats unified across clients rather than supporting client quirks in the server (so this would be changed in Loop, not NS), and 2) if we support multiple data formats, we'd centralise data normalization to one part of the code such as the data loader, so the application logic only needs to know of one data format. If we go with the approach you have here (support multiple fields across the codebase) that easily leads to bugs, as devs would have to remember to use both fields at all times and it's then also a matter of time before someone adds a third field with the same data and adding support for that would be significantly easier if we have just one place where we add the field. Related, check the data formats specified in the V3 API; for maintainability it'd be great if we aim to not expand the feature set of the v1 api and modify the accepted data formats, vs encourage clients to migrate to v3.

bartsluimer:
@ps2 would it be possible to have a look?

bewest:
Thanks for the good criticism @sulkaharo, I think I agree with your reasoning.
Do you mean 1.) Loop should be submitting correct `amount` instead of adding a second "more correct" version of `absolute`?  I agree, @ps2 can you add any detail on why adding the new `absolute` field is needed instead of assigning the correct value to `amount`?  (This would have been prior to September 2019).


2.) I saw a place to do this during my audit, but it changes the logic in this patch, which depends on absence by default.

ps2:
@bewest Super happy to see someone tackling this! I'm not 100% sure I understand your question, but I can explain the reasoning behind sending the original scheduled `rate` and a total delivery `amount`. Users get confused (as we have seen in Tidepool web) if we display a rate that was different from what they saw being scheduled, or even weirder, a rate the pump doesn't actually support. 

For example: If the Loop schedules a 1.2U/hr basal for 5.5 minutes on an Omnipod, which delivers 2 x 0.5 clicks  and it shows 1.2 U/hr (0.1U over 5.5m = 1.09090909 U/hr), instead of 1.2 U/hr, that's a very confusing experience. It's useful to have both pieces of information: the original scheduled rate, and the actual delivery. Reversing out the calculation of a rate from a delivery amount and time period produces numbers that can be unexpected and confusing for users.

ps2:
Also, an absolute "rate" of 1.09090909 U/hr is not even the actual rate of delivery happening then; that's just the *average* rate over those 5.5 minutes, so it's additionally misleading.

I've made a lot of code updates in Loop to track these things separately. In retrospect (and this statement applies to both code bases). It would have been better for the model to contain `programmedRate`, and `totalDelivery`.  Those names are clear to future developers, and keep the needed distinction.

marionbarker:
@bewest Thanks for working on this.

I saw the comment from @sulkaharo 
``` quote
we try to get the data formats unified rather than supporting client quirks in the server  (so this would be changed in Loop, not NS)
````
I believe there is an underlying design issue that probably "grew" as NS reports were added with underlying assumptions that don't match some newer real-world use cases. (And I'm not trying to throw darts here - I completely understand how this happens.)

I'm assuming that this is understood, but just in case, I'll belabor the point. There is no pump on the planet that provides continuous insulin drip and therefore calculating the amount of insulin delivered as rate times duration for a basal will always be an estimate.  It is typically an overestimate for omnipod (because Insulet designed the pod to deliver its smallest increment at the end of a time unit - not the mid-point like Medtronic). I don't know about any other pump that might have data uploaded to Nightscout. I don't know if the folks supporting Omnipod for AAPS (or other instantiations for Omnipod) are uploading an ```amount``` in addition to rate and duration, but it seems reasonable to me for this to be supported by NS if it is provided. 

Regarding the second comment: 
``` quote
2) if we support multiple data formats, we'd centralise data normalization to one part of the code such as the data loader, so the application logic only needs to know of one data format.
```

This is great - can we add into this that ```profiles``` be used too.  At the current time, many of the NS reports only use the current ```profile``` which leads to inaccurate reports for daily insulin dosing, independent of the basal amount errors. 

I think that a lot of the design of NS reports assume ```rate*duration``` is a valid estimate for ```amount``` to be added or subtracted from the scheduled basal using the current, rather than the time-relevant, ```profile```.  Seems like a pretty far reaching problem in report design.

Thanks to all
BTW - I use Nightscout for many things (just not for daily dose records).  It's incredibly valuable, but could be even better.

sulkaharo:
Apologies for the delay - #5980 now does the data normalization of the amount/absolute fields on runtime data load, which solves the immediate issue for every usage in the codebase except reports with less new logic overall. Note that in this PR, some of the changes would not work as the code is inside code blocks that are not reached if the `absolute` data field is missing, so I think data normalization is the way to go (and this includes changes a bunch of things in the codebase; this is not Loop specific any way).

PR #5761, opened 2020-07-14
feat: exporter reporting module

jstarpl:
This PR implements an Export module, allowing easy export of data in various formats (JSON, CSV, XLSX). It uses the *Reporting* page as a host.

### DIFF
--- a/lib/report_plugins/export/CSVConverter.js
+++ b/lib/report_plugins/export/CSVConverter.js
@@ -1,0 +1,24 @@
+export class CSVConverter {
+    constructor(columns, data, options) {
+        this.columns = columns;
+        this.data = data;
+        this.separator = ',';
+        this.extension = 'csv';
+        if (options.semicolonSeparated) {
+            this.separator = ";";
+        }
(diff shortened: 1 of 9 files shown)


sulkaharo:
Hi! This is tagged as draft, did you progress further on this?

jstarpl:
> Hi! This is tagged as draft, did you progress further on this?

I didn't do much beyond this, but as far as I remember this is perfectly functional and should work. I guess I should actually mark it as ready for review.

sulkaharo:
Hi! Right - I'm a bit hesitant to add the xlsx export as a client feature as it seems to bloat the client javascript package quite a bit and adds a lot of new dependencies to the client (especially since we have newer PR that simplifies the packaging to just a single bundle). As such it'd be great to simplify this to just CSV / JSON or figure a way to support the file downloads from the server. Btw note the API itself can already return CSV files (and saves JSON by default) for treatments and entries, so theoretically the report side would just need to redirect the client to the right API endpoint with the right date query.

PR #5771, opened 2020-07-20
added calculator

sebastianrauer:


sebastianrauer:
as is

sulkaharo:
Hi! Was this intended to be included in the mainline Nightscout? We're not including external components such as this into the codebase right now, so if you want to have this feature, I'd suggest finding a way to include this as a node module or some other means of not linking to external sites on the client side.

sebastianrauer:
> Hi! Was this intended to be included in the mainline Nightscout? We're not including external components such as this into the codebase right now, so if you want to have this feature, I'd suggest finding a way to include this as a node module or some other means of not linking to external sites on the client side.

Did not meant to push that into the repo. Sorry. Was a mistake.

PR #5781, opened 2020-07-27
reinstall unusable NS site

ronnieg123:


ronnieg123:
My current NS site is unusable as I removed the mlab MONGODB before confirming MONGODB Atlas was operational.

ronnieg123:
cannot reinstall unusable NS site as app has been deleted.

ronnieg123:
need to create new NS app with Heroku

PR #5789, opened 2020-07-29
Dev

tjohnsen83:


sulkaharo:
Hi! Oops - it seems this pull request was made the wrong way round - you're trying to pull changes from your copy of Nightscout into the master version, when the correct direction is from master into your copy. Don't worry, no harm done! Please try using the Nightscout update tool again and pay special attention to the source (Nightscout master) and target (your copy) merge direction settings. We'll close this pull request, as it's now in the Nightscout repository instead of the intended your personal copy.

tjohnsen83:
Hi.. I saw it too late, but i put it the other way around, so i think It’s okei now?

Sendt fra min iPhone

PR #5945, opened 2020-09-03
Openaps performance improvement

jpcunningh:
Scroll performance improvement by drastically reducing the amount of times moment has to parse date strings. There is a corresponding oref0 update that provides the mills fields.


sulkaharo:
Great find! Actually, can you refactor this so that the objects are parsed on load (or somewhere else, but just once) to always have the mills field even if that's not in the database, which would allow simpler logic to run for the updates. We have such optimisations in multiple places already and at least to me it's better to preprocess data and keep runtime logic as simple as possible

PR #5969, opened 2020-09-09
Bug fixes for Service Worker 

### LINKED ISSUE #5920: Service Worker crashes on iOS Home Screen app
**Describe the bug**
When Nightscout is added as a Home Screen app on iOS, the app randomly crashes on load. I can see the screen show service worker fetch errors for about a second, after which the app starts with the progress indicator missing, but never succeeds in the launch, indicating all files from the service worker cache fail to load. On killing the app and reloading, it succeeds in starting up, but frequently the alarm MP3 fails to load until the app is killed and reloaded the second time.

Possibly related, some desktop users are having issues with the alarms not sounding, with an unknown cause.

Sadly catching this in Safari debugger seems to be impossible, as attaching to the related safari process is possible only after the process has started and the crash occurs before the attachment happens, so no logs can be viewed on exactly what thee exception is.

**To Reproduce**
*not a 100% consistent repro*
1. On an iOS device, add your Nightscout website as a Home Screen app
2. Open the app
3. Go to other applications so the app is suspended
4. Go back to the NS home screen app so it reloads and you'll see the crash

**Expected behavior**
The site loads every time without the crash.

### LINKED ISSUE #5943: (dev) reload loop in firefox

pazaan:
* Fixes #5920
* Fixes #5943


pazaan:
Closing and re-opening based on https://github.com/nightscout/cgm-remote-monitor/pull/5964

PR #5980, opened 2020-09-12
Load absolute dosing from amount field

sulkaharo:
Alternate implementation for #5740


bewest (approved):
LGTM! Thanks.  :ship:

PR #6056, opened 2020-09-20
Refactor in-memory caching

sulkaharo:
* Now supports cache updates and flushing from persistence layer, so treatment updates and removes are immediately reflected in runtime state
* Cleaner separation of concerns between data loader and caching


bewest:
This looks like it extends the previous effort all the way to the storage layer for each collection.

Do we have any idea which apps or how many people would be affected by an app with degraded performance due to rejected entries?  Arguably "none" since that data is essentially orphaned.

sulkaharo:
Merged given thumbs up from multiple testers, including AAPS users

Issue #6138, opened 2020-09-27
14.0.5 master could not get Loop data


bjornoleh:
I changed to 14.0.5 master today for two NS instances fed with Loop data and CGM via Dexcom Bridge. CGM data was displayed, but no new Loop data was shown in the main screen after updating, even after >20 minutes. The Loop pill (FreeAPS) reported the age of the old Loop data 

I did not yet check reports.

Reverting to Dev 14.0.3 made Loop data load as usual.

bjornoleh:
@sulkaharo  I changed one of the NS instances back to 14.0.5 master again, with the same result. Some Loop data are shown in Loop, as far as I can see this is temp basals and boluses, and the COB pill. The IOB is outdated.

![image](https://user-images.githubusercontent.com/63544115/94371773-420df600-00f9-11eb-85ee-6e315bd6a640.png)

bjornoleh:
It looks pretty much the same in the reports:

![image](https://user-images.githubusercontent.com/63544115/94371837-af218b80-00f9-11eb-87cb-851ca49bc5bd.png)

bjornoleh:
Reverting to 14.0.4 Master: IOB and Loop/FreeAPS data are again up to date

![image](https://user-images.githubusercontent.com/63544115/94372110-6bc81c80-00fb-11eb-8a44-f4a1ca12f29e.png)

sulkaharo:
That's very curious. Is there anything in Nightscout logs that acts as a hint on what's going on? None of the changes in the update are Loop specific, all unit tests work and I'm seeing data updates to all collections going through as expected.

bjornoleh:
Which logs should I look at?

søn. 27. sep. 2020, 20:07 skrev Sulka Haro <notifications@github.com>:

sulkaharo:
The one Nightscout outputs - if you're on Heroku, check Papertrail

bjornoleh:
There are no new entries in devicestatur collection (mLab).

Yes, I'm on Heroku. In Papertrail there is this, some problem with the recent change with /pebble possibly? We have set up both NS instances DEVICESTATUS: denied, could there be some issue with the tokens?

Sep 27 12:02:16 xxxx app/web.1 **DENIED**:   158.xxx.xxx.63 no-token api:pebble,entries:read
[...]

PR #6147, opened 2020-09-28
Support uploading device statuses in batches

sulkaharo:
Fixes uploading of device status records in batches


sulkaharo:
Should fix #6138

sulkaharo:
Ok @bjornoleh validated it's working

Issue #5632, opened 2020-04-24
Alexa plugin error


apocsantos:
**Describe the bug**
Alexa plugin does not work, it keeps returning "i'm not sure about that"

**To Reproduce**
1. Go to 'test in alexa skills'
2. Click on 'type: "ask Nightscout how am I doing" '
3. returns "i'm not sure about that"

**Expected behavior**
reply with my current nightscout blood glucose level.

**Your setup information**
- What version of Nightscout Updated this week
- What type of CGM, and how do you get your data there? (miaomiao)

**Additional context**

inventor96:
Both Amazon Alexa and Google Assistant don't play well with the name when we use the one-word version "Nightscout". Try separating it, like "night scout" both in your invocation name, as well as in what you type in the testing page.

apocsantos:
Hi,
Thank you!

Tried that but no luck it replies: "There was a problem with the requested skill's response" to the question :  "ask night scout how am I doing".

Here is my Json Skill Code


`{
    "interactionModel": {
        "languageModel": {
            "invocationName": "night scout",
            "intents": [
                {
                    "name": "NSStatus",
                    "slots": [],
                    "samples": [
                        "How am I doing"
                    ]
                },
                {
                    "name": "LastLoop",
                    "slots": [],
                    "samples": [
                        "When was my last loop"
                    ]
                },
                {
                    "name": "MetricNow",
                    "slots": [
                        {
                            "name": "metric",
                            "type": "LIST_OF_METRICS",
                            "samples": [
                                "what {pwd} {metric} is",
                                "what my {metric} is",
                                "how {pwd} {metric} is",
                                "how my {metric} is",
                                "how much {metric} does {pwd} have",
                                "how much {metric} I have",
                                "how much {metric}",
                                "{pwd} {metric}",
                                "{metric}",
                                "my {metric}"
                            ]
                        },
                        {
                            "name": "pwd",
                            "type": "AMAZON.US_FIRST_NAME"
                        }
                    ],
                    "samples": [
                        "how much {metric} does {pwd} have left",
                        "what's {metric}",
                        "what's my {metric}",
                        "how much {metric} is left",
                        "what's {pwd} {metric}",
                        "how much {metric}",
                        "how is {metric}",
                        "how is my {metric}",
                        "how is {pwd} {metric}",
                        "how my {metric} is",
                        "what is {metric}",
                        "how much {metric} do I have",
                        "how much {metric} does {pwd} have",
[...]

PR #6196, opened 2020-10-05
Use lodash.get() in virtual assistant API endpoints

inventor96:
When troubleshooting issue #5632, I found that the webhook from Alexa doesn't always include the user's language, so I've migrated a couple things to use `lodash.get()` to eliminate unnecessary errors and to simplify the code that finds slots.


sulkaharo:
Looks like you based this on master? Can you merge to DEV and re-submit without the change to README, so this won't conflict later when we merge dev to master

inventor96:
Closed in favor of #6199

PR #6199, opened 2020-10-05
Use lodash.get() in virtual assistant API endpoints (v2)

inventor96:
When troubleshooting issue #5632, I found that the webhook from Alexa doesn't always include the user's language, so I've migrated a couple things to use lodash.get() to eliminate unnecessary errors and to simplify the code that finds slots.


inventor96:
@sulkaharo Not sure why the tests failed on this one... I look through the test log, and the only thing that looked like an issue to me was on line 1647 (https://github.com/nightscout/cgm-remote-monitor/runs/1245569388?check_suite_focus=true#step:7:1647), which is not related to the changes in this PR. It almost looks like there was a connection issue within the test VM...? Can we request the tests to be re-run to see if it persists?

inventor96:
@sulkaharo I circled back to this and realized that I totally didn't see the very clear description of what went wrong in the tests... 🤦  My bad. Now that I saw that, I fixed the issue, and this PR is ready to go.

Issue #5991, opened 2020-09-14
Translation issues (Some text is not translated)


smilodonis:
**Describe the bug**
The Target BG range bottom text is not translated even if the translation is ready and present in the language.js file. Also the loopalyzer tap in report has some text that is not in the language.js file.

**To Reproduce**
1. Opan any nightscout site
2. Change the language to anything else then english
3. Go to reports
4. Under the checkboxes where you can select the individual days there is a text : Target BG range bottom. That should be tranlsated properly.

**Expected behavior**
The text should be translated

**Screenshots**
![Capture](https://user-images.githubusercontent.com/56167205/93092726-1a05a800-f6a0-11ea-9c56-9c98c9420874.PNG)
 
**Your setup information**
- What version of Nightscout: version 13.1.0
- What type of CGM, and how do you get your data there? (e.g. G4 and ShareBridge, or wired receiver, etc.)
I have a 640g uploader to NS directly
- Is your issue specific to a browser (Firefox/Safari/Chrome?) or a device (Android phone, etc.)?
Tested only on chrome

**Additional context**

Laga:
Looks like this was resolved? If so, this issue could probably be closed. Thanks!

PR #6217, opened 2020-10-08
#5991 Typo in translation

mg80:


mg80:
Typo in language key for Target **BG** range bottom - easiest thing to do is fix the key. Considered using the case insensitivity option on the translate function [here](https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/language.js#L15190), but I'm new to this project and not sure if the case sensitivity translation is intentional.

PR #6248, opened 2020-10-14
Fix batch

sulkaharo:
* Use the delta plugin data to show the delta in the clocks
* Fix disabling the BG alarms
* Load battery and other rare events up to two months back


ville-juhani-honka:
I deployed to my NS app the parts of the batch (ar2.js and simplealarms.js) that have an impact to the alarm behaviour and did some testing yesterday. I found that:
1) For "simple" alarm type, now ALARM_HIGH alarm is indeed disabled, if it's disabled in the config variables.
2) However, it doesn't seem to be possible to enable the high alarm anymore on browser/app level, if ALARM_HIGH is "off" in the vars
3) While the above ones apply to "simple" alarm type, it does not seem to apply for "predict"; even if I had ALARM_HIGH off in the vars (and also on the app level in browser), I was receiving a prediction-based high alarm, if I enabled

My wife's BG was yesterday in a good balance, so I did not check the behaviour with other limits. I might do that as well by changing them in the vars, but maybe already the above findings help further.

@sulkaharo

PR #6269, opened 2020-10-19
merge19102020

heliax01:


sulkaharo:
Hi! Oops - it seems this pull request was made the wrong way round - you're trying to pull changes from your copy of Nightscout into the master version, when the correct direction is from master into your copy.

Don't worry about this, no harm done! The message is here just to help you progress in the update in case you couldn't complete the update yet. Please try using the Nightscout update tool again and pay special attention to the source (Nightscout master) and target (your copy) merge direction settings. We'll close this pull request, as it's now in the Nightscout repository instead of the intended your personal copy.

PR #6311, opened 2020-10-27
Sanitize extendedClientSettings

jcantara:
## What

Sanitize extendedClientSettings to not include the private APNS (Apple Push Notification Service) key, and other related values (that are less sensitive, but still should not be shared: APNS Key Id, and Developer Team ID).

## Why

The APNS key is used for loop "Remote Override" functionality, where Nightscout can send an Apple push notification to the loop app running on the looper's iPhone to enable or disable a temporary override remotely. 

This private key is currently being exposed to the public when someone has configured Nightscout to use this remote overrides loop setting.

This private key could be used to craft a remote override command to any looper's iPhone that is exposing this key, allowing any random person to change the insulin delivery settings at will. This is a **_very bad thing_**. 

Here is a (anonymized) screenshot demonstrating the key found with a simple google search:

leaked_pkey

## Testing

I have written a unit test to check that the private key is not shared.

I have been running this code live for quite some time now, with no unintentional regressions or side-effects. 

## Remediation

All Nightscout users that have configured remote overrides should (after updating their nightscout site) immediately replace the APNS key with a new one, and delete the old key.

### DIFF
--- a/lib/plugins/index.js
+++ b/lib/plugins/index.js
@@ -216,7 +216,15 @@ function init (ctx) {
   plugins.extendedClientSettings = function extendedClientSettings (allExtendedSettings) {
     var clientSettings = {};
     _each(clientDefaultPlugins, function eachClientPlugin (plugin) {
-      clientSettings[plugin.name] = allExtendedSettings[plugin.name];
+      var settings = allExtendedSettings[plugin.name];
+      if(settings) {
+        // sanitize private settings
+        var settingscopy = JSON.parse(JSON.stringify(settings));
+        delete settingscopy.apnsKey;
+        delete settingscopy.apnsKeyId;
+        delete settingscopy.developerTeamId;
+        clientSettings[plugin.name] = settingscopy;
+      }
     });
 
     //HACK:  include devicestatus
--- a/tests/api.status.sanitized.test.js
+++ b/tests/api.status.sanitized.test.js
@@ -1,0 +1,52 @@
+'use strict';
+
+var request = require('supertest');
+var language = require('../lib/language')();
+
+require('should');
+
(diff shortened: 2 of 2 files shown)


jcantara:
The CI test check doesn't appear to have run the unit test I added; not immediately obvious to me why it wouldn't have.

lib/plugins/index.js
```diff
@@ -216,7 +216,15 @@ function init (ctx) {
   plugins.extendedClientSettings = function extendedClientSettings (allExtendedSettings) {
     var clientSettings = {};
     _each(clientDefaultPlugins, function eachClientPlugin (plugin) {
-      clientSettings[plugin.name] = allExtendedSettings[plugin.name];
+      var settings = allExtendedSettings[plugin.name];
+      if(settings) {
+        // sanitize private settings
+        var settingscopy = JSON.parse(JSON.stringify(settings));
```
sulkaharo, line 222: This repository uses lodash clone for this, see _.clone() or _.cloneDeep()

lib/plugins/index.js
```diff
@@ -216,7 +216,15 @@ function init (ctx) {
...
+        delete settingscopy.apnsKey;
```
sulkaharo, line 223: Rather than hardcoding the settings here, we should extend on filteredSettings() in settings.js and probably implement the filtering in settings.js, which is a better domain for filtering than the plugins.

sulkaharo: Actually looks like the only thing that's needed here is adding passing allExtendedSettings through settings.filteredSettings() in the beginning of the function as the rest of this is already implemented there

jcantara:
already addressed in release 14, my mistake

PR #6399, opened 2020-11-09
Update broken Markdown links in README

cascer1:
Some links in de README used square brackets instead of roud brackets. This commit replaces them with the correct brackets so that the links become clickable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,15 +195,15 @@ Want to help with development, or just see how Nightscout works? Great! See [CON
 
 # Usage
 
-The data being uploaded from the server to the client is from a MongoDB server such as [MongoDB Atlas][https://www.mongodb.com].
+The data being uploaded from the server to the client is from a MongoDB server such as [MongoDB Atlas](https://www.mongodb.com).
 
 [autoconfigure]: https://nightscout.github.io/pages/configure/
 [mongostring]: https://nightscout.github.io/pages/mongostring/
 
 ## Updating my version?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,15 +195,15 @@ Want to help with development, or just see how Nightscout works? Great! See [CON
...
 
 [autoconfigure]: https://nightscout.github.io/pages/configure/
 [mongostring]: https://nightscout.github.io/pages/mongostring/
 
 ## Updating my version?
 
-The easiest way to update your version of cgm-remote-monitor to the latest version is to use the [update tool][update-fork]. A step-by-step guide is available [here][http://www.nightscout.info/wiki/welcome/how-to-update-to-latest-cgm-remote-monitor-aka-cookie].
-To downgrade to an older version, follow [this guide][http://www.nightscout.info/wiki/welcome/how-to-deploy-an-older-version-of-nightscout].
+The easiest way to update your version of cgm-remote-monitor to the latest version is to use the [update tool][update-fork]. A step-by-step guide is available [here](http://www.nightscout.info/wiki/welcome/how-to-update-to-latest-cgm-remote-monitor-aka-cookie).
+To downgrade to an older version, follow [this guide](http://www.nightscout.info/wiki/welcome/how-to-deploy-an-older-version-of-nightscout).
```
sulkaharo, line 206: The link doesn't work anymore...

cascer1: hmm I didn't check which links were alive. I looked for it and can't find a page which describes that content on another url either. I could remove it or use an internet archive link?

sulkaharo: The best would be to PR the content that used to be there to the new docs in github.io ;)

PR #6445, opened 2020-11-11
Release 14.0.7

### LINKED ISSUE #6260: App crashed on heroku after deplay new build 14.0.6
**Describe the bug**
I've run nightscout on heroku with mLab. But I'm not sure which version I've got before. After updating to the latest version and deploy it to heroku the app crashed with the following message in the log:
`nightscout at=error code=h10 desc="app crashed" method=get path="/socket.io/?eio=3`

After restarting the app on heroku I've got the following message:
`error: helmet no longer supports `true` as a middleware option.`

**Your setup information**
- What version of Nightscout 14.0.6
- AndroidAPS
- xDrip
- This issue is not specific to a browser (Firefox/Safari/Chrome?) or a device (Android phone, etc.)

**Additional context**
As a workaround I switch the config var SECURE_HSTS_HEADER from
true
to 
false.

After that is works.

Sorry that i cannot give more information as I do not have save the logs. Perhaps you can instruct me how to get them from heroku.

sulkaharo:
* Test if database is in read only mode when Nightscout starts and give an error if read only mode is detected
* The client now checks if the server has loaded the initial data and is ready to server the client before letting user into the main UI
* Google Home and Amazon Alexa fixes - users who use either probably should update their Nightscout
* Fixed Pushover crashing Nightscout if Pushover servers are returning an internal server error
* Fix swagger for APIv3
* Language updates
* Use the delta plugin data to show the delta in the clock views
* Update Node version checks during startup to refer to Node 10 and 12 & allow any LTS version
* Fix disabling the BG alarms for simple alarms
* Load battery and other rare events up to two months back
* Possibly fixes compatibility with ios9 - needs testing
* Unified black and color clock layouts
* Update clock data every 20 seconds
* Update clock time every second
* Fix how CSP policy is set for Helmet, fixes #6260
* Authorization fix for misformatted URLs
* Added unit test for batch upload of CGM entries
* Improved / removed some logging


inventor96:
Not sure if it's an issue or not, but when I first navigated to my site after deploying the updated dev branch, I got the following message: `Redirecting you to the Profile Editor to create a new profile.` It looks like it was a JS `alert()` call, not an HTML modal or anything. I copied the text to write this post, then when I went back, the message was gone and the profile editor was loaded. It said I was still admin authorized, and all the values looked normal to me. I didn't do anything with the page, and just clicked again on the link to my site on my bookmarks bar in the browser. Everything appeared to load up just fine.

I have a second NS site pointed at the same Mongo DB (config issue between OpenAPS and xDrip+, so that's my workaround). When I load it, I did not get the same message. It did an initial load, then one reload, and then everything was just fine. So probably a one-off deal, but figured I'd report it, just in case.

sulkaharo:
@inventor96 Right - this seems to be a timing issue, where if you load the site before the backend has got the full dataset loaded, it'll show the dialog. I'll check if there's a simple workaround to enable the site to not load the UI before the backend is hot.

sulkaharo:
Ok I added a new feature, where the client now checks if the server is ready to actually serve the client - if the server is still in process of booting up and loading data from Mongo, the client waits until the load is complete.

lib/client/clock-client.js
```diff
@@ -150,7 +122,7 @@ client.render = function render (xhr) {
     let bgTargetBottom = client.settings.thresholds.bgTargetBottom;
     let bgTargetTop = client.settings.thresholds.bgTargetTop;
 
-    let bgNum = parseFloat(rec.sgv);
+    let bgNum = parseFloat(rec.mgdl);
```
unsoluble, line 125: Why was this changed? It breaks for anyone using mmol/L in their core settings.

sulkaharo: The BG targets are converted to mgdl for all users for Nightscout's logic. If you check what the target values this gets compared to, the values are in mgdl even for mmol users. Have you observed issues or did you comment this just from reading code?

unsoluble: It's hard to trace the logic here, but yes, something in this commit is a breaking change when your thresholds are set in mmol. Previously the clock colours were correct, now they are not. If I change all my thresholds in Config Vars to mg/dl equivalents, the colours are correct again.

unsoluble: Actually, it appears to be more complex than that — with my thresholds as mg/dl values, the colours were correct within a certain range (when the kid was lower), but now that he's past BG_TARGET_TOP it's not rendering the correct colour no matter which units I use. But it's correct in the main display:

Screen Shot 2020-11-29 at 11 41 59 PM
Screen Shot 2020-11-29 at 11 41 50 PM

sulkaharo: Right, I found a bug in the init order of the clock client (which has been there for a long time), so looks like the clock has been using the target defaults instead of the configured targets for some time now. Fixing. (The changes to the clock were made to have the clock use the same math as the rest of Nightscout, so things like the shown BG delta would match the rest of the system - with the custom delta calculation in the clocks, the shown values were almost but not the same.)

unsoluble: Gotcha. I was a few versions back before today, missed when that change happened I suppose. Thanks for finding the fix. :)

PR #6457, opened 2020-11-12
Alexa Intent Updates

inventor96:
Amazon has required new built-in intents.

### DIFF
--- a/docs/plugins/alexa-templates/en-us.json
+++ b/docs/plugins/alexa-templates/en-us.json
@@ -68,6 +68,14 @@
                 {
                     "name": "AMAZON.StopIntent",
                     "samples": []
+                },
+                {
+                    "name": "AMAZON.CancelIntent",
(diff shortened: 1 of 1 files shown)


inventor96:
@sulkaharo Unless new users to the Alexa integration know how to add intents, they will be unable to complete the setup. Is it possible to push this one through pretty quick? The changes were generated by Amazon's dev tools for Alexa, so don't think we need to worry about testing it much.

sulkaharo:
Just to confirm, do people who have an existing setup need to update to this, or does this only affect new users?

inventor96:
Umm... I believe existing setups will continue to work for now. I can't imagine Amazon stopping all existing skills because of a new requirement. They might possibly make a big stink about it in the future, though.

I came across this issue because I was helping a friend setup his skill for this, so I can confirm that it will for sure be a requirement for new users.

PR #6540, opened 2020-11-17
Update CONTRIBUTING.md

bjornoleh:
regarding translation


bjornoleh:
Hi @sulkaharo , just a ping before this drowns in all the misplaced PRs. Let me know if it is better to put some minor issue like this in an issue report instead.

PR #6559, opened 2020-11-21
 Added support for multiple uploaders in virtual assistant delta calculation

inventor96:


sulkaharo:
Should we add support to entries or some other place to actually make this generic and easy to call from every plugin?

inventor96:
Where does the delta display get the value from? It would probably be better if I just used the same source as that.

sulkaharo:
Right, if you just want to load the current delta, that's contained in the SBX object the plugin gets as call. You might want to console.log() the sbx after data is loaded and see what the bgnow plugin has offered into the sandbox.

inventor96:
As I was looking into it, I remembered I went with this implementation so I could get the time between each value. When I was first drafting the code to provide the delta to virtual assistants, I went with the value from `sbx`, but that's just the value; no timestamps are attached. With that in mind, personally, I don't know that I want to create another function when the delta features in the `bgnow` plugin would suffice for 95% of the needs going forward. If you still think it would be useful, I can totally move it, but that's my 2 cents.

sulkaharo:
Note the delta calculation landing into sbx also takes into account the time delta in case data is missing. If you want to have the time delta available as well, that sounds like a great little addition to the bgnow plugin?

inventor96:
Ahh... Fair point. Yeah, I'll move the functionality to the `bgnow` plugin.

inventor96:
OK, I moved it over to the `bgnow` plugin and added handling for interpolation.

translations/bg_BG.json
```diff
@@ -613,6 +613,7 @@
     "virtAsstCGMBattOne": "Your CGM battery was %1 volts as of %2.",
     "virtAsstCGMBattTwo": "Your CGM battery levels were %1 volts and %2 volts as of %3.",
     "virtAsstDelta": "Your delta was %1 between %2 and %3.",
+    "virtAsstDeltaEstimated": "Your estimated delta was %1 between %2 and %3.",
```
sulkaharo, line 616: Can you remove this file from the PR? The language files are changed through Crowdin and pull requests are are not expected to change these files.

sulkaharo: If you want a new key in the system, add that to language/en/en.json

inventor96: Got it. These have been removed and merge conflicts resolved.

PR #6622, opened 2020-12-06
Security improvement batch

sulkaharo:
Adds a new method for the server to push notifies to the client, which require administration privileges from the user. If there are messages in queue but user is not privileged, she is notified of pending messages.

* The messages are aggregated if repeated
* Non-persistent messages are cleaned from queue every 12 hours
* An admin message is generated on auth fail
* You can send admin messages over the bus

*XSS protections*

* Objects sent over the REST API V1 are now filtered for XSS injection code

TODO:
* Check translations
* Add more messages?


lib/client/adminnotifiesclient.js
```diff
@@ -0,0 +1,83 @@
...
+        var m = messages[i];
+        html += wrapmessage(m.title, m.message);
+      }
+    } else {
+      if (messageCount > 0) {
+        html = wrapmessage('Admin messages in queue', 'Please sign in using the API_SECRET to see your administration messages');
+      } else {
+        html = wrapmessage('Queue empty', 'There are no notifies in queue');
+      }
+    }
```
pazaan, line 60:
```suggestion
    } else if (messageCount > 0) {
      html = wrapmessage('Admin messages in queue', 'Please sign in using the API_SECRET to see your administration messages');
    } else {
      html = wrapmessage('Queue empty', 'There are no notifies in queue');
    }
```

pazaan (commented):
A little code clarity suggestion.

Also wondering if using `notice`/`notices` makes more sense than `notify`/`notifies`.

PR #6688, opened 2020-12-27
APIv3: Cache invalidation + refactoring

PetrOndrusek:
Recently a caching mechanism has been added into the project and since apiv3 also modifies the data, it is responsible for properly invalidating the cache. 
This PR solves that by adding a new storage implementation `mongoCachedCollection` which wraps the original mongo DB storage `mongoCollection` and invalidates the cache when needed.
Furthermore, all APIv3 tests were revised and added to check for proper cache invalidation.
At the same time, incorrect async calls in tests were fixed.


sulkaharo:
Ooh love the way to you added the storage abstraction layer <3

PR #6718, opened 2021-01-07
 Don't run docker push on forks

StephenBrown2:
I noticed that I got an error in CI after updating to the latest release, so looked into restricting the docker push steps to not run unless the repo owner is 'nightscout'.

This should do the trick, as evidenced by the latest run on my fork: https://github.com/StephenBrown2/cgm-remote-monitor/runs/1663926897

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
     name: Publish dev branch to Docker Hub
     needs: test
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/dev'
+    if: github.ref == 'refs/heads/dev' && github.repository_owner == 'nightscout'
     env:
       DOCKER_IMAGE: nightscout/cgm-remote-monitor
     steps:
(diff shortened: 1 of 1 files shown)


sulkaharo:
Ping @pazaan

sulkaharo:
Looks like this would make sense for an exception where we merge to master now and then merge this from master to dev

StephenBrown2:
I'm not sure you would need an exception, as it would still run just as before _in this repository_, since it would be checking that it's being run from the `nightscout` org, which is true here.

sulkaharo:
I meant process exception, where we usually merge all master changes through dev, but in this case the other direction would make sense.

StephenBrown2:
Right, that's what I understood. I'm just not sure what the reason for going through master first in this case would be.

sulkaharo:
Well you issued the PR against master, right? ;) The reason in this case would be, DEV already has content for the next release, so if we merge to dev, this will be merged to master much later, and given this doesn't change any of the app logic, it should be safe to merge to master without further testing.

StephenBrown2:
> Well you issued the PR against master, right? ;)

🤦🏻 Yes, that would be me mistake. I just changed the base, it should have been against dev. I'd say moving it through the normal dev process should be fine, and then with the next point release it'll get merged to master and builds won't be attempted on forks after that.

> given this doesn't change any of the app logic, it should be safe to merge to master without further testing.

Up to y'all, feel free to work is as you wish.

pazaan:
Yep, it looks like a sensible change to me 😄

pazaan (approved):

PR #6755, opened 2021-01-14
Plugin for Medtronic Guardian Connect CGM

szymjaw:
It's a plugin for Medtronic Guardian Connect CGM.

I changed _"lib/plugins/pluginbase.js"_ for manage colors in tooltip
GC tooltip

Now you can use warnBackground or urgentBackground as classname  :
`      info.push({
        label: 'Label text: '
        , value: 'Some value'
        , class: 'classname '
      });`


sulkaharo:
Initial comments: 1) you need to remove the polish JSON file changes. All the localisations in that folder come from Crowdin and they will have merge issues if we manually edit those files. 2) Please add the README changes to document this to users. Has someone tested this in addition to yourself?

sulkaharo:
Hi! One more small change - you now put in a commit that deletes the pl_PL file from the repo when this is merged. The desired PR would not touch that file at all, so what you'd want to do is reset the state of the file for the PR so it's not touched when this is merged.

szymjaw:
I did, what you wrote in the previous message in point 1.

szymjaw:
When I create plugin, I can't push the translation for this plugin in the same PR ?

sulkaharo:
@szymjaw You can and should push the change the English source file in translations/en/en.json, but the actual translations are made in Crowdin, which loads that file as the source of what localisation keys exist in the system. None of the localisation happens over "hand made" git pull requests anymore.

sulkaharo:
> I did, what you wrote in the previous message in point 1.

I asked you to remove the changes in the Polish file, not to commit a file deletion, which are very different operation in Git

PR #6760, opened 2021-01-14
##6523 Typo in translation

mg80:


sulkaharo:
If you wan to change the capitalisation, you should change the English content file in /translations/en/en.json, not the source that pulls in the localisation data, as this would invalidate all the existing localisation keys.

mg80:
@sulkaharo is there a reason that translations are case sensitive? I'm not aware of a language that translates differently based on case. It seems like we could simplify things by always being case insensitive when looking up the key. Am I overlooking something?

PR #6776, opened 2021-01-20
Round insulin value in treatment tooltip

bassettb:
Fix occasional issue where insulin value has 10+ decimal places in the treatment tooltip.
Also renamed the helper method to be more clear on its behavior.  View commit diffs separately.
Screen Shot 2021-01-18 at 5 16 03 PM


lib/client/renderer.js
```diff
@@ -630,7 +630,7 @@ function init (client, d3) {
       var glucose = treatment.glucose;
       if (client.settings.units != client.ddata.profile.getUnits()) {
         glucose *= (client.settings.units === 'mmol' ? (1 / consts.MMOL_TO_MGDL) : consts.MMOL_TO_MGDL);
-        var decimals = (client.settings.units === 'mmol' ? 10 : 1);
+        const decimals = (client.settings.units === 'mmol' ? 10 : 1);
 
         glucose = Math.round(glucose * decimals) / decimals;
```
sulkaharo, line 635: Why not refactor this to use the function as well?

PR #6777, opened 2021-01-20
Catch error from websocket device status dedupe

sulkaharo:


lib/server/websocket.js
```diff
@@ -408,12 +408,18 @@ function init (env, ctx, server) {
...
-              console.log(LOG_DEDUP + 'Devicestatus exact match');
-              if (callback) {
-                callback([array[0]]);
-              }
+            if (err) {
+              console.error(err);
+              callback([]);
               return;
+            } else {
+              if (array.length > 0) {
```
bassettb, line 416: "else if" instead?

Issue #6775, opened 2021-01-20
Nightscout crash: "MongoDB connection failed! Double check the MONGODB_URI setting in Heroku"


philipgo:
**Describe the bug**
Every day or so my Nightscout crashes with the error message "MongoDB connection failed! Double check the MONGODB_URI setting in Heroku." 

The problem appears to be with Nightscout (or AndroidAPS??), I do not need to restart the MongoDB instance and other Nightscout instances continue to work with the same MONGODB_URI without any issues. However I do need to manually restart the Nightscout instance. 

**Expected behavior**
Nightscout should not crash or if it does it should recover automatically.

**Screenshots**
![NS crash](https://user-images.githubusercontent.com/29184757/105187314-ee556b80-5b32-11eb-8a27-100112935570.jpg)

**Your setup information**
MongoDB self hosted on a private server, Nightscout Docker (latest dev, but same issue on master) hosted on a different private server. My uploader is AndroidAPS 2.8.1.1

**Logs**
```
nightscout_1  | DEDUPE: Devicestatus exact match
nightscout_1  | /opt/app/node_modules/mongodb/lib/utils.js:668
nightscout_1  |           throw error;
nightscout_1  |           ^
nightscout_1  |
nightscout_1  | TypeError: Cannot read property '0' of undefined
nightscout_1  |     at findResult (/opt/app/lib/server/websocket.js:414:32)
nightscout_1  |     at /opt/app/node_modules/mongodb/lib/utils.js:665:9
nightscout_1  |     at handleCallback (/opt/app/node_modules/mongodb/lib/utils.j                                                                                                                                                             s:102:55)
nightscout_1  |     at /opt/app/node_modules/mongodb/lib/cursor.js:834:20
nightscout_1  |     at /opt/app/node_modules/mongodb/lib/core/cursor.js:736:9
nightscout_1  |     at Object.callback (/opt/app/node_modules/mongodb/lib/core/c                                                                                                                                                             ursor.js:432:11)
nightscout_1  |     at Timeout._onTimeout (/opt/app/node_modules/mongodb/lib/cor                                                                                                                                                             e/sdam/topology.js:443:25)
nightscout_1  |     at listOnTimeout (internal/timers.js:554:17)
nightscout_1  |     at processTimers (internal/timers.js:497:7)
nightscout_1  | Executing startBoot
nightscout_1  | Executing checkNodeVersion
nightscout_1  | Node LTS version v14.15.3 is supported
[...]

PR #6811, opened 2021-01-28
Mongo default timeout

philipgo:
- removed the short time outs so MongoDB can revert to the longer default time outs instead (socketTimeoutMS = 360000 and connectTimeoutMS=10000) as discussed [here](https://github.com/nightscout/cgm-remote-monitor/issues/6775)
- removed MongoDB_URI (including user name and password) from public error message
- fixed typos


sulkaharo:
Did you test the behavior with a wrong URI with the timeouts entirely removed? I suspect if we have the connect timeout that long, it might cause issues with users not detecting the error during install

philipgo:
> Did you test the behavior with a wrong URI with the timeouts entirely removed? 
Thank you, I have tested this case now. Works fine with wrong URL, wrong user name or wrong password. This should be determined by connectTimeoutMS (which is 10s now and thus unchanged).

According to https://docs.mongodb.com/drivers/node/faq socketTimeoutMS "applies only to sockets that have already been connected"

Issue #6701, opened 2021-01-02
Remember selections in Day to day report (IOB, COB etc)


bjornoleh:
I look at the Day to day report a lot, and always want to see IOB and COB. I wish selecting these would be remembered the next time. Alternatively they could be on as default. It would save a few clicks, and when viewing the report on a phone, I also have to zoom in a bit to manage to select IOB and COB. 

Thanks for considering this!

bjornoleh:
Thanks @sulkaharo and @mg80 , the current dev seems to work as I was hoping! At least for Day to day, havent checked other reports yet.

The Day to day graph size doesn't seem to save though. I believe it did before?

mg80:
@bjornoleh my change only saves these preferences:
![image](https://user-images.githubusercontent.com/6451350/106675019-1dbf9b80-6569-11eb-8b2a-1e21b66ea0fd.png)
@sulkaharo let me know your thoughts on saving other report preferences and I can include them.

sulkaharo:
"The Day to day graph size doesn't seem to save though. I believe it did before?" AFAIK this has not ever persisted, so if it has, that's probably been some behaviour of your browser, not Nightscout itself.

bjornoleh:
That is possible, in that case it is with Chrome (Win/Android). But I am unsure about this.

PR #6814, opened 2021-01-29
#6701 Report storage tests

mg80:
Functional and unit tests

### DIFF
--- a/tests/hashauth.test.js
+++ b/tests/hashauth.test.js
@@ -16,6 +16,8 @@ describe('hashauth', function ( ) {
   });
 
   after(function (done) {
+    // cleanup js-storage as it evaluates if the test is running in the window or not when first required
+    delete require.cache[require.resolve('js-storage')];
     done( );
   });
 
--- a/tests/reports.test.js
+++ b/tests/reports.test.js
@@ -340,5 +340,19 @@ describe('reports', function ( ) {
     });
     
   });
+
+    it ('Report preferences should stick when refreshing page', function (done) {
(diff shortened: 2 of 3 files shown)


mg80:
@sulkaharo I added some tests, but it looks like the build is broken due to an unrelated issue

sulkaharo:
Looks like the tests breakage is caused by some change in the tests in this PR interfering with the other tests - the DEV tests run on without these changes but this PR fails both in the cloud and locally. I suspect the issue is with the report storage initialising the local storage which is also used but other tests. I'm not sure I have the time immediately to check what's going on, so if you have time to check what's causing the test to fail and help implement a clean teardown between tests, that'd be great.

mg80:
@sulkaharo wow that was a nasty one. js-storage evaluates if it's running on the window or not as soon as it's required. This becomes a problem when js-storage is required anywhere in the AST before the window object is setup. I'm guessing this wasn't a problem until now as I added a pure unit test.

tests/hashauth.test.js
```diff
@@ -16,6 +16,8 @@ describe('hashauth', function ( ) {
   });
 
   after(function (done) {
+    // cleanup js-storage as it evaluates if the test is running in the window or not when first required
+    delete require.cache[require.resolve('js-storage')];
```
sulkaharo, line 20: 😍👍

sulkaharo:
GREAT catch! Makes me wonder if we should have a generic teardown include file that deletes the require cache for this and other core requires between all tests. 🤔

PR #6825, opened 2021-01-31
Fix Douple space _ English

parapenT1sta:


parapenT1sta:
@sulkaharo Is how it appens with xDrip+ and AndroidAPS :) fixing a typo makes the text to be translated again, but in fact the recomendation of the old text is shown.

But I understand your point of view and it's just a small typo.

PR #6844, opened 2021-02-04
Update webpack to v5

jpcunningh:
This is a work in progress.

Things to do:
* Still looking for more ways to reduce webpack size - suggestions welcome


sulkaharo:
Check https://github.com/nightscout/cgm-remote-monitor/commit/7d5b368283569e833efffd1dce0236263081bb91 for a missing CSS import that fixes reporting layout

New behaviour that I can't pinpoint: the packaging process creates tmp/public/images folder and generates the Nightscout logo image to the folder?

jpcunningh:
I believe the logo is put into the images folder to resolve the url in drawer.css pointing to logo2.png

Should I disable webpack from resolving the URLs in CSS?  If so, I would need to change the drawer.css URL back to its original reference.

PR #6855, opened 2021-02-09
Api3 remove Date header

PetrOndrusek:
Discussions with other developers have shown that the real security benefit of Date header is insignificant.
Therefore, this PR removes the obligation to attach a Date header (or now query parameter) and also removes mentions in the documentation.


sulkaharo:
LGTM. Test counts goes under 500 this! Eep! :D

PR #6863, opened 2021-02-11
Api3 remove auth token

PetrOndrusek:
After discussion with NS developers it turns out that it is not very safe to use accessToken for APIv3 calls.
So the only (reasonable) way to authenticate now is JWT.
This PR removes support to authenticate APIv3 calls by accessToken + updates documentation accordingly.


MilosKozak:
👍

sulkaharo:
Finally merging. Petr apologies for the very long delay.

PR #6867, opened 2021-02-14
Update nightscout

lmdeaton1:


lmdeaton1:
not sure why it thinks I want to merge back into master.

PR #6974, opened 2021-03-22
Fix Alexa partial translations

inventor96:
This loads the localization file on API calls to the Alexa and Google Home endpoints. Previously, the language was set, but never loaded.


sulkaharo:
Hey @inventor96 actually, can you check why the language object instance in sbx doesn't have the localisations loaded in the first place? There should be just one language object instance active in the runtime and looks like each of the plugin objects is getting a separate instance, so I'd prefer a fix where the initially instantiated copy that should have the localisation loaded is passed correctly to the plugins.

inventor96:
@sulkaharo As I've been thinking about this, I think we need to figure out the most appropriate direction to work in, because I didn't think of all the possibilities to start with.

While probably unlikely, it's technically possible for NS to be in one language and the virtual assistant to be in another. You make a good point about the language already being loaded. If the virtual assistant is in a different language, this PR (as it is) will probably change the language of the whole NS site. So with that in mind, should the virtual assistant responses automatically be in the same language as the site irrespective of the language used by the virtual assistant? Or should the virtual assistant responses respect the language of the virtual assistant?

If we were to go with the second option, we would either need to load the language at the beginning of the request, and return the site back to the config var language after the response is sent back; or create a separate language instance used by the virtual assistant functionality which would contain the language specified by the virtual assistant.

What are your thoughts?

inventor96:
@sulkaharo Could you share your thoughts on this?

fjpezuela:
Checked that in version 14.2.5 the same thing continues

cronicadiabetes:
Hello,

I've been trying alexa again in Spanish and it still continues to show some of the answers in English and some in Spanish. do you know what's going on?

imagen

Words like "and holding as of" "Your temp basal of xx units per hour will end" "and you have xx units of insulin on board", are shown in English in the answer.

Thank you very much

AndyLow91:
Thank you for this contribution, and apologies that the earlier discussion about the appropriate localization approach was left unresolved.
I plan to close this PR in its current form. The Google Assistant integration it targets has been retired, and the proposed fix still needs to address the shared-language concern discussed above: loading an assistant’s translations into the shared server language object can affect other server/plugin responses.
The Alexa mixed-language issue may still warrant a fix. If you feel this is worthwhile revisiting, please feel free to reopen and update the PR, ideally with isolated assistant localization and tests covering different site and assistant languages.

PR #7001, opened 2021-04-18
Loop settings in profile report

yodax:
firefox_iGUlpqG6lp

When looking at profiles when using Loop it's convenient to see the Loop specific settings in the report. This allows the team to evaluate the settings of the application remotely.


sulkaharo:
I'm hesitant to merge this in, since the Loop settings should be moved out from the profiles and any additional functionality that assumes the data is in the profiles adds friction to moving them out. The correct place would be to either use the device statuses or the settings API in V3 REST API.

yodax:
Fair enough :) I'll have a look at the Loop code, it's a bit in flux at the moment.

PR #7019, opened 2021-04-29
google home locale

pcastelo:
i  am adding spanish locale for google home.
Following the plug in instructions, I added the file in English to google actions  and then I translated it to spanish and later make a export from config.


inventor96:
Hi @pcastelo,

Before exporting, could you change the webhook URL in DialogFlow to be a generic test URL? Right now it looks like you have your personal Nightscout URL in there.

Also, it looks like there's a few areas that haven't been translated yet:
- Default Fallback Intent
- Default Welcome Intent
- The follow-up prompts for the "metric" parameter in the "MetricNow" intent

If you could update these, that would be great!

PR #7043, opened 2021-06-04
Build Docker image for arm64 architecture

### LINKED ISSUE #6932: Official Docker image for Nightscout on linux/arm64
I would like to use an official NS docker image on my RPI 4, but there is none so I can't and I don't like it.
Please add automated CI build for linux/arm64 (and maybe others as well?).

How do I assign it to @pazaan ? ;)

p5nbTgip0r:
This adds multi-platform Docker image building and publishing support to the CI via [`docker buildx`](https://docs.docker.com/buildx/working-with-buildx/). In addition, the Docker image will be built for `arm64` alongside `amd64`. For simplicity, I also merged the two `publish_dev` and `publish_master` jobs into one job named `publish`. 

Fixes #6932.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,12 @@ jobs:
     if: github.ref == 'refs/heads/dev' && github.repository_owner == 'nightscout'
     env:
       DOCKER_IMAGE: nightscout/cgm-remote-monitor
+      PLATFORMS: linux/amd64,linux/arm64
     steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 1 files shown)


pazaan:
Closes #6932. Thanks @p5nbTgip0r!

master3395:
Can this docker image also set up mongodb?
Because right now it only sets up Nightscout, and not the database.
Having the docker do both, makes it easier for VPS owners with docker to set up Nightscout :)

pazaan:
> Can this docker image also set up mongodb?
> 
> Because right now it only sets up Nightscout, and not the database.
> 
> Having the docker do both, makes it easier for VPS owners with docker to set up Nightscout :)

That's not what Docker is for. Containers are supposed to run discrete apps. 
There's a `docker-compose.yml` at the top level of this repo for running the full stack.

PR #7131, opened 2021-10-13
respect per route body-parser configuration

bewest:
This change allows each route to express policies for interpreting and parsing
the request body.  Before this change uploads to entries or treatments api
would error if they were larger than 100Kb due to the preference being set for
the whole server.  This change removes the global preference in favor of
allowing each route to choose it's own request size limit.

We also refactor usage of body-parser to be more consistent throughout the code
base.  Most routes can use jsonParser, rawPraser, and urlencodedParser provided
by the common wares component.  Anything doing something else should be called
out as such.  For example, treatments, activity, and entries allow uploads up
to 50Mb.  Other v1 endpoints are using the common configuration set to 1Mb.

### DIFF
--- a/lib/server/app.js
+++ b/lib/server/app.js
@@ -26,8 +26,6 @@ function create (env, ctx) {
   var appInfo = env.name + ' ' + env.version;
   app.set('title', appInfo);
   app.enable('trust proxy'); // Allows req.secure test on heroku https connections.
-  app.use(bodyParser.json());
-  app.use(bodyParser.urlencoded({ extended: true }));
   var insecureUseHttp = env.insecureUseHttp;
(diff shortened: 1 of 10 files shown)


lib/server/app.js
```diff
@@ -26,8 +26,6 @@ function create (env, ctx) {
   var appInfo = env.name + ' ' + env.version;
   app.set('title', appInfo);
   app.enable('trust proxy'); // Allows req.secure test on heroku https connections.
-  app.use(bodyParser.json());
-  app.use(bodyParser.urlencoded({ extended: true }));
```
bewest, line 30: These lines enforce a 100Kb limit for the entire server, regardless of per route configurations.  Removing these lines allows the routes to properly express their size limit policies.

PR #7154, opened 2021-10-31
Fix api-secret header name in swagger configuration.

MJDSys:
Without this change, the /api-docs/ endpoint would send the header "`api_secret`" instead of the expected "`api-secret`" header.


bewest:
Thanks for your input!  Including this in the next release.

PR #7158, opened 2021-11-02
Update Nightscout

Nadov:


Nita19 (commented):
Update nightscout

Nita19 (commented):
Update nightscout

PR #7207, opened 2021-12-02
Ensure that HIGH and LOW values are reported when using share2bridge

cpitchford:
Dexcom apps/receivers report HIGH for 401mg/dl and LOW for 39mg/dl

When using share2nightscout-bridge, the device is set as "share2" and this corrects the code so it can display HIGH and LOW for these limit values when received.

Without this, the high and low values are reported numerically and differently to the Dexcom app/receiver


bewest:
Thank you very much for this and the mmol enhancement!  Including for next release.

cpitchford:
Awesome! Thanks for merging. I’d like to get my local changes merged in so I can take up the request for testing. Sadly, I’ve been running an older version (without problem) and tweaking as I went. It’ll be good to get back to master/dev

PR #7231, opened 2021-12-10
Optimize time between polling share2nightscout-bridge to reduce ingest lag

cpitchford:
This updates the polling interval for the share2nightscout-bridge fetch routine.

# Current behaviour

Where as previously the share2nightscout-bridge is called at a regular interval to poll Dexcom for new data, we now use the data we received to schedule the next poll. With a default poll time of 2.5 minutes, depending on when the nightscout process is started, we could see up to a 2.5 minute lag between the new data being sent by the Dexcom app and it appearing in the nightscout feed.

# Optimisation

We know that Dexcom updates every 5 minutes. We also know that it takes under 20 seconds for the new data sent by the Dexcom App to appear in the Dexcom feed.

We look at the timestamp of the most recently collected data, add 5 minutes and 20 seconds to this time to find the next poll time.

If this time is in the past or if this poll time is more than the poll interval in the future, we revert to regular interval polling. This reverts to the previous behaviour whilst we await new data from Dexcom.

Once new data is provided, we attempt to re-schedule the next collection optimally.

For example:

  * The last data was collected at 12:00:00
  * The next poll should be 12:00:00 + 5 minutes + 20 seconds: 12:05:20
  * The time now is 12:04:00
  * The next poll should be in 1minutes 20 seconds. This is a reduction over the periodic polling interval of 2.5 minutes.

# Result expected

This update will reduce the random lag of up to 2.5 minutes to a reliable 20 seconds


bewest:
Thank you! I will see about getting this into the pending release.

cpitchford:
@bewest No urgency! I've moved my deployment to dev in order to try and help with testing.

I have been working on support for [LaMetric Time LED clocks](https://www.lametric.com/) devices. The endpoint displays the current trend/bg and 3 hours of history in a graph.

Lametric_Nightscout_demo

I've been running these features in my local copy for about 18 months. Raising PRs here and tracking against dev is going to get me up to speed for help with testing the dev branch.

rahim:
Love this addition, thank you!

khskekec:
Hey,

First of all i would like to thank you for this wonderful improvement. Everyone who deals with diabetes knows that time matters very often.

To be honest I am not very familiar with the technical architecture. Actually I am using the bridge to get values from follower and push it to nightscout. I thought that this is a totally separate application and does not have any dependencies to nightscout beside the Api.

I call the bridge as systemctl service and set its settings over env vars. Now I am confused that there is a strict integration into nightscout?! How does it work from technical perspective. Am I doing something wrong?

Best regards

Selçuk

cpitchford:
The bridge can be invoked directly, but nightscout typically treats it as a code module, invoking the methods to collect the Dexcom data on its own timer

Running as a stand alone process does a similar job, but the bridge manages the timing.

This PR I raised alters the timer part in nightscout (looking at the time stamp of the most recently collected reading to schedule the next collection 5mins after that) I didn't make the same change to the bridge's own timer code but it wouldn't be a stretch to implement it there too.


So you shouldn't need to run the bridge independently, though that would work (provided the bridge process knows how to post the data back to the nightscout process)

Let me know if I could help more

khskekec:
Thank you for your answer. Is there some kind of documentation which describes how to setup/configure and run this?

PR #7249, opened 2021-12-14
obscure private device provenance

bewest:
Allow users to choose whether or not to reveal the provenance of the data to
consumers.  This is intended to help protect data rights to portability and
access.  If consumers start abusing or discriminating against Nightscout users
due to the source of the data, this allows users to conceal this information.

### DIFF
--- a/lib/middleware/obscure-provenance.js
+++ b/lib/middleware/obscure-provenance.js
@@ -1,0 +1,12 @@
+
+module.exports = function create_device_obscurity (env) {
+  function obscure_device (req, res, next) {
+    if (res.entries && env.settings.obscureDeviceProvenance) {
+      for (var i = 0; i < res.entries.length; i++) {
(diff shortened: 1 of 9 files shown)


lib/middleware/obscure-provenance.js
```diff
@@ -0,0 +1,12 @@
+
+module.exports = function create_device_obscurity (env) {
+  function obscure_device (req, res, next) {
+    if (res.entries && env.settings.obscureDeviceProvenance) {
+      for (var i = 0; i < res.entries.length; i++) {
+        res.entries[i].device = env.settings.obscureDeviceProvenance;
```
sulkaharo, line 6: Does this change the in-memory cache data entries? Smells like this should make a copy of the data before modifying it

bewest: Good idea.

PR #7273, opened 2022-01-04
Fixes an issue where treatment units were all rewritten on load, regardless of units

sulkaharo:
Fixes an issue reported by AAPS dev team. The current behavior overwrites treatment units to mg/dl regardless of the original units and adds null targets on events missing the target. The fixed code changes the code to work as intended, where the units are only changed for temp target treatments.


bewest:
Looks good to me.

sulkaharo:
Fixes https://github.com/nightscout/AndroidAPS/pull/1094

PR #7285, opened 2022-01-07
Change deduping interval

MilosKozak:
Current behaviour of 1 min interval for deduplication is effectively blocking synchronization of 2 consequent events of the same type (like profile switch). I believe 2 sec should be enough for handling data with accuracy of seconds only

### DIFF
--- a/lib/server/websocket.js
+++ b/lib/server/websocket.js
@@ -294,7 +294,7 @@ function init (env, ctx, server) {
       socket.on('dbAdd', function dbAdd (data, callback) {
         console.log(LOG_WS + 'dbAdd client ID: ', socket.client.id, ' data: ', data);
         var collection = supportedCollections[data.collection];
-        var maxtimediff = times.mins(1).msecs;
+        var maxtimediff = times.secs(2).msecs;
 
         var check = checkConditions('dbAdd', data);
         if (check) {


bewest (approved):
LGTM, I'm willing to try this.

bewest:
For context on medtronic systems, `Bolus` records generally closely follow `BolusWizard` records.  BolusWizard records have information concerning meal announcements and bolus calculator recommendations.  However, only the `Bolus` record contains the amount of insulin actually requested as a dose at the completion of a bolus delivery, which occurs some time later.  The duration of this lag is generally controlled by the size of the bolus and the max rate of bolus delivery per pump and four minutes was chosen as a reasonable default.  Nightscout historically expects "meal events" to contain both insulin and carbs so that the bolus circle can also illustrate the carb to insulin ratio.  In order to deliver these events to Nightscout, it's necessary to munge the raw device stream to merge these events together.
The device stream of events is a log of operations while Nightscout historically wants a log of therapeutic decisions.

Here's some further discussion: 
* https://github.com/openaps/oref0/issues/443
* https://github.com/openaps/oref0/blob/dev/lib/bolus.js

It would be nice to gain some observability on the impact of these changes.

MilosKozak:
this constant is affecting only websocket comm. I believe currenly only AAPS is using this channel for data upload

sulkaharo:
FWIW I still think the websocket persistence code should be either fully removed or fully rewritten so the REST API and the websocket use the same semantics. We still have the live issue where AAPS users can have duplicate record issues due to NSClient using different data deduplication logic from the REST API. OTOH the 1 minute interval is also clearly wrong logic, so this won't hurt?

MilosKozak:
I agree. Next change will be to API v3

juzi:
Sorry for the naive question (I don't know much about your code): 
I've heard a lot about problems with synchronization and duplicate entries in the past, and I've always wondered why you don't introduce a UUID for each entry that is sent to Nightscout? Like that you could always distinguish two events with the same values. But maybe that is not relevant here.

PR #7311, opened 2022-01-26
Fix override

DriverTicks:
Replace apn with @parse/node-apn to fix loop override


sulkaharo:
It looks like the override broke due something being off in the Apple servers? Can you elaborate how this package change fixes the issue? Having said that, migrating the code to a maintained library would be good.

DriverTicks:
end

PR #7342, opened 2022-02-15
implement comprehensive glucose pentagon plugin

jgr-lab:


sulkaharo:
* mmol support seems to be missing
* I'm seeing PGR score: NaN on all days for no obvious reason

mushroom-dev:
I can't reproduce it. Are partial results correct? If so please share it for a few days where NaN occurs. They should be listed below PGR score, i.e.:

```
PGR score: 1.6

ToR: 326
CV: 27.3
Hypo: 2620.3
Hyper: 9580.1
Mean: 120
```

What time span did you define? All days radio button corresponds to all days between specified time range, can you narrow it down to a single day?

Can you share sgv/time data for reproduction of NaNs?

oddst:
It is sad that this work completely halted, as this is functionality that would have been very appreciated by the T1D community. However "Nightscout Reporter" has this, and links should be established in Nightscout to point to "Nightscout Reporter" as an excellent 3rd party report generator...!

PR #7380, opened 2022-03-21
Omnipod reservoir fixes

ps2:
This patch has two updates:
* Special handling of reservoir for Eros pods extended to all Insulet pods. 
* Allow uploaders to set urgency reservoir_level_override, when reservoir_display_override


bewest:
LGTM.

PR #7385, opened 2022-03-28
Support automatic bolus volumes in the enacted field for loop status.

ps2:
With automatic bolusing in Loop, the last temp basal adjustment might be quite old. The current loop plugin only shows last temp basal adjustment. This change updates the plugin to show automatic bolus enactments in the loop pill as well, so the user has a clearer understanding of the time that Loop last adjusted dosing.

This must be merged before https://github.com/ps2/NightscoutService/pull/17 can land in Loop.

### DIFF
--- a/lib/plugins/loop.js
+++ b/lib/plugins/loop.js
@@ -231,7 +231,7 @@ function init (ctx) {
         , split: false
         , targets: false
         , reasons: reasonconf
-        , otp: true 
+        , otp: true
         , submitHook: postLoopNotification
       },
       {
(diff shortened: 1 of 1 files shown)


ps2:
@bewest, let me know if there is anything I can do to get this merged, or if there are any concerns that are preventing it from being merged. Thanks! Also, https://github.com/nightscout/cgm-remote-monitor/pull/7380, and https://github.com/nightscout/cgm-remote-monitor/pull/7375 are waiting as well, and I think I've addressed all comments. Or  let me know if I should be talking to @MilosKozak or @sulkaharo, or someone else.

PR #7440, opened 2022-05-25
Fixes node 16

GermanBluefox:
I tested nightscout under node16 and it works fine. As node12 is out, you should take in count node 16.

I could not build GUI, so I added "vm-browserify" for asn-1 and Buffer browserfy modules. 

I suggest you to remove support of node10


PieterGit:
@GermanBluefox . I incorporated the typos you fixed in https://github.com/nightscout/cgm-remote-monitor/pull/7514 and that also fixes the Node16 updates. 

I don't see the use of adding vm-browserify support to Nightscout, and think we don't want to include extra default npm modules (that have not been maintained for 3 years). Is it ok for you to drop this, or is there an added value for it?

GermanBluefox:
> I don't see the use of adding vm-browserify support to Nightscout,

My porblem is, that I couldn't  build GUI without `vm-browserify`. But I have no Problem with excluding it, as far as YOU can build GUI.

PR #7465, opened 2022-06-25
Multi insulin api

gruoner:
Patch provides a new entity of type "insulin" and provided that via a new API (copied from food API).
xDrip and Nightscout-reporter are using this API to store and fetch multiple insulin profiles (names, curves, activation within GUIs and color codes).
Currently those insulin profiles aren't respected in any way by nightscouts GUI and backends.


AndyLow91:
This can be closed as superseded by the newer multi-insulin work in #8261.

This PR is quite old now, and the remaining discussion/implementation should happen in the newer PR so we have one current place to review the API shape, compatibility concerns, and test coverage. Thanks for starting this work.

Issue #7475, opened 2022-07-07
Build of current dev is failing on Heroku


yeagerm:
**Describe the bug**
From log of build attempt:
```
-----> Installing dependencies
       Installing node modules
       npm ERR! code EBADPLATFORM
       npm ERR! notsup Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin"} (current: {"os":"linux","arch":"x64"})
       npm ERR! notsup Valid OS:    darwin
       npm ERR! notsup Valid Arch:  undefined
       npm ERR! notsup Actual OS:   linux
       npm ERR! notsup Actual Arch: x64
       
       npm ERR! A complete log of this run can be found in:
       npm ERR!     /tmp/npmcache.mWHMw/_logs/2022-07-07T17_23_13_488Z-debug-0.log
-----> Build failed
```

psonnera:
Same here, [cgm-remote-monitor](https://github.com/nightscout/cgm-remote-monitor/tree/dev)/package-lock.json specifies darwin as the OS and that's not what's Heroku wants...

sulkaharo:
As discussed with Ben in private, the issue seems to be the lock file in the repo has been generated with a newer version of npm than the npm version in package.json. Heroku honors the defined node and npm versions, which then fail to interpret the file in the repo

bewest:
```
  "engines": {
    "node": "^16.x || ^14.x || ^10.22.0 || ^12.18.4",
    "npm": "^8.x || ^6.14.6"
  },
```
I've removed the `npm` requirements line and regenerated `package-lock.json` using node 10.x via `n 10`.  This lets the build succeed assuming the lowest common denominator is chosen.
https://github.com/nightscout/cgm-remote-monitor/pull/7487

sulkaharo:
We should be able to remove Node 10 and 12 from the officially supported versions, as these are not maintained anymore. OpenJS lists Node 14 as the Maintenance LTS version: https://nodejs.org/en/about/releases/

bewest:
Node 10, 12, and 14 produce identical package-lock.json files using npm 6.

bewest:
Should be resolved.

PR #7487, opened 2022-07-18
rm package-lock.json && npm install

bewest:
npm ls fsevent didn't list any output.  I'm running on linux,
and regenerating this file after removing it seems to remove
the fsevent package from the file.
It's likely that fsevent may be used in development mode as part
of one of the build utilities, but it shouldn't block anyone from
deploying.

https://github.com/nightscout/cgm-remote-monitor/issues/7475


gestrich:
Just FYI as I see this the branch name indicates this is work-in-progress... But I tried deploying this on a test heroku instance to work around issue reported in [7475](https://github.com/nightscout/cgm-remote-monitor/issues/7475) which I get in the latest dev. However, on this branch in the PR I continue to get the same deploy error regarding fsevents. Maybe that is not what this intends to fix.

gestrich:
@bewest I deployed the latest changes again as I see there were updates since my comments. It deployed successful on my Heroku test instance 👍  It would be great to get this into dev when we are able to restore Heroku builds of the dev branch.

bewest:
@sulkaharo I'd like to release pending `14.2.6` with everything in play and dev with node 12, followed by `14.2.7` release which only supports node `14.x` and `16.x`.  This allows anyone accidentally depending on node 10 or node 12 a clean transition point via `14.2.6` as a known quantity.

PR #7512, opened 2022-08-13
Allow Remote Carb Entries in Past or Future

gestrich:
This will send the Event Time from the Careportal in a push notification when sending a remote carb entry. This will allow users to input carbs in the past or in the future. The feature has been requested a few times in Loop's Zulip.

The screenshot below is just to show where the date fields exists in the Careportal. There is no new UI coming to the Careportal.

[Loop PR 1771](https://github.com/LoopKit/Loop/pull/1771) contains the Loop side changes to parse this field from the notification.

Screen Shot 2022-08-13 at 1 03 35 PM


lib/server/loop.js
```diff
@@ -81,6 +81,9 @@ function init (env, ctx) {
          if (data.otp !== undefined && data.otp.length > 0) {
             payload["otp"] = ""+data.otp
          }
+        if (data.created_at !== undefined) {
+          payload['created-at'] = data.created_at;
```
ps2, line 85: I see that NS is referring this to as "created-at", but a better name for the Loop side of things would be "start-time", since Loop uses "created at" as the time of entry, which can be different from the time of eating, which is start time.

gestrich: This has been updated to "start-time". Thanks.

gestrich: Just a note - There is an problem on dev that is causing dev to be not deployable on Heroku: https://github.com/nightscout/cgm-remote-monitor/issues/7475

I've worked around that by testing with the branch that Ben has which is linked in that case in another PR.

I'm mentioning as whomever wants to try this change, that uses Heroku, will need to cherry-pick it to a NS branch that is working like master or Ben's branch.

bewest: Thanks for contributing!  Cherry-pick shouldn't be needed. We'll merge this in.

PR #7513, opened 2022-08-14
Remove unused OTP field from Remote Override form.

gestrich:
We've had a few users on dev that were confused by the OTP field when entering a remote override. This field is not used by Loop for remote overrides. Maybe we should in the future but this PR removes the field until it is implemented on the Loop side and enforced.

**Before**

Screen Shot 2022-08-14 at 6 50 04 AM

**After**

Screen Shot 2022-08-14 at 6 56 54 AM


bewest:
How does this sit with the much anticipated release of Loop?

bill-foreflight:
@bewest This PR was just to avoid confusion for users that don't use remote bolus/carbs as the OTP field is not relevant. Merging this will lessen the support burden of people asking if they need to input something there. It doens't depend on a particular Loop release though. Let me know if that doens't answer your question.  cc @ps2

gestrich:
@bewest This PR was just to avoid confusion for users that don't use remote bolus/carbs as the OTP field is not relevant. Merging this will lessen the support burden of people asking if they need to input something there. It doens't depend on a particular Loop release though. Let me know if that doens't answer your question. cc @ps2

bewest:
Thanks @gestrich.  FWIW, I agree with your reasoning, and just want to be sure we understand the impacts of the changes.  Does AndroidAPS have Temporary Overrides that require OTP?

gestrich:
Hi @bewest. From my research, this remote override feature is a Loop only feature. Here's how I derived that:

* The "Temporary Override" event type is part of loop.js.
* I asked about this in [this Discord thread](https://discord.com/channels/629952586895851530/801760569342165004/1035609483969298502) and seemed to get confirmation they don't use this.
* I also downloaded NSClient just to verify if they have an "Override" feature like loop. I can't find anything so that supports this is not something used by AndroidAPS. [Their documentation](https://androidaps.readthedocs.io/en/dev/Children/Children.html) mentions remote "Profile switches" which appears to be a different feature

bewest:
@gestrich Thanks for all your work and thorough research! Your findings are extremely helpful in verifying our experiences.

Issue #7546, opened 2022-09-13
v2/properties world readable with AUTH_DEFAULT_ROLES=denied


cryptoluks:
**Describe the bug**
`/api/v2/properties` endpoint not authenticated.

**To Reproduce**
Grab any nightscout server which has env `AUTH_DEFAULT_ROLES=denied` active and test endpoint:

```
# curl -I https://example.herokuapp.com/api/v2/properties

HTTP/1.1 200 OK
[..]

# curl https://example.herokuapp.com/api/v2/properties

{"bgnow":{"mean"[..]
```

**Expected behavior**
All API endpoints should follow the `AUTH_DEFAULT_RULES` setting.

**Your setup information**
- Reproduced on docker `latest 195f1356c335`, `latest_dev 184a46dbce10` and several public instances.
- `ENABLE: careportal rawbg iob bridge`

**Additional context**
Found out about the v2 API after iOS nightguard got glucose values without providing any token.

Please let me know if this maybe desired behavior for some legacy devices, as nightguard called it _Pebble API_. As sugarmate and the webinterface seemingly can work without `v2/properties` this maybe should be optional or at least mentioned in the hardening part of the docs.

sulkaharo:
Oof ping @bewest

bewest:
Thanks for the report.  I think there are several parts of the API that remain active when `AUTH_DEFAULT_ROLES=denied`

* `/api/v2/properties`
* `/api/v2/ddata/at/:at`
* `/api-docs/*`
* `/api3-docs`
* `/swagger.{json,yaml}`
* `/swagger-ui-dist/*`

Any others?

bewest:
Of the affected endpoints, these two respond with information from the database:

* /api/v2/properties
* /api/v2/ddata/at/:at
These both return a mix of entities and treatment data.
This suggests permissions for: `api:entries:read`, as well as `api:treatments:read`.

These would be the files affected.
* https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/api/properties.js#L21
* https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/data/endpoints.js#L66
Something like this:

```
+  properties.use(ctx.authorization.isPermitted('api:entries:read'),
+    ctx.authorization.isPermitted('api:treatments:read'));
+  api.get('/at/:at?', ensure_at, get_ddata, format_result);

```

[...]

PR #7554, opened 2022-09-21
minimum attempt to protect /api/v2/ endpoints

bewest:
These endpoints should be protected as reported in https://github.com/nightscout/cgm-remote-monitor/issues/7546.


bewest:
* [ ] Needs additional tests.

sulkaharo:
Adding a better implementation to todo list and merging this for now

PR #7560, opened 2022-09-27
Add sample fly.toml for the hoster fly.io

ninelore:
Related to 
- https://github.com/nightscout/nightscout.github.io/issues/153
- https://github.com/nightscout/nightscout.github.io/issues/147

### DIFF
--- a/lib/language.js
+++ b/lib/language.js
@@ -39,6 +39,7 @@ function init (fs) {
     , { code: 'sl', file: 'sl_SL', language: 'Slovenščina', speechCode: 'sl-SL' }
     , { code: 'sv', file: 'sv_SE', language: 'Svenska', speechCode: 'sv-SE' }
     , { code: 'tr', file: 'tr_TR', language: 'Türkçe', speechCode: 'tr-TR' }
+    , { code: 'uk', file: 'uk_UA', language: 'українська', speechCode: 'uk-UA' }
     , { code: 'zh_cn', file: 'zh_CN', language: '中文（简体）', speechCode: 'cmn-Hans-CN' }
 //    , { code: 'zh_tw', file: 'zh_TW', language: '中文（繁體）', speechCode: 'cmn-Hant-TW' }
   ];


bewest:
Can this file be hosted in the docs repo? Would it be better placed in a new repo called `deploy-nightscout` with similar templates, or `nightscout-on-fly.io` as a standalone template?  It would be desirable to separate updating and maintaining this file from the release cadence of Nightscout.  The same logic holds for other deployment for other platforms.

ninelore:
> Can this file be hosted in the docs repo? Would it be better placed in a new repo called `deploy-nightscout` with similar templates, or `nightscout-on-fly.io` as a standalone template? It would be desirable to separate updating and maintaining this file from the release cadence of Nightscout. The same logic holds for other deployment for other platforms.

If we wpould host this file somewhere else it would unnecessarily complicate the required command to deploy on fly. Ive opened this PR because the Heroku files are in the repo aswell for the same reason.

psonnera:
A fly.toml is present in https://nightscout.github.io/vendors/fly.io/fly.toml but at this point we have many deployment files in the main repo that could be moved somewhere else. Just makes deployment more clumsy IMHO as people fork this repo and try to deploy directly from it. Maybe I misunderstand the strategy of deploy-nightscout and separating DIY solutions deployment files from (?) what exactly?

bewest:
No, the goal is to decouple the release cycle for cgm-remote-monitor from its deployment options.  If a deployment button changes, it's less likely to get fixed in a timely manner if it is also tied to needing a full release for `cgm-remote-monitor`.  We follow [12 factor deploy](https://12factor.net/).  Since the code shouldn't need to change when third party servers change, it makes sense to me to keep deploy templates somewhere separate from the code as well.  This will open up more deployment scenarios for more people.

bewest:
There's another argument in terms of sponsorship, does the button in the official repository signal sponsorship or support of that platform?  Does it signal support from service providers?  It would be an opportune time to de-brand Nightscout, to grow some distance from a single hosting company.  Here's a question: if Nightscout offers links to companies like fly.io, or Salesforce, should it also offer similar links and content to dedicated hosting providers, for fairness?  To the extent that there are commercial vendors contributing to Nightscout, what should the link for those companies look like, then?  Should similar companies be credited for sponsoring releases?  Having deployment resources cleanly separate from the code eliminates all of this and allows the code to be apolitical and brand agnostic.

psonnera:
@bewest making it more complex for people to deploy cgm-remote-monitor removing vendors deployment files and rejecting new vendors doesn't add any technical value to the project. If you want to remove the Heroku deploy button please make it in dev so that we can update the documentation accordingly before committing to master. If sponsoring companies contribute to the development of Nightscout they will show in contributors and the community will thank them for the effort.

sulkaharo:
So.. I think separating the deployment repo from the development repo could be a good thing. As Ben mentions, this would make it a lot more flexible to add deployment options without needing to make full releases of Nightscout itself. This would also make the main repo easier to maintain (less pull requests to the codebase and less files). I've considered a setup where we'd publish the main repo to NPM and set it up as a dependency of the deployment repo, possibly even so that the latest Nightscout would in some cases be pulled in by just redeploying the deployment repo, without the need to merge updates in Github, which would make it easier for people to keep their sites up to date.

I'd actually also consider the deployment repo should be something other than the documentation repo, as putting the deployment stuff into the docs would then introduce a dependency on docs publishing cadence and the permissioning of who's allowed to edit and publish that repository (where docs and deployment are separate concerns).

bewest:
To lend some data driven perspective, the last release cycles for Nightscout code was over 9 months.  While I'm trying to capture the momentum we've built recently, what if servicing these button requests takes 9 months to reach the main audience through release processes?  There may be a dozen different PaaS providers that are compatible with Nightscout.
There are many PaaS providers that will service variants of Nightscout including sidecars like tandemtconnsync, medtronic and libre proxy services.  Since each PaaS generally only allows one variant (a single file in a defined location), these variants will have to live in separate repositories.  I suggest we create a new github organization, something like `cgm-in-the-cloud` with repositories named like `nightscout-on-digitalocean`, `nightscout-do-tconnectsync`, `nightscout-on-fly.io`, `nightscout-on-[railway,northflank,...]`.  That way the url would be `cgm-in-the-cloud/nightscout-on-digitalocean` and similar.

Here's an example of using a PaaS button to set up tconnectsync as a sidecar.  https://elements.heroku.com/buttons/jwoglom/tconnectsync-heroku  I was inspired by this approach of having a repo dedicated to the deployment strategy and separate from the code base.

It's not clear how many repos there should be.  If tconnect, mongo local vs atlas, libre, and medtronic sidecars represent more than 5 flavors or variants of Nightscout deployments, then there may be around 50 or more buttons across all the compatible PaaS providers.  Presumably several simple setups for several common public infrastructure providers could be serviced from a single repository.  Something in the current `nightscout` organization might be fine, perhaps something like `cloud-deployments`, so the url for that might be `nightscout/cloud-deployments`, and this could hold the most popular templates for each PaaS?  Consider support for local vs mongodb atlas on Railway buttons.  If each has a repo, they both get buttons. On a tangent, here is a stub for using digital ocean as a PaaS. https://gist.github.com/bewest/2a4b58dfd7c8fa6d002eeecf1b5780c9.  I'm in a similar scenario for where to put files like this, and not particularly eager to include it in this repo for all the same reasons, and wondering also where to put equivalent file for including tconnectsync.  So we currently have at least 4 files in this category that need a new home outside of the Nightscout release process.
Thanks for considering and suggestions are welcome.

PR #7593, opened 2022-10-06
Forwarded fix

dnicolaas:
While trying to install Nightscout behind a reverse proxy, things weren't working right. I found out that authorization wasn't provided with an IP address in lib/server/websocket.js, because it was looking at the request only, while in other places, forwarded-for was used. Changing this fixed my issue.
I replaced all loose occurrences of 'forwarded' in that file and in /lib/api3/storageSocket.js with a function 'getRemoteIP', which is consistent with all other places where this is needed.
I'm new to the project, so I was not able to work out a feasible test for this...


dnicolaas:
Is there anything I need to do to have this pull request being looked at?

PR #7647, opened 2022-10-30
Fix empty import config

k2s:
If `env.IMPORT_CONFIG` is not provided and is null we try to `url.parse` and print error `console.error('Parsing config URL from IMPORT_CONFIG failed');`.


bewest:
Looks good, thank you!

PR #7699, opened 2022-11-21
Update history length to 12 housr

chiper99:


chiper99:
Sorry, created by accident

PR #7714, opened 2022-11-27
style svg with css

Joerg-Schoemer:
- use css instead of element styling
- try to print one chart with insulin distribution on a landscape page

### DIFF
--- a/lib/report_plugins/daytoday.js
+++ b/lib/report_plugins/daytoday.js
@@ -302,7 +302,7 @@ daytoday.report = function report_daytoday (datastorage, sorteddaystoshow, optio
         })
         .on('mouseover', function(d) {
           if (options.openAps && d.openaps) {
-            client.tooltip.style('opacity', .9);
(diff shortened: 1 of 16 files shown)


lib/report_plugins/daytoday.js
```diff
@@ -734,9 +727,7 @@ daytoday.report = function report_daytoday (datastorage, sorteddaystoshow, optio
             .attr('x1', xScale2(treatment.mills) + padding.left)
             .attr('y1', lastfoodtext * 15 + padding.top)
             .attr('x2', xScale2(treatment.mills) + padding.left)
-            .attr('y2', padding.top + treatment.carbs ? yCarbsScale(treatment.carbs) : (treatment.insulin ? yInsulinScale(treatment.insulin) : chartHeight))
```
Joerg-Schoemer, line 737: oops, I've missed the openaps tooltip. Will add a class 😄

PR #7784, opened 2022-12-20
Fix Server error

Joerg-Schoemer:
* check if d1 is not nil before call getTime


sulkaharo:
Under which conditions are you seeing this trigger a null pointer exception?

Joerg-Schoemer:
I'm not sure which of the following requests trigger the null-pointer:
/api/v1/treatments.json?token=<token>&find[eventType]=Sensor%20Change&count=1
/api/v1/treatments.json?token=<token>&find[eventType]=Site%20Change&count=1
maybe both!

lib/api/treatments/index.js
```diff
@@ -63,18 +63,20 @@ function configure (app, wares, ctx, env) {
       }
     });
 
-    if (!_isNil(d1)) res.setHeader('Last-Modified', d1.toUTCString());
```
Joerg-Schoemer, line 66: it is possible that d1 is null, but the next call d1.getTime() was not 'not null safe'.

Joerg-Schoemer:
I‘ve  found the🪳
When no treatment is found by the query, d1 is null.

PR #7791, opened 2022-12-29
Remote Commands

gestrich:


sulkaharo:
So... Please write up a description of what this change is intended to do? FWIW if I'm right at guessing what this is intended for, the architectural way we're discussed Nightscout would handle this is providing an API that allows users to store "treatment requests" that cannot be deleted so the system retains a track record of data. Also note for a new feature to be added to Nightscout, it should be implemented as a full class feature, including the ability add the data using the web view and visualising the data in the timeline.

gestrich:
@sulkaharo I created this as a Draft just to begin sharing some early information with the community. This feature has been in discussions on the Loop side in Zulip (see the Caregiver stream) but I realize we need to get alignment with the Nightscout community on how this happens. I don't know how this fits in with other Nightscout plans but welcome feedback on how this can advance.

sulkaharo:
So - ideally any features added into Nightscout are implemented in manner that makes them cross-compatible across multiple systems with similar features. So in case of being able to remotely request treatments, it'd be great if folks such as the AndroidAPS and Loop teams worked together to figure out a spec on how this could work, including doing the risk analysis over how a feature like this can be implemented without compromising safety. One of the issues I'm discussing with a variety of folks right now is how to proceed with Nightscout development, where a key problem I'd like to solve is that the codebase includes features that have been implemented for a specific external application (including Loop and AndroidAPS) while at the same time these features don't have good automated test coverage and nobody is maintaining the code. What this has lead to is, as of right now, the dev branch of Nightscout has multiple dependencies with critical security issues and those cannot be upgraded without a large refactor of the entire codebase, which is practically impossible without cleaning unmaintained features away as I can't test a lot of the aforementioned code and the Nightscout maintenance has for years pretty much been a single man job, with help from Ben and some other contributors. So I guess my question here is - if you take a long view of how this feature can exist in a sustainable way, how'd you implement it?

gestrich:
@sulkaharo Thanks for the background. I suppose a good place to start is for me to chat with the AAPS developers to understand how this could fit into their feature set. Who or where would be the right place to initiate that discussion?

sulkaharo:
To add more context - the current Loop push notification implementation uses the `apn` package, which hasn't been updated in 5 years and now depends on npm modules that have critical security issues. If you have the time to update the current code to using components that are still maintained, that'd be awesome.

bill-foreflight:
I can look into upgrading the library as part of this work. [Node-apn from parse-community](https://github.com/parse-community/node-apn) seems like the successor to use.

In terms of the coordination you suggested with AAPS developers, I'll plan to post a proposal in the #nightscout-developers channel in [Discord](https://discord.com/channels/629952586895851530/756770295368384542). But let me know if there is a more preferred method to get this proposal to the AAPS group for discussion.

gestrich:
I can look into upgrading the library as part of this work. [Node-apn from parse-community](https://github.com/parse-community/node-apn) seems like the successor to use.

In terms of the coordination you suggested with AAPS developers, I'll plan to post a proposal in the #nightscout-developers channel in [Discord](https://discord.com/channels/629952586895851530/756770295368384542). But let me know if there is a more preferred method to get this proposal to the AAPS group for discussion.

gestrich:
> I can look into upgrading the library as part of this work. [Node-apn from parse-community](https://github.com/parse-community/node-apn) seems like the successor to use.
> 
> In terms of the coordination you suggested with AAPS developers, I'll plan to post a proposal in the #nightscout-developers channel in [Discord](https://discord.com/channels/629952586895851530/756770295368384542). But let me know if there is a more preferred method to get this proposal to the AAPS group for discussion.

@sulkaharo  Here's a draft PR for the APN update. If you are able to give any feedback as I do some of my own additional testing, let me know. https://github.com/nightscout/cgm-remote-monitor/pull/8047

PR #7807, opened 2023-01-02
apiv3 fix patch operation

MilosKozak:


sulkaharo:
If we ship parsing the dates, guess there should be checks to see those fields aren't changed? @PetrOndrusek how do you think the semantics of this operation should work?

MilosKozak:
this is done in validate.js

PR #7810, opened 2023-01-04
dia2ns

LouisRobert2017:
i wasn't able to get the diasend 2 Nightscout thing done

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,8 @@ jobs:
         run: npm install
       - name: Run Tests
         run: npm run-script test-ci
+        env:
+          NODE_OPTIONS: "--max_old_space_size=4096"
       - name: Send Coverage
         run: npm run-script coverage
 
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -37,6 +37,8 @@ jobs:
         run: npm install
       - name: Run Tests
         run: npm run-script test-ci
+        env:
+          NODE_OPTIONS: "--max_old_space_size=4096"
```
sulkaharo, line 41: Does this really require this much memory? If yes, this PR is unmergeable as most people host Nightscout in environments with less than 512 MB of RAM

sulkaharo:
Hi! Was this PR intentional? Looks like it was made accidentally?

.github/workflows/main.yml
burnedikt: https://github.com/nightscout/cgm-remote-monitor/pull/7820#issuecomment-1375368711

sulkaharo:
Closing, see #7820

Issue #4901, opened 2019-08-17
Fixed offset timezones


ps2:
**Is your feature request related to a problem? Please describe.**
When Loop uploads fixed offset timezones, like "GMT-0500" they don't show up in the timezone picker for the profile editor. The schedule rates themselves seem to be displaying properly in the main screen, so it seems like there is some level of support for them.  Fixed offset timezones are important for Loop, as they are more representative of how time is tracked in a pump than timezones that have special rules for daylight savings time or other switches.

**Describe the solution you'd like**
The offset that the schedule is stored in show be displayed in the picker.

sulkaharo:
I'm not sure I follow what the desired end goal is here? Where do you see the profile time zone being used? Is Loop reading it in? Is some data being displayed wrong somewhere?

Nightscout's current date implementation works so that when data is sent to the REST API, the API expects the dates to use the ISO-8601 format. The date is then normalized to UTC in the database for the purpose of REST queries to work as expected (given the dates are treated as Strings in the database layer and if the system allowed mixed format non-normalized dates, the API would give incorrect query results). The API then returns the normalized strings, unless the instance is configured to de-normalize strings back to format that has the time zone. Internally in the runtime the dates are deserialized to Date objects and logic either uses the API in Date() or "UTC milliseconds from January 1, 1970 00:00:00 UTC" as per Date.now() representation for comparisons. Nightscout clients are expected to basically use whatever native date implementation is in the platform they've been implemented in, and deserialize the ISO-8601 dates to the native date objects and not make assumptions on the string representation other than the date being in the ISO format.

ps2:
The profile timezone should be shown in the profile editor, and it isn't for fixed offset timezones.

ps2:
I also realized that the console log was filling with error related to the fixed offset timezones: 
<img width="413" alt="211_-8_→" src="https://user-images.githubusercontent.com/14649/63401551-4cf64300-c39d-11e9-96a4-5011bbeb76d9.png">
Based on those errors, I patched Loop to upload fixed offset timezones with identifiers like `ETC/GMT+6` instead of "GMT-0600" (https://github.com/ps2/rileylink_ios/pull/535), which does fix the console errors. But, unfortunately, the dropdown in the profile editor is still blank, so I'll keep this ticket open.

Nightfoxy:
[...]

PR #7836, opened 2023-01-15
daytoday reports are always in the profile's time zone

bewest:
Midnight for the profile should always be at the start of the day (midnight) for the chart.  There's logic for this in `lib/report/reportclient.js`, but believe it needs to be here as well.


marionbarker:
Been a little busy with Loop 3 docs. I appreciate this and I will try it soon.

bewest:
Thanks for the updates!
https://github.com/nightscout/cgm-remote-monitor/issues/4901#issuecomment-1386536027

I expected there to be a kind of "off by one" error, so the right column second row actually makes sense to me.  This is progress.

As for the date being wrong, can you be a bit more explicit.  There seems to be a date rollover effect with respect to midnight.  Did you pull up these reports at a time when the -2, +3 would have resulting in rolling over midnight?  Is the date correct in the other timezone for one of those cases?  We're close.  If possible, I'd like to peek at your NS instance to poke at it, we can discuss that in a private channel.

bewest:
Tests do not pass, but I suspect it's something with the tests themselves.  This is worth trying, as it should handle both remaining issues:
* title of the day is incorrect (always the previous day?)
* midnight wrapping around to next day when crossing the datetime line.

sulkaharo:
Yea if you've changed the date logic, it's likely the report outputs different values. Check reports.test.js starting line 273

marionbarker:
I tried the changes through 8764b84 alone and with 7833 patched into master. Did not fix the problem.
I'm installing straight dev now. When I have time, I'll try the patches for dev.

bewest:
Strange tests... on my machine, many tests fail (careportal, hashauth) because loading the html into `self.$('body').html(indexHTML)` in `tests/fixtures/headless.js` takes more than 60 seconds allowed by the test.  If I adjust it to 65 seconds some tests work, but I still get an error in "hashauth should store hash and the remove authentication:"`[

```
not ok 298 hashauth should store hash and the remove authentication
  Cannot read property 'should' of null
  TypeError: Cannot read property 'should' of null
      at Context.<anonymous> (tests/hashauth.test.js:126:49)
      at benv.setup.url (tests/fixtures/headless.js:174:7)
      at Object.module.exports.setup (node_modules/benv/index.js:24:56)
      at Function.init [as setup] (tests/fixtures/headless.js:24:10)
      at Context.<anonymous> (tests/hashauth.test.js:28:14)
      at processImmediate (internal/timers.js:464:21)

```
This is puzzling as these changes don't interact with any of that code.

Notably, this is different from the failures seen in github actions, which fails the report test, which does not occur on my machine.  In theory just widening the date range for the query should in theory should allow the test to pass, but given the other oddities here, it's a little odd.

bewest:
closing in favor of https://github.com/nightscout/cgm-remote-monitor/pull/7857

PR #7839, opened 2023-01-18
Delete v3 api identifiers when cloning a record

sulkaharo:


lib/profile/profileeditor.js
```diff
@@ -389,6 +389,9 @@ var init = function init () {
       newname += '1';
     }
     record.store[newname] = _.cloneDeep(record.store[currentprofile]);
+    var r = record.store[newname];
+    if (r.srvModified) delete r.srvModified;
```
bewest, line 393: Maybe consider `record.store[newname] = _.omit(r, 'srvModified')` ?  Optional.

sulkaharo:
Ok _now_ it might work

sulkaharo:
Hmm, tests run fine locally with the same Node version, need to investigate later

MilosKozak:
there is one line deleted in other code

sulkaharo:
The line 319 deletion is due to adding the deletion of the `_id` field to the same line that deletes the apiv3 identifiers

MilosKozak:
line 391 must be there. it does copy current profile to new profile (not related to cloning records)

sulkaharo:
Yes, the _id field needs to be deleted. That logic is now on line 315, which now deletes both the _id and the fields related to APIv3.

MilosKozak:
you are mixing 2 different operations 
* cloning record - creating new record in mongo
* cloning profile - creating another profile inside the same record

sulkaharo:
Are we both looking at the this same code? 

Screenshot 2023-01-19 at 21 31 52

Here line 315 pushes a new record to the top of the `mongorecords` array.
316 updates the `currentrecord` pointer to point to the object that was just pushed.
317 updates the startDate field of the record to be the moment the record was duplicated.
318 sets current profile to be the default of the new record.
319 used to delete the _id field of the new record (not the profile), but this line is obsolete since line 315 that creates the record copy removes the '_id', 'srvModified', 'srvCreated' and 'identifier' fields from the object.

The '_id' parameter given to omit is the same as not omitting the field and then deleting it after cloning. Debugger says the field is gone. I don't understand what the bug is?

MilosKozak:
this is ok. but following is not 
image

sulkaharo:
🙈

MilosKozak:
👍 :)

MilosKozak:
damn.
must be this file in git when it's get modified runtime?

MilosKozak:
i'd remove it and add to .gitignore

MilosKozak:
@sulkaharo can we merge?

PR #7858, opened 2023-01-24
Alarm sockets for api v3

MilosKozak:


lib/api3/alarmSocket.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
+
+const apiConst = require('./const');
+const forwarded = require('forwarded-for');
+
+function getRemoteIP (req) {
```
sulkaharo, line 6: We have this copied to multiple places. :( I can clean this after the merge

sulkaharo:
Tested, merging

PR #7884, opened 2023-02-08
Add logging limits to docker-compose.yml for all services (mongo, nightscout and traefik)

bjornoleh:
Common log limit parameters for json-file driver defined in x-logging extension field with options:
    max-size: '10m'
    max-file: '5'

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,10 +1,18 @@
 version: '3'
 
+x-logging:
+  &default-logging
+  options:
+    max-size: '10m'
+    max-file: '5'
+  driver: json-file
+
 services:
   mongo:
     image: mongo:4.4
     volumes:
       - ${NS_MONGO_DATA_DIR:-./mongo-data}:/data/db:cached
+    logging: *default-logging
 
   nightscout:
     image: nightscout/cgm-remote-monitor:latest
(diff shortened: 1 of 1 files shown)


bjornoleh:
The template for [docker-compose.yml](https://github.com/nightscout/cgm-remote-monitor/blob/master/docker-compose.yml) does not include any limits for log retention. Depending on uploaders etc in use, logs will eventually exhaust available disc space on the host running Docker, potentially rendering it unable to boot. This has happened for me and a few others who have been using similar setups to mine.

I have been running Nightscout, mongo and traefik with docker compose based on the template from cgm-remore-monitor for a couple of months now. I have run it on a free tier Google Cloud Compute Engine (E2-micro, 1 GB RAM, 30 GB disc, Ubuntu 20.04). 

Uploaders include FreeAPS X, which produce more data traffic and generate more logs than most comparable apps. When I previously had Nightscout running on Heroku, the 10 Mb/day Papertrail free logging was filled up after less than half a day. Hence my use case may reveal this issue much sooner than others.

My current setup with mongo, traefik and four instances of NS had a daily increase in disk usage of roughly 700 Mb. It seemed like most of this was due to logging.

After running for a day with the suggested limits to log retention, disk usage seems to have levelled off:
image

I am open to suggestions for the logging parameters:
```
options:
  max-size: '10m'
  max-file: '5'
driver: json-file
```

Tagging @pazaan who originally submitted docker-compose.yml

pazaan (commented):
LGTM 🚀 Thanks for the contribution!

PR #7887, opened 2023-02-10
add plugins, weather, graph and tab to clock views

tynbendad:
Disclaimers/notes:
- This is the client side only, i have not done anything on the config side but it can be used by manually creating the URL.
- Weather requires an Open Weathermap API key (you can get them for free at their website), and it is passed in the URL.
- Plugins are all or none (not individually specified), and only include the ones I regularly use - some are likely missing (e.g., "loop".) If a warning or alert level occurs the entire plugin text will highlight (independently from the SGV threshold highlight), with the specific alerting plugins uppercased.
- The graph only loads extra SGVs initially and doesn't backfill if it misses some (e.g., network down > 4 readings).


tynbendad:
image

tynbendad:
url used for the above: https://<site>/clock/by15-ow16:F:<zip>:<owm key>-nl-nl-tm20-nl-nl-sg16-ar12-dt12-nl-ag8-nl-nl-pl5-nl-nl-gr350:150:8:2

tynbendad:
a plugin alert:

Screenshot 2023-02-10 092712

tynbendad:
also there is a 1 line change that fixes a bug in the existing clock view that will not signal the data is out of date if a web request never passes.

tynbendad:
some usage comments are in the change:
    // open weathermap params: -ow{font size}:{F|C}:{zip}:{api key}
    // example: -ow16:F:01108:D17C57239B5661055D459A608F770208 (note: this example is not a valid OWM key)

    // SGV graph size params: -gr{width}:{height}:{radius pixels}:{hours}
    // example: -gr100:100:10:1

sulkaharo:
I think this needs a refactor that splits the clock client into multiple files (which might be compiled to a single file) that separates the concerns so the clock code would still be easily maintainable even if this is merged.

tynbendad:
@sulkaharo I'm not too familiar with JavaScript (it was a stretch for me to get it this far), but I would look into splitting to multiple files if you could point me to more info on how to make that work...

tynbendad:
i'm not sure why a div with different colors ends up on separate lines but i kind of like this style plugin highlighting instead - will update the pr once i've tested enough.

image

tynbendad:
figured that out, going with this for now:
image

tynbendad:
i decided to attempt config today, should have another pr update later with it...

tynbendad:
image

tynbendad:
image

tynbendad:
> I think this needs a refactor that splits the clock client into multiple files (which might be compiled to a single file) that separates the concerns so the clock code would still be easily maintainable even if this is merged.

@sulkaharo I think I might have figured out refactoring, just want to check - is your goal to get each .js file to be under 500 lines? or, what is the goal?

tynbendad:
i updated with the refactored code, hope i got close to what you wanted.
also improved the stale data handling so the age/threshold is handled by the update-clock, since i was seeing my old ipad die with the clock working but nothing else updating (which meant stale data was being shown but not flagged as old.)
and improved the graph bg axis to be partially dynamically ranged, and annotated it.

tynbendad:
there are now 3 bug fixes to the original clock view included in this PR as well as the new functionality:
- as long as the time can be updated the stale bg alert (background/foreground color reversing) will fire (previously if the query timer thread failed, which can and did happen, the page would keep displaying old BG/delta/age info.)
- the delta display will disappear when there is no delta info., as it does in the main NS page (previously it would show the previous delta value, even though the BG may have been updated.)
- out of range 40-400 BG values will show up as HIGH/LOW/?? as per the main NS page (previously they would just show as the out of range value.)

tynbendad:
ancient tablet still usable

robster7674:
This is an awesome enhancement, it would be great if this could be merged soon!

@tynbendad : how would I go about and use this based off your branch whilst awaiting merging?

tynbendad:
> This is an awesome enhancement, it would be great if this could be merged soon!
> 
> @tynbendad : how would I go about and use this based off your branch whilst awaiting merging?

Thanks I use it every day, it's my main go to page now as well as on 2 older tablets in the house... I think you should be able to create your own branch based on this one (tynbendad:10feb23) but I don't know how really.

tynbendad:
fwiw, i've been using a v15 dev branch tree with these clock changes since last august: https://github.com/tynbendad/cgm-remote-monitor/tree/dev-6aug23-clock
also recall, it has 3 clock page bug fixes (i don't know if they've been fixed in dev or master since though), per: https://github.com/nightscout/cgm-remote-monitor/pull/7887#issuecomment-1454873889

PR #7888, opened 2023-02-12
Add event type to event on treatment notify if it's missing

sulkaharo:
Fixes treatment notification not creating a notify if user doesn't select an event in Care Portal or a client uploading data with no treatment event type


MilosKozak:
👍 working

PR #7894, opened 2023-02-16
Fix loading JWT when auth subject has no permissions

sulkaharo:
Fix an issue reported by Milos, when a JWT is loaded with a subject that has no permissions


MilosKozak:
I didn't test but it seems to be ok

PR #7911, opened 2023-02-25
Fixes #7910

### LINKED ISSUE #7910: Issue with api v1 returning stale data
I have a Mac OS X app that I made that pulls the latest CGM values from Nightscout using v1/entries - a user recently reported that their values are all showing as stale / old. When I looked into it, I can see that he's on the DEV branch of cgm-remote-monitor, and when I hit his v1/entries endpoint and compare to v2/properties it's clear that the v1 endpoint is no longer reporting up-to-date values. 

I have since upgraded my own instance and I can see the same behaviour.

**To Reproduce**
1. Go to api/v1/entries on a NS instance running 15.0
2. Look at the timestamps / epoc times and you will see they are multiple hours old versus the latest bgnow readings at api/v2/properties

**Expected behavior**
api/v1/entries should be up to date.

- NS version 15.0
- My data comes from loop://iPhone but the other user's data comes from xDrip-DexcomG5 G6 Native
- Not specific to a browser or device

sulkaharo:
As reported on #7910, the entries API call without the type specifier was returning data from the runtime state in wrong order, resulting in 48 hour old entries being returned first. Expected behaviour was to return latest entries first.

The PR additionally fixes insertion of CGM entries to /entries using the moment of insertion for entries that sent a numeric date moment for the record, rather than a dateString.

### DIFF
--- a/lib/server/entries.js
+++ b/lib/server/entries.js
@@ -99,8 +99,11 @@ function storage (env, ctx) {
 
       // Normalize dates to be in UTC, store offset in utcOffset
 
-      var _sysTime = moment(doc.dateString).isValid() ? moment.parseZone(doc.dateString) : moment(doc.date);
-      _sysTime = _sysTime.isValid() ? _sysTime : moment();
+      var _sysTime;
+
+      if (doc.dateString) { _sysTime = moment.parseZone(doc.dateString); }
+      if (!_sysTime && doc.date) { _sysTime = moment(doc.date); }
+      if (!_sysTime) _sysTime = moment();
 
       doc.utcOffset = _sysTime.utcOffset();
       doc.sysTime = _sysTime.toISOString();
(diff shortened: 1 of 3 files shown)


lib/server/entries.js
```diff
@@ -99,8 +99,11 @@ function storage (env, ctx) {
 
       // Normalize dates to be in UTC, store offset in utcOffset
 
-      var _sysTime = moment(doc.dateString).isValid() ? moment.parseZone(doc.dateString) : moment(doc.date);
```
sulkaharo, line 102: The logic mistake here is - if dateString is missing or not valid, Moment defaults to the same behaviour as if just moment() was called, so the subsequent .isValid() calls are always truthy

bewest: I looked at git blame; I think this code has been refactored so much that it's safe to remove any of this logic that does not make sense.  Git blame helped me recall that originally, about https://github.com/nightscout/cgm-remote-monitor/blame/f9d59c4e3e34c381b0925a60a5029f83d4e0d8e4/lib/server/entries.js#L83-L100, it was a major effort to switch from mills, which had bugs and issues in different timezones, to using any variant of 8601.  In addition, we were receiving two times system time and display time from the original G4 receivers, and hadn't decided which one to use yet.  In any case, over the years mobile uploaders have made syncing on universal time more trustworthy.

PR #7949, opened 2023-03-12
emoji clock

bjorkert:
After being inspired by Jon Faucet's SugarPixel, I decided to enhance the custom clock view by adding the ability to display an emoji that corresponds to the current blood sugar value. My 9-year-old child with type 1 diabetes helped me choose the emojis. We are using a custom clock view with svg, emoji and time on a raspberry with a small screen in kiosk mode. I have thoroughly tested the implementation in different environments, including dev, master, and units of measurement (mmol/L and mg/dL). Overall, I're thrilled with the outcome and hope it can benefit others as well.

Examples:
image
image

### DIFF
--- a/lib/client/clock-client.js
+++ b/lib/client/clock-client.js
@@ -118,6 +118,8 @@ client.render = function render () {
   // Insert the delta value text.
   $('.dt').html(deltaDisplayValue);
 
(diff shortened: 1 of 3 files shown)


bewest:
Nice!  People have special regions of the brain for recognizing faces and what they mean.  Good idea.

PR #8026, opened 2023-05-29
Less frequent Mongo updates

### LINKED ISSUE #8024: MongoDB Atlas reports too many connections to the database
**If you need support for Nightscout, PLEASE DO NOT FILE A TICKET HERE**
For support, please post a question to the "CGM in The Cloud" group in Facebook
(https://www.facebook.com/groups/cgminthecloud) or visit the WeAreNotWaiting Discord at https://discord.gg/zg7CvCQ

**Describe the bug**
We have several reports of AAPS/Loop users receiving a mail from MongoDB specifying they are using more than 500 connections and M0 doesn't support this. As a result, Nightscout will crash.

**To Reproduce**
Not known yet, users confirm they didn't change the amount of uploaders/downloaders

**Expected behavior**
500 connections seem impossible for Nightscout

**Screenshots**

image
image

**Your setup information**
- Nightscout 14.2.x
- AAPS/Loop

**Additional context**
Not known yet

(Will edit and update with new information, not using Atlas I haven't seen the issue)

sulkaharo:
Increase debounce for data loads to 15 seconds and trigger a data-loaded event to process local data updates even if the call is debounced

Maybe fixes #8024

### DIFF
--- a/lib/server/bootevent.js
+++ b/lib/server/bootevent.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const _ = require('lodash');
-const UPDATE_THROTTLE = 5000;
+const UPDATE_THROTTLE = 15000;
 
 function boot (env, language) {
 
@@ -292,6 +292,7 @@ function boot (env, language) {
 
     ctx.bus.on('data-received', function forceReloadData ( ) {
       console.info('got data-received event, requesting reload');
+      ctx.bus.emit('data-loaded'); // Since we update local sandbox instantly, process data-loaded right away in case this gets debounced
       updateData();
     });
 


bewest:
Nice upgrade to mongodb drivers!

bewest:
Testing this, and it does something weird in some of my test environments. On some environments it can't bootstrap the connection correctly, but it's not clear if it may be a server misconfiguration issue.

RickyGrassmuck:
Looking forward to seeing this get merged!

I explored using FerretDB (An alternative to Mongodb that is backed by postgres), which requires using updated mongo driver compatible with Mongodb 6+ to work, as a replacement for Mongodb and early testing looked promising but I didn't have the bandwidth for diving in to fix all the little incompatiblies needed so I had to shelve the idea.

If/when this gets merged, definitely going to be revisiting that project and see if it's truly a viable option to replace Mongodb!

RickyGrassmuck:
> Testing this, and it does something weird in some of my test environments. On some environments it can't bootstrap the connection correctly, but it's not clear if it may be a server misconfiguration issue.

Happy to spin up some test instances and see if I can replicate the connection issues.

bewest:
Howdy @RickyGrassmuck, this is on track to be merged.  I've been ultra-conservative testing scenarios on databases that run in replica set and that have been migrated or have mixed upgrade/migration status or even broken cluster configurations.

bewest:
Finally catching up on my backlog.  Additional testing, and I like this.  If the startup boot sequence will more predictably take 15 seconds instead of 5 seconds it might make a nice first-time contribution for someone.
Additionally, this looks ok in node 14, 16, and 18, so thinking about adding node 18 to the compat list.

bewest:
Looking at bringing this back in...

PR #8029, opened 2023-06-04
Summary V2 API and API V2 cleanup

sulkaharo:
* Creates new /api/v2/summary V2 API to load minimized current system state in one request for use in watchfaces and other display-only applications, optimised for small data load size and plotting
** Basal curve, preprocessed for plotting from temporary basals and profile
** Insulin and carb treatments
** Profile
** CGM data
** Use hours=n GET parameter to load data from most recent n hours, up to 48
* Cleans V2 API implementation
* Removes unnecessary server init calls in API calls that were unnecessary
* Expose latest SBX state through the CTX object


bewest:
Looks good; intent is refactor + new feature, no breaking changes, it looks like?

bewest:
FWIW, I merged this with my nightscout-connect branch for easy of testing. https://github.com/nightscout/cgm-remote-monitor/pull/7983

PR #8064, opened 2023-07-30
Fix Care Portal Default Event Date for Manual Entry

jpcunningh:
The care portal's default event date that is set for modification if the user select the "Other" radio button is set to the UTC date instead of the user's locale date. This has caused us to enter an event one day into the future a few times inadvertently when putting in carbs after the fact and adjusting the time accordingly.


jpcunningh:
@sulkaharo, the default value for the date/time selection has the wrong date if my current time is not in the same day as UTC. If I select the correct date, it works fine. However, if I'm in that window of time during the day that my date is not the same date as UTC (7 in the evening through midnight for me), I have to manually change the date one day back to my local date.

For you, I think it would affect your Nightscout instance from your local time of midnight to 3 in the morning. The default date would be one day behind.

AndyLow91:
Thanks for raising this and for the clear explanation.

I think this has now been superseded by #8307, which addresses the same careportal local-date issue but uses moment.format('YYYY-MM-DD') rather than splitting an ISO string, and also adds regression coverage for the UTC/local-date boundary case.

I’m going to close this PR in favour of continuing with #8307 so we keep the remaining review and follow-up work in one place.

PR #8076, opened 2023-08-16
Add notes about the Google Assistant integration being broken

inventor96:


bewest:
Thanks for confirming, should we remove the code and log a deprecation warning as well?

bewest:
I'm interested in otherwise comparable alternatives.  To the extent we can remove any dependencies, we would like to do so, especially non-functional ones.  We've spent the better part of the last year working on bringing dependencies up to date and to prepare for removing things that aren't going to help us move forward.  It makes sense to me to remove all the google-assistant specific code and replace it with a warning on the console, especially if it allows us to remove a dependency, especially any additional http libraries.

inventor96:
Good to know. Thankfully there's no dependencies

inventor96:
Whoops, wrong button.

There's no dependencies that are unique to the Google Assistant integration, so we're clear there. I suppose we could remove the code, and maybe I'll stick a link in the docs to the last commit with the code. That way anyone who figures out a good way to do it again can use it as a reference.

inventor96:
In what context do you think a warning log message will be needed? Like if they have the plugin in the enabled list?

bewest:
Yes, if someone enables the plugin, what should happen?
I now face a similar issue with medtronic specific support for nighscout-connect.

inventor96:
As it stands right now, if the plugin is enabled, the app will have the routes setup for this plugin. Since nothing will be making calls to those API endpoints, nothing will be happening after that. There's no "active" part to this plugin; it's all "reactive" since nothing happens until a request is made to that specific endpoint.

Issue #6916, opened 2021-02-22
OpenAps features visible in pillar


maja-lofgren:
I just want to share a fiew tweeks that I've made to my NS that helps allot in visualizing important data.

**Describe the solution you'd like**
Modifications I've implemented:
* IOB: Added the activity
* COB: buggfix to show history COB when scrolling the timeline backwards
* BASAL: added normal basal after current temp basal in "()"
* OpenAPS: Added current Target (T), ISF, CR and sensibility ratio (sr) 

![image](https://user-images.githubusercontent.com/73176049/108684881-be173900-74f3-11eb-949d-ad04cdc3491a.png)

**Additional context**
My code is not properly tested, and might fail if not used by AAPS.... more lika a proof of consept!
[my branch diff against dev](https://github.com/nightscout/cgm-remote-monitor/compare/dev...maja-lofgren:wip/customTest)

bjornoleh:
Hi @maja-lofgren 

I tested two of your individual commits now:

- Fix to display historic COB-value (from OpenAps) when scolling back in time https://github.com/nightscout/cgm-remote-monitor/commit/7aa606ddac735fce821119e5bb674c08bc8441f0

- Added profile basal rate in () next to current temp basal https://github.com/nightscout/cgm-remote-monitor/commit/58495f788f68fd2810e2b59725d14b19f3425398

Both of these seems to work well here. Thanks for this!

We are not using AAPS, so can't test the "IOB: Added the activity" patch.

I also could not identify the relevant commit(s) for "OpenAPS: Added current Target (T), ISF, CR and sensibility ratio (sr)". I might test it later, but it will also take up a lot of space in NS (on a phone), so not sure if I 'd use it.

I am not involved in NS development, but would encourage you to prepare branches for your patches to allow PRs to NS dev. Please keep your PRs as small as possible, without bundling too much into each one. And start off with at clean branch that is even with dev.

AndyLow91:
The COB-specific part of this issue has now been addressed by #8492.

That PR fixes historical COB display when scrolling/rewinding the timeline, using the viewed timeline time rather than the current real-world time when deciding whether controller-reported COB is still relevant. It also covers both OpenAPS-style and Loop-style COB data.

The other items in this issue, such as IOB activity, basal display changes, and additional OpenAPS pill details, remain separate from that fix, and may still require attention.

Issue #5241, opened 2019-11-25
COB is always zero when rewinding/scrolling back


naromero77:
**Describe the bug**
I believe this worked in the recent master but not in dev. The **current** COB pill displays correct value and updates correctly as time progress. However, if you rewind/scroll back the NS viewer you to a prior time the COB pill always displays zero. 

**To Reproduce**
Set-up is NS dev from a couple of weeks ago. Recent version of Loop dev.

Take NS viewer in a Firefox or another browser and just scroll back: COB is displaying zero, but the current instantaneous COB is correct.

**Expected behavior**
If you rewind/scroll back the NS viewer, I expect the COB displayed to show the correct value at the prior time.

**Screenshots**

**Your setup information**
- NS dev
- G6 Sharebridge
- FireFox

**Additional context**
I can add a screenshot, but it would not be interesting. It should always shows "0" in the COB pill when you scroll back. In the present time, COB is displaying correctly.

woodduckie:
I've observed this issue after updating to v13.  I can't be certain it wasn't there before and I haven't yet had the time to roll back my deployment.  To add to the original issue, IOB reflects historical values when scrolling, but COB just shows 0.  FWIW, I'm using loop, Firefox /Safari, G6 via bridge.

Sophieinthecloud:
![NS_COB](https://user-images.githubusercontent.com/10387075/88825015-e74d3200-d17b-11ea-8abb-68b2d4e178aa.png)

I'm having this same issue. When scrolling back in time, The COB shows in the Loop pill when I click it but the COB pill shows 0(see screenshot #1). 
A new bolus does show the COB (see screenshot #2). 

I'm not exactly sure when it started but I first noticed it after I cleaned mLab. Not sure if that is related or a coincidence.

bjornoleh:
@sulkaharo I believe this was mentioned in CGM in the cloud quite recently. Not sure why the issue was closed just now, it still could use some attention.

https://github.com/nightscout/cgm-remote-monitor/issues/6916 mentions a possible fix for that particular issue:
_COB: buggfix to show history COB when scrolling the timeline backwards_

naromero77:
It looks like the bug fix should go in the DIY software which is uploading information into NS. Thus the fix would have to go into OpenAPS, AndroidAPS, and Loop.

sulkaharo:
[...]

PR #8081, opened 2023-08-21
Fix to display historic COB-value (from OpenAps) when scolling back in time

bjornoleh:
Copied from https://github.com/nightscout/cgm-remote-monitor/commit/7aa606ddac735fce821119e5bb674c08bc8441f0 by @klalle / @maja-lofgren

Please see https://github.com/nightscout/cgm-remote-monitor/issues/6916 and https://github.com/nightscout/cgm-remote-monitor/issues/5241 for details.

### DIFF
--- a/lib/plugins/cob.js
+++ b/lib/plugins/cob.js
@@ -282,9 +282,10 @@ function init (ctx) {
       var when = new Date(lastCarbs.mills).toLocaleString();
       var amount = lastCarbs.carbs + 'g';
       info.push({ label: translate('Last Carbs'), value: amount + ' @ ' + when });
+      displayCob = Math.round(sbx.properties.openaps.lastSuggested.COB);
     }
 
-    sbx.pluginBase.updatePillText(sbx, {
+    sbx.pluginBase.updatePillText(cob, {
       value: displayCob + 'g'
       , label: translate('COB')
       , info: info


bjornoleh:
The CI test failed at https://github.com/nightscout/cgm-remote-monitor/actions/runs/5924138923/job/16061111728#step:7:5346

COB
       set a pill to the current COB:
     TypeError: Cannot read properties of undefined (reading 'lastSuggested')
      at Object.updateVisualisation (lib/plugins/cob.js:8:3117)
      at Context.<anonymous> (tests/cob.test.js:95:9)
      at processImmediate (node:internal/timers:466:21)

bjornoleh:
I don’t know if the proposed fix is a good one or a hack. But it has been working well for us using iAPS (previously called FreeAPS X). It should also work with AAPS, which is what this fix was targeted at by the original author. I don’t know if it would work with iOS DIY Loop

JohnKitching:
Is this still being worked on?
I'm not sure from the comments what it's waiting on.

AndyLow91:
Thanks for putting this together and for carrying over the earlier fix from #6916.

I’ve opened #8492 as an alternative implementation for the same underlying issue. The main reason is that this PR fixes the symptom through the OpenAPS path specifically, but the same historical COB problem can also affect Loop-style `loop.cob` data.

The root cause appears to be in the COB plugin’s freshness check: when scrolling back, Nightscout was comparing the historical device-status COB timestamp against the real current time, so valid historical COB could be treated as stale and replaced by treatment-derived COB.

#8492 changes that check to compare against the timeline time being viewed instead. It keeps using the existing device-status COB parser, so it covers both OpenAPS/iAPS/AAPS-style COB and Loop-style COB. I also added regression tests for both paths, plus a fallback test for genuinely stale historical COB.

So I think #8492 supersedes this PR, but this work was useful prior art and helped identify the right area to fix.

bjornoleh:
Awesome, thanks for tackling old issues!

Issue #8079, opened 2023-08-17
Nightscout should not use Google Fonts


gardenrobot:
I propose Nightscout should self-host fonts, instead of using a 3rd party to serve them.

I have run into issues with my homeserver where Nightscout became unusable due to an internet outage. If Google Fonts is not available for whatever reason, it would break Nightscout. There is also the issue of privacy, as Google uses fonts to collect data.

As far as I know, using Google Fonts doesn't offer much benefit to a project like this, where deployments are for a small set of users. I could be missing something though.

I have checked the licensing and Ubuntu and Open Sans both allow for them to be used in the repo.

If this change sounds good, I can implement this and create a PR. I have already created a quick POC. An alternative could also be to have self-hosted fonts as a fallback to Google.

sulkaharo:
Sure, if you can check and show the license allows for this, please make a PR. And please check how the caching of the fonts works, one reason to use the Google fonts is they're cached quite well and using them improves the site responsiveness most of the time.

gardenrobot:
Sounds good. The PR has been made. Is there a process or someone to ping to get a review?

gardenrobot:
@sulkaharo 
Hello, I want to follow up with this. A few people have expressed wanting this. Is there there anything I can do to get this merged in?

PieterGit:
Hi @gardenrobot . Sorry for the lack of communication. @sulkaharo has not been really active lately and fortuately @bewest and @AndyLow91 are doing a lot to get Nightscout back into shape. If they agree this might be a PR that can be merged. But if I am correctly the PR must be expanded so that it's fully legally ok (I am not a lawyer).

From a legal/compliance perspective, the bundled font files should be treated as third-party assets with their own licences, not as part of the project’s main licence.

Could you please add a short third-party licence/notice section to the README that explicitly states:

- `static/font/ubuntu/*` is licensed under the Ubuntu Font Licence 1.0;
- `static/font/open-sans/*` is licensed under the SIL Open Font License 1.1;
- the font files were copied unmodified from their upstream source, or otherwise describe any conversion/subsetting;
- the project’s main licence does not relicense these font files.

This avoids ambiguity, because the Ubuntu Font Licence requires the font software to remain under that licence and not be distributed under another licence.

gardenrobot:
Hi @PieterGit , I've updated the PR with the changes. I am also not a lawyer but I believe everything is ok now.

[...]

PR #8084, opened 2023-09-01
Replace Google Fonts

gardenrobot:
This PR removes Google Fonts (from [this issue](https://github.com/nightscout/cgm-remote-monitor/issues/8079)) and bundles the used fonts (Ubuntu and Google's Open Sans distribution) into the repo. This makes Nightscout useable if access if Fonts is interrupted.

I took care to make sure there is no change to appearance. The only change is the italicized text in /admin. It now uses the italic font instead of slanting the normal font.

before:
before

after:
after

The woff2 files cache, and I have not found a noticable difference in load time.

## Licensing

Both fonts allow distribution in open source projects:

https://github.com/googlefonts/opensans/blob/main/OFL.txt
> The OFL allows the licensed fonts to be used, studied, modified and redistributed freely as long as they are not sold by themselves. The fonts, including any derivative works, can be bundled, embedded, redistributed and/or sold with any software provided that any reserved names are not used by derivative works. The fonts and derivatives, however, cannot be released under any other type of license. The requirement for fonts to remain under this license does not apply to any document created using the fonts or their derivatives.

https://ubuntu.com/legal/font-licence
> This licence allows the licensed fonts to be used, studied, modified and redistributed freely. The fonts, including any derivative works, can be bundled, embedded, and redistributed provided the terms of this licence are met. The fonts and derivatives, however, cannot be released under any other licence. The requirement for fonts to remain under this licence does not require any document created using the fonts or their derivatives to be published under this licence, as long as the primary purpose of the document is not to be a vehicle for the distribution of the fonts.


static/api-docs.html
```diff
@@ -4,7 +4,6 @@
 <head>
   <meta charset="UTF-8">
   <title>Swagger UI: Nightscout API</title>
-  <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Source+Code+Pro:300,600|Titillium+Web:400,600,700" rel="stylesheet">
```
gardenrobot, line 7: As far as I can tell, this isn't used

t1diotac:
I self host NightScout with Podman, and I have tablets that display the clock view which are on a separate isolated network. This is an excellent change that makes it so devices on those isolated networks don't need any firewall exceptions for Google Font API servers.

CynthiaArmstrong:
I love this idea too. 👍

zetric:
I also vote for this.

Especially here in Germany there are still some data protection aspects in clarification when google fonts are automatically loaded from google servers which reveals the IP address of the requester to google (which is defined to be personal data as per GDPR). And if he did not consent to this, it can cost the website owner much money. In the past there were lawyers with automated crawlers who made a lot of money by sending written warnings (also legally a grey zone).

Just want to say that this point is also worth to be considered.

PR #8100, opened 2023-09-25
APIv3 sync only on srvModified

MilosKozak:


MilosKozak:
@bewest  as no other app is using v3 so far it's great opportunity to have the interface clear
it breaks nothing

PR #8108, opened 2023-10-16
This fixes the issue of a "stale" CGM value displaying indefinately.

gth001:
…It displays "Internet offline." until internet is restored.

This fix addresses the issue that with a lost internet connection, clock view continues to display a stale CGM value forever, with no clue that internet connectivity is lost. This fix shows "Internet offline." for the duration that is it lost, then returns to normal when internet is restored.


bjornoleh:
Hi @gth001, I saw your message on Discord. You should first change the target branch to `dev`. Then I hope your fix is a good one that can be merged, I have seen this issue myself.

There are lots of erroneous PRs being made to `master` when people try to update their forks. If you think your PR is being overlooked, you could try to ping one of the core developers like bewest or sulkaharo. Use @ to tag the usernames I mentioned.

gth001:
>You should first change the target branch to dev. Then you should also [...] merge dev into your branch.

Thank you, done. 

I have been using this fix for a year or two now, successfully. I have Amazon Fire tablets around my house always showing clock view, and sometimes they lose their wifi connection. Since railway stopped being free, I switched to fly.io but forgot to re-add this fix until I noticed a stale clock at home. So I forked, added the fix again, pointed to my fork, and deployed again to fly.io. Here is the fix in action:

https://github.com/nightscout/cgm-remote-monitor/assets/10224533/1738963c-d2c6-49ff-8e4b-b747ca453efb

(By the way @bewest I am George Francis on facebook, if you remember me, we chatted sometimes.)

bewest:
Howdy, this is a great idea, @gth001.  Thanks for this.  Apologies for the late response as there have been several personal barriers in the last year, although there are 7 maintainers listed besides myself.  @bjornoleh, thanks for your encouragement on this, also!

PR #8126, opened 2023-10-21
restore compat with socket.io v2 client

thecubic:
socket.io v4 won't talk to a socket.io v2 client unless you enable a compatibility flag 

https://socket.io/docs/v4/server-options/#alloweio3

tested directly in prod, version screenshot

image

PR #8127, opened 2023-10-21
socket.io: enable legacy v2 connections

bewest:
The allowEIO3, when set to true allows v2 clients to connect while the default is false. This patch should allow versions of AndroidAPS older than 3.2 to connect to NS15 and above.
Thanks to Dave Carlson for researching and experimenting with this option!
https://socket.io/docs/v4/migrating-from-2-x-to-3-0/#how-to-upgrade-an-existing-production-deployment https://socket.io/docs/v4/server-options/#alloweio3


bewest:
Same as https://github.com/nightscout/cgm-remote-monitor/pull/8126

PR #8252, opened 2024-04-17
Correct APIv3 handling of multiple filter parameters for a single field

bniels707:
Previously, only the last filter parameter for a field would be applied. After this change, filter parameters for fields are aggregated instead of being replaced.

This allows for range queries, ex: date$gt=1491719030000&date$lt=1491719930000

Additional tests were added to both verify the new behavior as well as to verify other documented API functionality, such as the use of ISO8601 dates.

### DIFF
--- a/lib/api3/storage/mongoCollection/utils.js
+++ b/lib/api3/storage/mongoCollection/utils.js
@@ -27,6 +27,7 @@ function normalizeDoc (doc) {
  */
 function parseFilter (filterDef, logicalOperator, onlyValid) {
   let filter = { };
+
   if (!filterDef)
     return filter;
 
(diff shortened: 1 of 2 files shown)


bniels707:
The API version should probably be incremented after accepting this, or it could be added to this MR.

ryceg:
Hi, I've been going through the backlog of pull requests, is this still relevant / worth merging? Thanks!

AndyLow91:
Thanks for this. I’m going to close this as superseded by #8253 

The underlying issue is still relevant, and the same-field filter fix from this PR is worth keeping. #8253 builds on this work by including the same core fix and adding support for the documented `filter_parameters` query argument as well.

We’ll use #8253 as the canonical PR/thread for the API v3 filter work, then rebase/adapt it against current `dev` so the tests match the newer API v3 search test structure.

PR #8253, opened 2024-04-18
APIv3: Support filter paramters passed via filter_parameters array

bniels707:
This contains two commits:

The first adds support for accepting filter parameters via the `filter_parameters` query argument. This makes it possible to actually pass filter parameters as described in the Swagger definition. The old behavior of passing filter parameters via arguments not explicitly named in the Swagger definition is preserved.

e.g. `filter_parameters=date$gt%3D1491719030000` == `date$gt=1491719030000`

The second commit makes it explicit in the Swagger definition that `filter_parameters` is an array of string arguments. The JSON definition was generated from the YAML definition (`java -jar openapi-generator-cli-7.4.0.jar generate -i cgm-remote-monitor/lib/api3/swagger.yaml -g openapi --skip-validate-spec`).

If someone knows of a way to generate the JSON that results in less churn (other than modifying it by hand...) I would gladly redo that commit.

This assumes #8252 has been merged.

The API version should probably be incremented after accepting #8252, this, or both.


AndyLow91:
I’ve rebased this PR onto current `dev` and pushed the updated branch.

What is now built into the PR:

- preserves multiple filter operators on the same field, e.g. `date$gt=...&date$lt=...`
- adds support for the documented `filter_parameters` query argument
- keeps the regular unlabeled filter parameter behavior working
- updates the OpenAPI definition for `filter_parameters` as a `spaceDelimited` string array
- aligns the API v3 runtime/OpenAPI version to `3.0.5`
- adapts the tests to the current shared API v3 search test harness
- preserves the ISO8601 date filter coverage from #8252, including timezone handling

I also reduced the generated `swagger.json` churn so the final diff is reviewable.

Validation run locally against MongoDB:

- `tests/api3.search.test.js`: 140 passing
- `tests/api3.basic.test.js`: 1 passing
- `eslint lib/api3/generic/search/input.js lib/api3/storage/mongoCollection/utils.js`: passing

This has not been merged into `dev`; it is only reflected on this PR branch.

Issue #7879, opened 2023-02-06
support basal injections (trends and logging)


rdpeake:
**Is your feature request related to a problem? Please describe.**
I am only just starting with nightscout, however one of the things that it is missing, and i have seen mentioned elsewhere is the ability to enter injections for basal, as intentioned data points that can be rendered, tracked, reported, and potentially in future count towards various statistic trackings (I'm not far enough into the UI to know what is useful there)

**Describe the solution you'd like**
Similar to being able to add meal bolus, or even temp basal entries via care portal. i would like to be able to select basal injection as an event type, enter a dose and on save, have it show as its own marker on the graph, and tracked in reports showing time of day markers on when it is taken, and dose statistics by time of day.

**Describe alternatives you've considered**
Two different features have been suggested - 1 is to use notes in the care portal, so it is at least being logged somewhere - though this doesn't allow the ability to make reports dependent on time of day taken, volume taken, or to try and spot tends (especially if meal tracking is being done such that you can spot a fasting period to determine if the basal value may need adjusting)

The other feature suggestion is to set the basal rate based on dosage over time - however i'm not entirely certain that is in anyway reflective of how non - pumped basal will actually work in the system, or give any meaningful statistics for doctors or people with diabetes. 

**Additional context**
There is a lot of 'would be cool' like features being mentioned in this FR, but it would just be useful to have a simple system that is independent of any existing basal features, that just allows you to log it, and report it, such that meaningful decisions can be made by viewing trends of doses, and times taken when compared to the existing BG trend reporting.

At a later date, trying to incorporate it into plugins and the like which can help predictions or other features of the system could then be considered.

Issue #6774, opened 2021-01-19
Support different injection types from xDrip+


BernardJu:
**Is your feature request related to a problem? Please describe.**
Currently it is assumed that the basal rate is always constant. Because I don't use a pump I inject basal once a day and inside xDrip I log it. (after enabling that in the settings)

However this messes up the nightscout reports as it adds it all together inside bolus. In the profile I can setup a constant value for basal which is the amount I inject divided by 24, however this clashes with the manual xDrip inserted log events which then skews the reports, making it less usable to present to a doctor, etc.

**Describe the solution you'd like**
I would like if nightscout could properly interpret the basal injection events as such instead of combining them all to simply insulin.

**Describe alternatives you've considered**
An alternative would be to not log basal and have it set up inside nightscout with my injection amount divided by 24, however this breaks my current workflow using xDrip.

**Additional context**
The events sent by xDrip are already available in treatments.json like this:
```js
{
    carbs: null
    created_at: "2021-01-17T05:10:28.000Z"
    enteredBy: "xdrip"
    eventType: "<none>"
    insulin: 76
    insulinInjections: "[{"insulin":"Actrapid","units":20.0},{"insulin":"Toujeo","units":56.0}]"
    sysTime: "2021-01-17T06:10:28.947+0100"
    timestamp: 1610860228947
    utcOffset: 0
    uuid: "1b3b6caf-18b4-467d-b95f-ee51569561bb"
    _id: "1b3b6caf18b4467db95fee51"
}
```

I am using Actrapid/Toujeo as they are similar to what I am actually using, though they are not exactly what I am using. The different types have different "strengths" / lengths how much the IOB is affected as well as if they are basal or bolus type so I would need to have that configurable like in xDrip+.

PR #8261, opened 2024-05-09
Multi insulin api

gruoner:
Patch provides a new entity of type "insulin" and provided that via a new API (copied from food API).
xDrip and Nightscout-reporter are using this API to store and fetch multiple insulin profiles (names, curves, activation within GUIs and color codes).
Currently those insulin profiles aren't respected in any way by nightscouts GUI and backends.


SeIex:
Open issues that might be related:
Support different injection types from xDrip+ #6774 
support basal injections (trends and logging) #7879

PR #8278, opened 2024-07-29
Sequence of Carb / Fat / Protein graph rendering

aug0211:
- No new code, only swap the sequence of two lines
- Formatting to use carbs / fat / protein sequence (instead of carbs / protein / fat)
- Matches the more generally used sequence of C/F/P to create less confusion as people read their charts


aug0211:
Been open for too long, no movement or discussion, closing this.

PR #8281, opened 2024-08-07
WIP: Increase rendering granularity to support AAPS TBRs

MilosKozak:


ghost (approved):

AndyLow91:
I tested this locally against current `dev` with the full CI test workload and Mocha timeouts disabled, so we could see the actual runtime impact rather than just the timeout failure.

Setup:
- Node v22.21.1 / npm 10.9.4
- MongoDB 7 with a clean test DB for each run
- Rebuilt the client bundle on each branch before testing
- Ran the CI test command with coverage using `tests/ci.test.env`

Results:

| Run | Full suite | `reports should produce some html` | `reports should produce week to week report` |
|---|---:|---:|---:|
| current `dev` | 54s | 1.775s | 3.713s |
| this PR | ~334s | 92.353s | 191.929s |

Both branches pass when timeouts are disabled, but this PR makes the report tests roughly 52x slower and adds about 280 seconds to the full CI run. The existing reports suite timeout is 80 seconds, so this explains the CI failure: the first report test alone now takes ~92s, and the week-to-week report takes ~192s.

In real-world terms, this would likely make longer reports very slow or unusable for users, and may also cause browser “page unresponsive” behavior on lower-powered devices. The main chart rendering path also changes basal rendering from 1-minute polling to 1-second polling, so wider chart ranges may get noticeably slower there too.

I agree with the goal of improving AAPS TBR rendering precision, but I do not think we should merge this implementation as-is. The costly part is sampling every second across each report day. A better approach would be to preserve second-level precision only around actual temp basal boundaries, or otherwise use a boundary/interval-driven calculation, rather than increasing the whole report loop from 288 samples/day to 86,400 samples/day.

AndyLow91:
I pushed `846bb69` to address the performance issue from the original implementation.

That commit changes the basal rendering approach so we no longer sample every second across the whole report/chart range. Instead:

- reports keep IOB/COB sampling at the existing 5-minute cadence
- basal rendering now includes exact temp basal/profile/combo boundaries
- basal insulin totals are calculated over intervals between those boundary/sample points
- the live chart uses the same boundary-aware basal sample list
- `getTempBasal()` cache keys now use the exact timestamp, so mid-minute basal changes are not hidden by minute-rounded cache entries

This keeps the intended AAPS TBR precision improvement while avoiding the earlier report slowdown. In local testing, the full CI workload returned to baseline timing: the patched branch completed in `54s`, compared with `~334s` for the original PR version, and all current GitHub CI test jobs are passing.

PR #8292, opened 2024-09-14
Fix nightscout crashes after failed APNs delivery

AndyLow91:
Updated the loop.js file to handle the error more gracefully and display an error message to the user rather than crashing the Nightscout instance.

### DIFF
--- a/.github/workflows/image-build.yml
+++ b/.github/workflows/image-build.yml
@@ -1,0 +1,26 @@
+name: Build Docker Image
+on:
+  push:
+    branches:
+      - upstream/wip/nspro/dev
+jobs:
+    build:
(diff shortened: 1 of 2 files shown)


AndyLow91:
tested by creating a new image on dockerhub with a sample of NSPro users. Their sites are no longer crashing and are handling the error gracefully, displaying to the user when the problem exists.

AndyLow91:
@bewest  @sulkaharo  - I'm not sure who has the correct level of access to merge these PR's. Any Insights?

bewest:
Looks good, can we move the workflow to a different branch/PR?

ryceg:
Can this be closed / merged?

PR #8296, opened 2024-09-24
return generic '500' status on bootError page

daaanosaur:
Currently Nightscout Web Monitor does not return a HTTP error status code when displaying the `bootError` page. 

This means that while users can see there is a problem with the application, platforms it may be deployed on cannot.

Returning a 500 status error in this case will at least enable health checks to be performed and deployments to be restarted when there's an issue with the application.

### DIFF
--- a/lib/server/booterror.js
+++ b/lib/server/booterror.js
@@ -29,7 +29,7 @@ function bootError(env, ctx) {
       return '<dt><b>' + obj.desc + '</b></dt><dd>' + message.replace(/\\n/g, '<br/>') + '</dd>';
     }).join(' ');
 
-    res.render('error.html', {
+    res.status(500).render('error.html', {
       errors,
       locals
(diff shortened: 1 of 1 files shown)


lib/server/booterror.js
```diff
@@ -29,7 +29,7 @@ function bootError(env, ctx) {
       return '<dt><b>' + obj.desc + '</b></dt><dd>' + message.replace(/\\n/g, '<br/>') + '</dd>';
     }).join(' ');
 
-    res.render('error.html', {
+    res.status(500).render('error.html', {
```
daaanosaur, line 32: https://expressjs.com/en/api.html#res.status

This returns a `500 error` http status in the header, rather than the default `200 ok`

bewest: Good idea, I love it.

PR #8301, opened 2024-10-05
Update sensorage.js for 10 day sensors

Nightfoxy:
Changing default SAGE warn and urgent threshold defaults to be useful for 10 day sensors rather than 7 day sensors from the G5 era. SAGE pill turns yellow at day 9 and red 4 hours before expiration of 10 days.

Once G6 is retired, consider setting WARN to 10 days and URGENT to 10.5 days.

### DIFF
--- a/lib/plugins/sensorage.js
+++ b/lib/plugins/sensorage.js
@@ -16,9 +16,9 @@ function init(ctx) {
 
   sage.getPrefs = function getPrefs(sbx) {
     return {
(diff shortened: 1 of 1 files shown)


bewest:
Nice, I would also consider adding something like "default cgm model."  We are a cgm-centric project, so that might help simplify things for many users although it requires adding much more details about each specific model to the code base.

PR #8307, opened 2024-10-25
Use local timezone date in careportal. Fixes nightscout#8304

### LINKED ISSUE #8304: Logged Treatment has incorrect date. Advances on GMT rather than local time zone.
**Describe the bug**
When logging a treatment the date is changing based on GMT rather than on the local time zone. For us this means treatments logged after 6pm will have tomorrows date.

* Date has rolled over in GMT *
1. Go to '+'
2. Click on 'Event Time'
3. Enter a time
4. Date will be for GMT, not local

**Expected behavior**
Date would be local

**Screenshots**

**Your setup information**
- Version 15.0.2
 
**Additional context**
Using Heroku to roll back to Version 14.2.6 fixes issue.

adamlounds:
Resolves #8304

Not sure what the 50ms sleep (see line 105 copied from line 61) is for - tests pass on my machine without it. Should I uncomment it or remove it?

### DIFF
--- a/lib/client/boluscalc.js
+++ b/lib/client/boluscalc.js
@@ -45,9 +45,9 @@ function init (client, $) {
   }
 
   function setDateAndTime (time) {
-    time = time || new Date();
-    eventTime.val(time.getHours() + ":" + time.getMinutes());
-    eventDate.val(time.toISOString().split('T')[0]);
+    time = time ? (time.format ? time : client.ctx.moment(time)) : client.ctx.moment();
+    eventDate.val(time.format('YYYY-MM-DD'));
+    eventTime.val(time.format('HH:mm'));
   }
 
   function mergeDateAndTime () {
@@ -585,7 +585,7 @@ function init (client, $) {
     maybePrevent(event);
   }
 
-  function quickpickHideFood () {
+  function quickpickHideFood (event) {
     var qpiselected = $('#bc_quickpick').val();
 
     if (qpiselected >= 0) {
--- a/lib/client/careportal.js
+++ b/lib/client/careportal.js
@@ -18,8 +18,8 @@ function init (client, $) {
 
   function setDateAndTime (time) {
     time = time || client.ctx.moment();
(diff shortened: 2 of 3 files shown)


AndyLow91:
Thanks for the original fix here. I’ve updated the PR branch so it is rebased onto the current `dev` branch, then added a follow-up commit to strengthen the regression coverage for #8304.

The test now exercises the full frontend path described in the issue:

- sets a fake local time where the local date differs from the UTC date
- verifies the careportal drawer displays the local `eventDateValue` and `eventTimeValue`
- selects the “Other” event time path
- calls `save()` and captures the treatment POST payload
- asserts `created_at` is built from the local date, not the UTC-shifted next day

Focused validation passes locally:

```text
./node_modules/.bin/env-cmd -f ./tests/ci.test.env ./node_modules/.bin/mocha --timeout 5000 --require ./tests/hooks.js --exit ./tests/careportal.test.js

2 passing
```

This should give us coverage for both the visible date bug and the submitted timestamp regression from #8304.

AndyLow91:
I’ve added one more follow-up on this PR to cover the two remaining timezone-risk areas we discussed.

Additional changes:

- added a second careportal regression case for a local timezone ahead of UTC, so both UTC-boundary directions are covered
- applied the same local `YYYY-MM-DD` / `HH:mm` handling to the boluscalc date/time fields
- added a boluscalc other-time regression that verifies the submitted treatment event time is built from the local date, not the UTC-shifted date
- fixed a strict-mode cleanup issue in the boluscalc submit path where `quickpickHideFood()` referenced `event` without declaring it

Focused validation:

```text
./node_modules/.bin/env-cmd -f ./tests/ci.test.env ./node_modules/.bin/mocha --timeout 5000 --require ./tests/hooks.js --exit ./tests/careportal.test.js

4 passing
```

Lint also passes for the changed files:

```text
./node_modules/.bin/eslint lib/client/boluscalc.js tests/careportal.test.js
```

PR #8234, opened 2024-03-04
Fix missing mmol unit conversion for lastEnacted.bg in openaps.js

dsnallfot:
**Issue:** 
The OpenAPS pill always shows enacted bg in mg/dl and does not convert units to mmol/l when mmol/l is the chosen user setting. 

After some troubleshooting together with @bjornoleh and @JeremyStorring today we found out that there currently is no mmol units conversion in the code for prop.lastEnacted.bg. All other bg conversions in openaps.js works as expected.

**Solution:** 
This PR adds the units conversion for prop.lastEnacted.bg, and resolves the issue.

**Testing**
The issue was identified while using iAPS as devicestatus uploader, and the code change have been tested to work with iAPS as uploader, but are most likely relevant for openAPS uploads in general (not tested).

Please review when suitable for you, and get back with any questions @sulkaharo @bewest

PR #8321, opened 2024-12-13
Wip/bewest/collaborations

bewest:
From https://github.com/nightscout/cgm-remote-monitor/pull/8234 with tweak to pass tests.


Navid200:
Thanks for this.
Do I understand correctly that there is no limitation to Node 16 any longer?

Can I ask which newer versions of Node have been verified and are being used?
Is any particular version advised?

Thanks

PR #8324, opened 2025-01-12
Fixed the OpenAPS pill and forecast lines from not showing

yodax:
This is my local NS on the left and on the right is current main running on the same database.
msedge_rGyicw3aGf

Every so often the OpenAPS pill would stop updating and the forecast lines would disappear. I created a test to reproduce this using the data from my mongo database. This allowed me to narrow it down to the fix below. The lastEnacted field didn't always have a "rate" field set in each of these json objects.

### DIFF
--- a/lib/plugins/openaps.js
+++ b/lib/plugins/openaps.js
@@ -471,16 +471,26 @@ function init (ctx) {
     }
 
     if ('enacted' === prop.status.code) {
-      var canceled = prop.lastEnacted.rate === 0 && prop.lastEnacted.duration === 0;
+      var basalRate = Number(prop.lastEnacted.rate);
+      var hasBasalRate = Number.isFinite(basalRate);
+      var hasDuration = prop.lastEnacted.duration !== undefined && prop.lastEnacted.duration !== null;
+      var hasTempBasalDetails = hasBasalRate && hasDuration;
+      var canceled = hasTempBasalDetails && basalRate === 0 && prop.lastEnacted.duration === 0;
       var bg = displayBg(prop.lastEnacted.bg);
 
       var valueParts = [
         valueString('BG: ', bg)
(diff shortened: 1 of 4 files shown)


bewest:
Very very nice! :tada:

PieterDeBruijn:
Would be nice if this fix was merged!

yodax:
It’s been more than a year since the opening of the PR; a release was just made, the PR not included. 

I still see people complaining about this not working for them. The fix works and has tests associated with it. (I have been running the code for all that time without a single forecast line disappearing)

If there’s no intention to release it, I’m ok with that of course. I would suggest closing the PR.

AndyLow91:
Thanks for this fix and for adding the regression data. The crash path is real, and the test does reproduce it.

One issue before merge: the current fallback changes the exception into malformed pill text for the new fixture. Because `enacted` is missing both `rate` and `duration`, the pill renders:

`Temp Basal Started 0 for undefinedm`

That is still user-visible and a little misleading, especially since the reason text says `no temp required`.

Could we make the display conditional instead? For example, only append the `rate for duration` phrase when both `rate` and `duration` are present, otherwise omit that phrase and rely on the reason text. It would also be worth extending the new test to assert the generated pill text does not contain `undefined`.

AndyLow91:
Thanks again for the fix and the regression data. I pushed two small follow-up commits to cover the display edge cases from review:

- missing `rate`/`duration` now falls back to the BG/reason text instead of rendering malformed temp-basal text
- `null` and blank temp-basal fields are no longer coerced into `0`
- the regression test now asserts the pill text does not contain `undefined` or misleading `Temp Basal Started` text when details are missing

I re-reviewed the latest head and didn’t find any remaining blockers. CI is green, and the focused OpenAPS tests/lint pass locally.

PR #8326, opened 2025-01-19
ode V20 and apt-get to apt

hershyheilpern:
update to node V20 and apt-get to apt

### DIFF
--- a/bin/setup.sh
+++ b/bin/setup.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
-sudo apt-get update
-sudo apt-get install -y nodejs build-essential git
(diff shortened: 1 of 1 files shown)


hershyheilpern:
Is there a issue with this

ryceg:
I think we're yet to decide on which version of node we plan to upgrade to- there's a couple active pull requests already open.

AndyLow91:
Picking this back up while going through open PR's.

I’ve pushed a small follow-up commit to this PR: `3df5c6d4`.

The NodeSource change from `setup_8.x` to `setup_20.x` is the right fix now that Nightscout requires Node 20 or newer.

I adjusted the package-manager commands back from `apt` to `apt-get`, because `apt-get` is the better interface for non-interactive setup scripts. So this PR now keeps the Node 20 update while preserving the safer scripted install style.

I also checked the script syntax with `sh -n bin/setup.sh`.

PR #8330, opened 2025-02-12
Add GMI and Revised GMI to glucose distribution report

motinis:
Sources:
- https://web.archive.org/web/20250206062319/https://pmc.ncbi.nlm.nih.gov/articles/PMC6196826/pdf/dc181581.pdf
- https://web.archive.org/web/20250210181811/https://pmc.ncbi.nlm.nih.gov/articles/PMC11571357/


motinis:
@bewest - would it be possible to get this PR reviewed? Thank you.

AndyLow91:
Thanks for adding this. I like the direction, but I found an issue that should be resolved before merge:

1. The GMI values are currently calculated from `glucoseMean`, which is built inside the stability loop and only includes records that have a following sample within `maxGap`. That means segment-end readings, readings before gaps, and some hour-filter boundary records are excluded. GMI should use the mean of all selected CGM readings, similar to the existing `mgDlBgs` mean used for A1c estimation.

PR #8348, opened 2025-05-02
Remove Moment

ryceg:
This pull request removes the MomentJS dependency, and replaces it with the much lighter DayJS, which has a largely identical API. 

This has resulted in a package bundle size reduction of 630Kb. DayJS is not tree-shakeable, and we could get an ever so slightly smaller bundle size if we went with date-fns, but I figured that the 1:1 nature of the migration makes DayJS the clear winner.

Instead of completely remove the `.moment` api from the context, I've added `.dayjs()` in case there are some corners of the project which I've missed- I've Ctrl+F'd and replaced it wherever found it (but did NOT find and replace to change the dependency, as that would overwrite properties in `loop.js`.


sulkaharo:
I tried this earlier and at that point bumped into issues with DayJS having significant issues with date math & time zones. Seems some of this has been fixed, but curious to hear how extensively you've tested this? See https://github.com/iamkun/dayjs/issues/1805

ryceg:
I've only done preliminary testing- haven't run it through its paces yet, figured it would be useful to gather some feedback prior. Sounds like I was right. The idempotency issue is definitely something I could've missed, I'll let you know how I go with it!

dnzxy:
> I tried this earlier and at that point bumped into issues with DayJS having significant issues with date math & time zones. Seems some of this has been fixed, but curious to hear how extensively you've tested this? See [iamkun/dayjs#1805](https://github.com/iamkun/dayjs/issues/1805)

We have gone through similar issues at work, where we heavily rely on moment for a very date arithmetic and timezone reliant heavy software bundle we maintain. Careful with just swapping this, especially if (like here) there was not just reliance on moment but also moment-tz.

ryceg:
The only remaining test that's failing is the report-building. For some reason, it's returning `No data available` in the day-to-day report, but in use, it seems to be fine? 
image

ryceg:
Minor question @bewest; currently, the API rejects illegal date formats; 

```ts
it('should reject invalid date - illegal format', async () => {
    let res = await self.instance.post(self.url, self.jwt.create)
      .send(Object.assign({}, self.validDoc, { date: '2019-20-60T50:90:90' }))
      .expect(400);

    res.body.status.should.equal(400);
    res.body.message.should.equal('Bad or missing date field');
  });
```

However, DayJS is a little more lenient than MomentJS is with its parsing- it can parse that as a valid date, `2020-10-01T03:31:30`. The dayjs api isn't flexible enough to let us disallow correcting those "overflows" while keeping the rest of the parsing, so our other option is spelling out every "allowed" date format that we want to allow in. Considering the relative infrequency of date malformation via overflows, what's your feeling on removing the assertion that it rejects an overflowed date?

AndyLow91:
Thanks for the work on this proposal. We are considering options to potentially replace or reduce Moment as part of M27, the date/time workstream in the Nightscout modernization plan tracked by #8605.

We have now explicitly added **Day.js** to that evaluation alongside retaining/narrowing Moment, native `Intl` with scoped helpers, and Luxon. No replacement has been selected, and Day.js remains a candidate. `Intl` alone is not a complete replacement for parsing and date arithmetic.

We will compare the candidates against Nightscout's API parsing, profile schedules, IOB/COB, reports, timezone/DST and locale requirements, and measure bundle size, performance and runtime memory against the current modernization build. This PR provides useful groundwork for that review. Our initial inspection identified invalid-date normalization and DST calendar-day arithmetic cases that we have recorded as regression requirements; these need addressing during evaluation rather than automatically ruling Day.js out.

The updated [M27 plan](https://github.com/nightscout/cgm-remote-monitor/blob/chore/nightscout-modernization/docs/plans/nightscout-modernization.md) and [Day.js evaluation notes](https://github.com/nightscout/cgm-remote-monitor/blob/chore/nightscout-modernization/docs/plans/moment-timezone-review.md#dayjs-candidate-from-8348) describe the scope. We will make the retain/narrow/replace decision when we reach that phase, based on compatibility and measured benefit.

dnzxy:
@AndyLow91 @ryceg just wanting to note that with the [Temporal API proposal](https://github.com/tc39/proposal-temporal) finally having reached Stage 4, and given how long this PR has sat stale by now, it may be easier waiting for Temporal to become fully GA vs. moving from one library to another library. 

Just my `/2cents` though.

AndyLow91:
Thanks @dnzxy - I wasn’t aware of Temporal as an option, so thanks for bringing it to the table. We’ll add it to the M27 plan for consideration. When we reach that part of the modernization work, we can assess the available options and runtime/browser support at the time, decide which approach best fits Nightscout, and go from there.

dnzxy:
Sure, no worries!

I only know of it because we‘ve been waiting for it to become GA at my day job for like 3+ years now 😂

PR #8354, opened 2025-05-09
Add GitHub action to close accidental "sync" PRs

ryceg:
Pretty straightforward, and not _too_ eager, this closes PRs if three conditions are met:
* The title has one of the words in this array `["sync", "merge", "update", "pull", "new", "."]`
* There's no changed files or additions
* And it's not from the head repo

Given that there's not a lot of other reasons to make a PR with 0 changes, it _should_ be relatively safe. Not tested since nobody can be bothered to set up Act, but it should be okay.

### DIFF
--- a/.github/workflows/close-accidental-sync-prs.yml
+++ b/.github/workflows/close-accidental-sync-prs.yml
@@ -33,37 +33,19 @@ jobs:
             const message = `
 Hi! 👋 It looks like you're trying to sync your fork's \`${baseBranchName}\` branch with the \`${baseRepoFullName}\` repository.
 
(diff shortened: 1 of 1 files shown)


AndyLow91:
Thanks @ryceg for the original PR and premise here. I couldn't push the reviewed fixes back to `ryceg:dev` because both direct git push and the contents API returned permission errors, despite maintainer edits being reported as enabled. This PR is also based on `master`, while the replacement needs to be based on current `dev`.

I've opened #8507 from `nightscout:wip/close-accidental-prs` against `dev`, carrying this work forward and crediting you in the PR body and commit trailer. Closing this one in favor of #8507.

PR #8355, opened 2025-05-10
Remove crypto browserify library

ryceg:
This goes a step further than nightscout#8329 and instead uses the browser's native crypto library, or the node built-in for serverside.

This results in a bundle reduction of 0.87mb.

All tests are still passing. I opted not to change the SHA1 encoding- I saw that there's a SHA512 secret in enclave.js, so I assume that sha1 wasn't used just for lack of knowledge about its lack of cryptographic security?


je-l:
Using the browser crypto library makes sense. It's a small change, but as the bundle size drops, it should result in a visible performance improvement.

ryceg:
Yep! Moving towards native methods will definitely help. Moment and lodash are also on the chopping block.

ryceg:
Passes the test suite.

bewest (approved):
LGTM.

AndyLow91:
Superseded by #8506. I opened the replacement PR from an upstream branch because GitHub would not allow me to push the conflict-resolution commits back to this fork branch, despite maintainer edits being shown as enabled. The replacement keeps the original crypto-browserify removal and resolves the current `dev` conflicts.

PR #8360, opened 2025-05-19
Remove Lodash

ryceg:
This is a chonky PR. Removes all of Lodash. Reduces bundle by about half a megabyte. All tests pass, although obviously I haven't battle-tested it yet; would be great to beta test with actual users.


je-l:
After a quick testing with real data, I didn't find any obvious bugs. I looked at the front page graph, and the various tabs in Nightscout reporting.

lib/admin_plugins/roles.js
```diff
@@ -35,7 +33,7 @@ roles.actions = [{
 
 function createOrSaveRole (role, client, callback) {
 
-  var method = _.isEmpty(role._id) ? 'POST' : 'PUT';
+  var method = !role._id ? 'POST' : 'PUT';
```
sulkaharo, line 36: Not sure matters in this case, but isEmpty() is safe in case role is null, but the latter will throw due to accessing a property of a null object

lib/api3/generic/search/operation.js
```diff
@@ -43,11 +42,9 @@ async function search (opCtx) {
       , onlyValid });
 
     if (!result)
-      throw new Error('empty result');
+      throw new Error('empty result');    result.forEach(col.resolveDates);
```
sulkaharo, line 45: Accidental add? The forEach doesn't do anything

ryceg: No, the `resolveDates` method is defined in lib/api3/generic/collection.js and needs to be called on each of the results.

lib/admin_plugins/roles.js
ryceg: Good catch! Made these changes in commit 4387bcedd2a5c220427963e68fec856f55d62188

AndyLow91:
closing this as superseded by #8502 - I ran into some github auth issues that wouldn't allow me to resolve conflicts within this branch and make changes.

PR #8362, opened 2025-05-23
Use lower authorization delay to speed up test-ci script by 20%

je-l:
The lib/authorization/index.js module implements throttling for authorization. The test-ci script is using the default delay of 5000ms, which slows down tests. Using the same value as in docs/example-template.env should be sensible.

This change reduces test-ci script execution time from 66.1s to 52.2s on my laptop.


AndyLow91:
Thanks for this. We can close this as already addressed.

The core change from this PR was adding `AUTH_FAIL_DELAY=50` to `tests/ci.test.env` so CI tests don’t pay the default authorization failure delay. Current `dev` already includes that setting, and later test/performance work also touched the authorization delay behavior.

This branch is now behind current `dev` and should not be merged as-is, since the current test env has moved on. Thanks for spotting the slowdown and proposing the fix.

PR #8366, opened 2025-05-28
Wip/2025 reports

dburren:
Fixes some old bugs, and extends many of the report pages.
This is a refinement of the customisations we (David Burren and Nascence Biomed) have been using in-project with v14 for some time.

### Timeframes:
3-week and 2-month presets available.
Plus for every timeframe preset we can select the period BEFORE that.
14 days now the default.

### Now two BG ranges:
Goal (inner) range from BG_TARGET_BOTTOM and BG_TARGET_TOP.
Outer range from BG_LOW and BG_HIGH.
If the ranges are set to be the same, glucose values are split into 3 groups as before. But now we can have more

### Day to day
1200x400px now the default.
Both low and both high lines are shown.

### Insulin distribution:
The delivery of insulin via "basal" or "bolus" mechanisms is not relevant to profile analysis.
Now highlights TBB and TDD rather than mechanical basal/bolus.
Pie chart shows base basal rather than basal.
More decimal places in report (supporting high-precision pumps and low-volume users).

### Week to week
3-5 glucose brackets supported (dependent on goal and outer boundaries). With coloured headings.
CV added to StDev column.

### Percentile
Green background for in-range area.
Both high and both low lines shown.

### Distribution
GMI as well as eHbA1c displayed.
PGS dropped (it was always dependent on assumptions of the "Normal" range and was deeply flawed).
Displays CV ("relative SD") along with SD.
GRI available in the extended report.

Table rows always start with high and progress to low glucose values. Previously "low" was above "high".

Fixed some double-counting of start/end records.
Summarises overall data points per day (there are 288 5-minute samples in 24 hours). But because we don't know the user is using a 5-minute CGM we don't translate this to a capture percentage.
Old code counted regarded values "at" the high boundary (e.g. 180/10.0) as high, instead of the accepted "within range"). First column should make it clear how the boundaries are applied.

3 report styles available:

- Custom:
Between 3-5 brackets (depending on goal/outer range boundaries).
If a range matches TIR or TITR it is labelled as such.
- Standards:
All bracket boundaries hard-coded based on industry use.
All known manufacturer/study boundaries are included.
TITR and TIR are both shown.
- Standards (detail):
Adds a TING column, and displays GRI.

### General
Max report timeframe extended to one year (mostly useful for Distribution).
CSS update so colours in tables now transfer to printed copies.
New strings have been added to the end of the en translation.json, but not updated into other languages.

DistributionSample
Screenshot 2025-05-28 at 17 26 57


ryceg:
This looks great, David. As an aside, I'm working on an experimental branch using Svelte which replaces the frontend, and I'd be curious about what sort of reports you think would be useful that aren't currently implemented in NightScout- much easier to set them up in modern development environments compared to JQuery!

AndyLow91:
CI is currently failing because the existing report test still expects the old Daily Stats HTML output.

The failure is in `tests/reports.test.js:426`, where the test looks for the previous 3-bucket Daily Stats layout:

`<td class="tdborder">0%</td><td class="tdborder">100%</td><td class="tdborder">0%</td><td class="tdborder">2</td>`

This PR intentionally changes that report to use the new outer/goal range buckets and one-decimal percentages, so the generated HTML no longer matches that old assertion.

I think the next step is to update the report test to cover the new output shape rather than treating this as a product bug. It would also be worth making the assertion a bit less brittle, since it currently depends on a long exact HTML substring.

AndyLow91:
Thanks for the work on this. I’ve got the PR into a clean, mergeable state now, and CI is passing.

Before merging, though, I’d like to get some wider user feedback on the report UI changes. This touches familiar reporting views, so I think we should be careful that the new behaviour is understandable for everyday users rather than only technically correct.

A few things I’m unsure about:

- The new date selector rows labelled `Last:` and `Prev:` are not immediately clear at first glance. I understand what they do after playing with them, but I worry the distinction may confuse users.
- I’m not sure about the colour choices in the Distribution report. Blue is not usually associated with high glucose values, and the subtle differences between the green shades may be difficult for colourblind users.
- I don’t love the labels `Lowish` and `Highish`. They are understandable in English, but they feel informal and may not translate cleanly into other languages.

Overall, I welcome report UI improvements and I think there is a lot of useful work here. I’d just like to hear from others before deciding whether we should merge this as-is, or ask for some UX/labelling/colour changes first.

je-l:
A side note about the default glucose thresholds in this pr. I see there are two values at 3.5 and 6.7 mmol/l, but a more widely used, strict, before-meal glucose target for type 1 diabetes is 4.4-7.2 mmol/l, see [table 1 in this paper](https://diabetesjournals.org/care/article/44/11/2589/138492/The-Management-of-Type-1-Diabetes-in-Adults-A). I would consider using this range as a reasonable default, so the new thresholds would be 3, 3.9, 4.4, 7.2, 7.8, ...

dburren:
Thanks Andy.


> * The new date selector rows labelled `Last:` and `Prev:` are not immediately clear at first glance. I understand what they do after playing with them, but I worry the distinction may confuse users.

I've since been playing with ideas of a 2-part selector, where at a click we can set the length to one of the convenient lengths, but we also have separate Forward/Back controls to move both ends of the range by a day/week at a time.

With the traditional way and with what went into this set, it's awkward for example to quickly select a week that doesn't include today (which is incomplete).


> * I’m not sure about the colour choices in the Distribution report. Blue is not usually associated with high glucose values, and the subtle differences between the green shades may be difficult for colourblind users.

I tried to make the colours consistent across the Distribution and Daily Stats reports.
I ended up with Blue as it was visually different (for my eyes, at least) plus it's quite different to the "red" hypo status that I regard as more alarming.
No I didn't design the colour scheme with various CVDs in mind. That would be a nice touch.


> * I don’t love the labels `Lowish` and `Highish`. They are understandable in English, but they feel informal and may not translate cleanly into other languages.

Fair comment, but they were explicity chosen to be non-accusatory. For the same reason "Normal" is no longer used.

dburren:
> A side note about the default glucose thresholds in this pr. I see there are two values at 3.5 and 6.7 mmol/l, but a more widely used, strict, before-meal glucose target for type 1 diabetes is 4.4-7.2 mmol/l, see [table 1 in this paper](https://diabetesjournals.org/care/article/44/11/2589/138492/The-Management-of-Type-1-Diabetes-in-Adults-A). I would consider using this range as a reasonable default, so the new thresholds would be 3, 3.9, 4.4, 7.2, 7.8, ...

Interesting. That "more widely used" range (80-130 mg/dL) is new to me, and is only referenced in that paper as a pre-prandial target. Does it have more or less relevance for SMBG monitoring than CGM?

Certainly I see 3.5 referenced a lot in endocrinology discussions around here, plus a bunch of us have got used to looking at 6.7. But I'm probably not "married" to that one: admittedly the only place I've seen 6.7/120 referenced in studies so far has been analysis of data from people without diabetes.

Introducing extra thresholds will affect the layout of the ranges in the distribution table (including which ones get grouped together for convenient comparison with vendor reports) so some redesign might be involved.

PR #8381, opened 2025-07-03
update API v1 swagger docs to add missing endpoints

arfaomar:
added previously undocumented API v1 endpoints to swagger doc.


AndyLow91:
Thanks for working on filling out the v1 Swagger docs. I think this is useful, but this PR needs some cleanup before it is merge-ready.

A couple of issues I found:

1. `/api/v1/versions` is documented, but it does not appear to be mounted by the v1 API router. The route exists in `lib/api/root.js`, but `lib/api/index.js` does not attach that router, so the documented endpoint currently 404s.

2. The response schemas for `/adminnotifies` and `/verifyauth` do not match the actual v1 response shape. Both handlers use `res.sendJSONStatus(...)`, so the response is wrapped as:

   ```json
   {
     "status": 200,
     "message": { ... },
     "description": null
   }
The Swagger docs currently describe the inner payload directly, which will mislead clients generated from the spec.

Also, the branch is now stale and conflicts with current dev in lib/server/swagger.json, so it will need a rebase/conflict resolution before it can be merged.

Once those are fixed, this should be in much better shape.

PR #8382, opened 2025-07-12
Create .render.yaml

shahakz11:
Add Render config to fix deployment

### DIFF
--- a/.render.yaml
+++ b/.render.yaml
@@ -1,0 +1,21 @@
+services:
+  - type: web
+    name: nightscout
+    env: node
+    buildCommand: ""
+    startCommand: "node server.js"
+    envVars:
+      - key: MONGO_CONNECTION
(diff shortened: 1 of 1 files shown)


AndyLow91:
Thanks for the PR, but I don’t think this is safe to merge as submitted.

A few blocking issues:

- `render.yaml` includes committed secrets, including a MongoDB connection string and `API_SECRET`. Those cannot live in the repository.
- The Render config itself does not look correct. `PORT` is being set to `$PORT`, but Render already provides `PORT`, and blueprint env var values are not the right place for that kind of interpolation.
- The server startup change is not isolated to Render. It removes the explicit `HOSTNAME` binding, which changes behavior for other deployments too.
- The PR also includes unrelated stray files (`ds`, `s`, `.bunignore`), which should not be part of a deployment fix.

The only part here that looks generally reasonable is parsing `PORT` as an integer, but that should be proposed separately in a small clean PR with no secrets, no unrelated runtime changes, and no extra files.

If you want to revisit Render support, please open a reduced PR that keeps secrets out of the repo and limits the scope to the actual deployment fix.

PR #8385, opened 2025-07-26
Add an icon to toggle wake lock of the display

earldouglas:
This adds the Wake Lock plugin, which adds a clickable sun/moon icon to the toolbar to toggle wake lock using the [Screen Wake Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API):

unlocked
<br>
locked

The plugin must be enabled by the user by setting `ENABLE=wake-lock`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md)
         - [`ar2` (AR2 Forecasting)](#ar2-ar2-forecasting)
         - [`simplealarms` (Simple BG Alarms)](#simplealarms-simple-bg-alarms)
         - [`profile` (Treatment Profile)](#profile-treatment-profile)
+        - [`wake-lock` (Wake Lock)](#wake-lock-wake-lock)
       - [Advanced Plugins:](#advanced-plugins)
         - [`careportal` (Careportal)](#careportal-careportal)
         - [`boluscalc` (Bolus Wizard)](#boluscalc-bolus-wizard)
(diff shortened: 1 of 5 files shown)


mstry9:
Just curious, would this be a plugin like care portal that you could choose to enable or not?

AndyLow91:
Thanks for working on this. The feature itself is useful, and gating it behind `ENABLE=wake-lock` makes sense.

I don’t think this is merge-ready yet because the toolbar affordance is too ambiguous. The new wake lock control uses a padlock icon in the same toolbar area where Nightscout already uses a padlock for `lockedToggle` when Care Portal actions require authentication. When wake lock is active it also switches to `icon-lock`, so an unauthenticated user can see two lock icons in the same place with different meanings: one means “screen wake lock is active”, the other means “token needed”. For a security-related icon, that ambiguity is worth avoiding.

I’d suggest using a distinct icon and/or placement/state treatment for wake lock so it is not visually confused with authentication status. A screen/display, power, or “keep awake” style icon would be clearer than reusing the auth lock metaphor.

I also noticed two implementation details worth fixing:

- `navigator.wakeLock.request('screen')` and `wakeLock.release()` should handle rejected promises. If the browser denies the request or release fails, the current requested state/icon can drift from the actual wake-lock state and may also produce an unhandled rejection.
- `static/css/main.css` adds a `moreicons` `@font-face` pointing at `/font/moreicons.woff` and `/font/moreicons.svg`, but those files are not present and the font family does not appear to be used. That looks like leftover asset churn.

Checks I ran:

- `eslint lib/client/index.js lib/client/wake-lock.js` passes.
- `git diff --check` passes.

So I’m in favor of the feature, but I’d like to see the icon ambiguity and promise error handling addressed before merge.

earldouglas:
Thanks for the feedback!  I'll update accordingly over the weekend.

earldouglas:
> The new wake lock control uses a padlock icon in the same toolbar area where Nightscout already uses a padlock for `lockedToggle` when Care Portal actions require authentication.

Fixed this by using a sun/moon icon instead of a padlock icon:
unlocked
locked

If you can think of better icons, please let me know.  I can't find a convention for wake lock icons.

> `navigator.wakeLock.request('screen')` and `wakeLock.release()` should handle rejected promises

Fixed this by adding `.catch(...)` calls to each.

> `static/css/main.css` adds a `moreicons` `@font-face` pointing at `/font/moreicons.woff` and `/font/moreicons.svg`, but those files are not present and the font family does not appear to be used. That looks like leftover asset churn.

These are no longer needed as this change no longer introduces any new icons (i.e. the unlocked padlock icon).

Issue #8262, opened 2024-05-15
Ability to view BG in clock view as mmol/l or mg/dl


BrightScreenTV:
In the clock view the page only seems to display mg/dl and doesn't respect the user's settings on the main dashboard

I'd like to propse that a toggle style button is available on the clock view (in whatever variation) that allows for switching between the two different units.

I'd be happy to contribute by forking the project and coding this feature in. On first glance, the quickest route seems to be to alter the lib/client/clock-client.js file although more research might provide a better solution in terms of defaulting to the user's saved units setting when the clock appears.

Please see attached a mock-up of my suggestion:

![units-mock](https://github.com/nightscout/cgm-remote-monitor/assets/67185910/765b2514-06b4-4a0e-9286-1851ea246572)

psonnera:
@BrightScreenTV I believe any development would be welcome! Thanks!

AndyLow91:
Thanks for raising this. I’ve checked the current `dev` branch and there is a real bug here, although I think the best fix is slightly different from the proposed UI toggle.

The clock view currently appears to respect the site-level `DISPLAY_UNITS` setting, because it uses the precomputed BG value from `/api/v2/properties`. However, it does **not** respect the browser/user unit override from the main dashboard.

So the current behaviour is roughly:

- If `DISPLAY_UNITS=mmol`, the clock view should show mmol/L.
- If `DISPLAY_UNITS=mg/dl`, the clock view should show mg/dL.
- But if the user changes their browser setting in the main dashboard, the main UI updates while the clock view can continue showing the site default.

That last point is the bug we should fix.

I don’t think we should add a separate unit toggle directly to the clock view as the primary solution, because that would introduce another place where unit preferences can drift from the main dashboard. The cleaner fix is for the clock view to follow the same unit preference rules as the rest of the UI:

1. Use the browser/user unit setting if one exists.
2. Fall back to `DISPLAY_UNITS` if there is no browser override.
3. Render both BG and delta consistently in the selected units.

There is an existing PR, #8388, but I think we should close that and replace it with a smaller PR focused only on this behaviour. #8388 includes unrelated changes and adds a clock-specific toggle, which is more than we need to resolve the confirmed bug.

I’ll leave this issue open for now in case there is further discussion. If the reporter, or anyone else, wants to create a smaller focused PR for the behaviour above, that would be very welcome. Otherwise I may come back to this after finishing review of the other open issues and PRs.

[...]

PR #8388, opened 2025-08-11
Wip/clock units

KelvinKramp:
#8262 

- Applied styling guidelines to testdatarunner.js
- Added unit toggle button to clock view
- Synchronized the units with the browser view units

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "printWidth": 80
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,7 +1,0 @@
-{
-  "singleQuote": true,
-  "trailingComma": "es5",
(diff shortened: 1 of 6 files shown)


AndyLow91:
Thanks for working on this and for taking the time to put together a proposed fix for #8262.

I’ve reviewed the current behaviour on `dev`, and I agree there is a real bug: the clock view follows the site-level `DISPLAY_UNITS` value, but it does not currently respect the browser/user unit override from the main dashboard.

However, I think we should solve that with a smaller, more focused change rather than adding a separate unit toggle directly to the clock view. The clock view should use the same unit preference rules as the main UI:

1. Use the browser/user unit setting if one exists.
2. Fall back to `DISPLAY_UNITS` when there is no browser override.
3. Render both BG and delta consistently in the selected units.

That avoids creating a second clock-specific setting that could drift away from the main dashboard.

I’m going to close this PR for now because it also includes unrelated changes outside the clock unit handling. The underlying issue remains open in #8262, and a smaller PR focused only on making the clock respect the existing browser/user units setting would be very welcome.

Issue #8048, opened 2023-06-30
Clock whit seconds


Marcino86:
Hi,
Can You add an option to Nightscout for the clock as a Environment?
Predefined values for your browser settings (optional) - after TIME_FORMAT, could there be an option to show secends?
![clock](https://github.com/nightscout/cgm-remote-monitor/assets/117569010/ea4b1495-3444-4231-863a-b9435e977b35)

samuelpert:
Hi, my name is Samuel, I am a student from a Codepath course called AI301 where we look for projects to contribute into open source. I am excited to help on this issue. I will take a look and later I will make a pull request with the expected solution.

PR #8392, opened 2025-08-14
Show seconds

KelvinKramp:
#8048

- show seconds when checkbox ticked in browser settings or SHOW_SECONDS in env vars is "on"
- increase interval to seconds when show_seconds = on

I did not change the clockview as I can't make the current time appear there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,7 @@ autonomy for your data:
 ### Predefined values for your browser settings (optional)
 
   * `TIME_FORMAT` (`12`)- possible values `12` or `24`
+  * `SHOW_SECONDS` (`off`) - possible values `on` or `off`
   * `DAY_START` (`7.0`) - time for start of day (0.0 - 24.0) for features using day time / night time options
   * `DAY_END` (`21.0`) - time for end of day (0.0 - 24.0) for features using day time / night time options
   * `NIGHT_MODE` (`off`) - possible values `on` or `off`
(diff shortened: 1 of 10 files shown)


AndyLow91:
Thanks for working on this. The overall direction looks right for #8048: adding `SHOW_SECONDS` as a default browser setting and exposing a browser-level checkbox makes sense.

I don’t think this is merge-ready yet, though. I found a few issues that should be fixed first:

- `lib/client/index.js` currently redeclares `var interval` in both branches of `updateClock()`, which fails lint with `no-redeclare`. This can be fixed by declaring `interval` once before the `if`, then assigning it inside each branch.
- `tests/settings.test.js` adds `SHOW_SECONDS` to the expected env var list, but the hard-coded count is still `24`. The focused settings test now fails with `expected 25 to be 24`.
- `azuredeploy.json` declares the new `show_seconds` parameter, but does not wire it into `siteConfig.appSettings` as `SHOW_SECONDS`. As written, Azure deployments would show the parameter but it would not actually affect Nightscout.
- `lib/client/browser-settings.js` has trailing whitespace, which is caught by `git diff --check`.

Verification I ran:

- `eslint lib/client/index.js lib/client/browser-settings.js lib/settings.js tests/settings.test.js` fails on the `interval` redeclaration.
- `mocha ... tests/settings.test.js` fails on the expected env var count.
- `git diff --check` reports the trailing whitespace.

Once those are addressed, this should be much closer to mergeable.

PR #8402, opened 2025-11-11
Added database CSV exports for individual and/or research use

kelseyhuss:
New Features:

Added CSV Export Buttons to Main Homescreen: Users can now export data directly from the main screen via the menu when Authenticated
- Export Entries (CSV) - Blood glucose readings and CGM data
- Export Treatments (CSV) - Insulin, carbs, and other treatment data
- Export Profile (CSV) - Profile settings and configurations
- Export DeviceStatus (CSV) - Device and uploader status information

Modified backend API endpoints to properly handle CSV formatting for all data types


AndyLow91:
Thanks for putting this together. The export feature is useful, but I do not think this is merge-ready yet.

The main blocker is that the CSV output is not really a complete database export. Several formatters emit only selected fields, and the profile export in particular drops the `store` object, which contains the actual basal, sensitivity, carb ratio, and target schedules. For individual/research export use, that would produce incomplete or misleading data unless the reduced schema is explicitly intentional and documented.

There is also a data correctness issue where numeric zero values can be lost. Several fields use `value || ''`, so valid values like `0` for carbs, insulin, protein, fat, etc. become blank cells in the CSV.

A couple of cleanup items also need addressing:
- `static/report/js/exports.js` binds to report export buttons that are no longer present in `views/reportindex.html`, so that code is currently dead.
- The diff has trailing whitespace in several changed files.
- There are no tests covering the new CSV behavior.

I would suggest tightening this up by defining the intended CSV schema for each collection, preserving zero values, removing or completing the report-page export code, and adding tests for entries, treatments, profile, and devicestatus CSV output before merging.

PR #8405, opened 2025-11-18
Fix timezone display to show device timezone instead of browser timezone

### LINKED ISSUE #8403: Main screen clock should use device reported timezone
**Describe the bug**
Viewing nightscout in a different timezone the main clock is displayed in the local time for the browser. 
It should show the the time for the reporting device (Loop/AAPS/TRIO Phone)
This seems to apply for all timestamps shown on main page

**To Reproduce**
1. Go to Vancouver
2. Have your looper in Sweden
3. Open nightscout on computer in Vancouver
4. Look at main clock 
5. Time is in local time for Vancouver

**Your setup information**
- ns version 15.0.4
- Loop, Dexcom ONE+

ryceg:
Fixes #8403

## Changes
- Clock and main page now display device timezone from profile instead of browser local time
- When device and browser timezones differ (different UTC offsets), both times are shown
- When timezones match, only one time is displayed
- Created shared timezone utility module for code reuse

## Technical Details
- Fetches timezone from profile API
- Uses moment-timezone to handle timezone conversions
- Graceful fallback to browser time if profile unavailable

### DIFF
--- a/lib/client/clock-client.js
+++ b/lib/client/clock-client.js
@@ -1,8 +1,11 @@
 'use strict';
 
 var browserSettings = require('./browser-settings');
+var moment = require('moment-timezone');
+var timeformat = require('./timeformat');
 var client = {};
 var latestProperties = {};
+var profileTimezone = null;
 
 client.query = function query () {
   var parts = (location.search || '?').substring(1).split('&');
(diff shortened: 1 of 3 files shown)


AndyLow91:
Nice work @ryceg, and thanks for tackling this.

I think the intended behavior is right: the main UI should show the device/profile timezone rather than assuming the browser timezone, and showing both device and local time when they differ makes sense.

I don’t think we should merge this implementation as-is, though. The main issue is that the timezone-aware moment().tz(...) value is converted back to a plain JS Date, and then formatted with browser-local APIs like getHours() / d3.timeFormat(). Since Date does not retain the target timezone, the displayed “Device” time can still end up as the browser-local time.

There’s also a profile handling gap: the clock client fetches /api/v1/profile.json directly and reads store.Default.timezone, which can miss profiles where defaultProfile is not Default, and it bypasses the timezone normalization now used elsewhere for values like GMT+4 / GMT+5:30.

I think the best next step is to keep the product behavior from this PR, but rework the implementation so the timezone-aware value is formatted before converting to display text, and add tests for browser/device timezone mismatch plus fixed-offset profile timezone strings.

PR #8413, opened 2025-12-30
New files from Fly.io Launch

ceciliavinhas:


AndyLow91:
Thanks for the PR, but I don’t think this should be merged as submitted.

The `fly.toml` in this PR is generated for a specific Fly app and region, not a generic project-level configuration. In particular, it hardcodes an app name and primary region that appear to come directly from a personal `fly launch` run.

That makes this unsuitable for the main repo in its current form. If we want to support Fly.io officially, the better approach would be either:
- documentation for how to deploy Nightscout on Fly.io, or
- a truly generic Fly configuration plus docs, without any user-specific app name or region baked in.

If you want to revisit this, please open a reduced PR that makes the Fly setup reusable for other users.

PR #8419, opened 2026-01-15
Add tests for iOS loop push notifications and websockets

je-l:
Adds high-level integration tests related to iOS loop push notifications and websocket API. This should help with e.g. dependency updates and to avoid introducing regression bugs when making changes.

Statement coverage increased from 63.8% to 65.4%.
Branch coverage from 51.0% to 53.0%.


AndyLow91:
Thanks for the PR. The additional coverage looks useful, but I think this needs revision before merge.

A few issues should be addressed first:

- `loopnotifications.test.js` appears order-dependent around `API_SECRET`. The suite creates the app using `process.env.API_SECRET`, but authenticates using a hard-coded hash for a different secret. That makes the test fragile depending on prior test state.
- Both new suites start real Nightscout server instances but do not fully clean them up. Please add proper teardown for the server/bus and any other test resources.
- The fake APNs server uses a fixed port, which makes the test more brittle and can cause avoidable conflicts.

If you can make the new tests fully self-contained and deterministic, I’d be happy to take another look.

je-l:
Pushed fixes to all three comments, the teardown looks like this now:

```js
after(function(done) {
  inst.server.close();
  inst.ctx.bus.teardown();

  delete process.env.LOOP_APNS_KEY;
  delete process.env.LOOP_APNS_KEY_ID;
  delete process.env.LOOP_DEVELOPER_TEAM_ID;
  delete process.env.ENABLE;

  apn.Provider = OriginalApnProviderClass;

  guardedDrop(inst.ctx.profile(), err => {
    if (err) return done(err)

    fakeAPNServer.close(done);
  });
});
```

I removed websocket.test.js, the overlap is high with websocket.shape-handling.test.js.
Interestingly, the old websocket tests didn't pass. I think it shouldn't be a regression bug, but I will look a bit more later and create a new issue if needed. For me this PR is ready to merge.

je-l:
Updated with dev branch, I merged loop.server.test.js file with loopnotifications.test.js as they are testing the same feature. I think including both is still reasonable since the other uses fake API server and other mocks the library functions. The branch coverage difference is now only 48.29% vs. 48.53% though.

Also see this change:

```diff
commit b7323fc884264cccab967cd0dff4e6cbb1cc64c2
Author: je-l <je-l@users.noreply.github.com>
Date:   Thu Sep 3 20:10:04 2026 +0300

    Instrument all files during nyc runs

    See https://github.com/istanbuljs/nyc#common-configuration-options

    By default nyc doesn't instrument all files, and this caused discrepancy
    in the coverage results between branches, since the commonjs imports can
    change. It should be reasonable to always count every lib/ js file.

diff --git a/.nycrc.json b/.nycrc.json
new file mode 100644
index 00000000..e62acb26
--- /dev/null
+++ b/.nycrc.json
@@ -0,0 +1,6 @@
+{
+  "all": true,
+  "include": [
+    "lib/**/*.js"
+  ]
+}
```

I noticed the coverage statistic denominators jumped between branches. After the config change the branch coverage dropped from 55.32% to 48.53%. I think it's more useful like this.

Issue #8446, opened 2026-03-06
Fixes for Mongodb8


zehnBE:
**Describe the bug**
MongoDB 8 needs other 

**To Reproduce**
Use mongodb8.x

**Your setup information**
dev 15.0.7
master 15.0.6
# MongoDB 8.x compatibility fixes + WebSocket broadcast improvements

## Overview

MongoDB 8.x introduced stricter type validation and removed several deprecated
APIs that Nightscout relies on. This causes a range of runtime errors on any
Nightscout instance running against a MongoDB 8.x backend. This PR addresses
all known incompatibilities and includes two additional WebSocket correctness
fixes discovered during the investigation.

---

## Fix 1 — `lib/api3/storage/mongoCollection/find.js`: strict integers for limit/skip

**Error:**
```
MongoInvalidArgumentError: limit requires an integer
```

MongoDB 8.x enforces strict integer types for `.limit()` and `.skip()`. When
these values are passed as strings or floats (which can happen depending on how
query parameters are parsed), MongoDB now rejects the call outright instead of
coercing silently as it did in earlier versions.

**Fix:** Added a `toSafeInt(value, defaultValue)` helper that normalises the
value to a proper integer before passing it to the cursor chain.

---

## Fix 2 — `lib/server/websocket.js`: safe ObjectID construction

**Error:**
```
TypeError: ObjectID is not a constructor
```

MongoDB 8.x removed `ObjectID` as a named export from the `mongodb` Node.js
driver (it was already deprecated in favour of `ObjectId` with a lowercase `d`).
Any call to `new ObjectID(id)` in the legacy WebSocket handler therefore throws,
causing `dbUpdate` and `dbRemove` operations sent by older NSClient versions to
fail silently or operate on the wrong document.

**Fix:** Added a `safeObjectID(id)` helper that validates the 24-character hex
format before constructing the object, and falls back to using the raw string ID
if the format does not match. Applied to all 4 call sites in `websocket.js`.

---

## Fix 3 — `lib/server/treatments.js`: atomic operators required for update()

**Error:**
```
MongoInvalidArgumentError: Update document requires atomic operators
```

MongoDB 8.x no longer accepts plain replacement documents in `collection.update()`
— atomic operators such as `$set` are now mandatory. The legacy treatments upsert
code passed the raw document object directly, which worked in MongoDB 4.x/5.x
through implicit behaviour but is now explicitly rejected.

[...]

PR #8447, opened 2026-03-06
Merges Patched from both #8421 and #8446

AndyLow91:
Merges Patched from both #8421 and @8446 and merged recent updates from v15.0.6


bewest:
Nice! Thanks! :balloon:

tests/query.test.js
```diff
@@ -35,4 +35,18 @@ describe('query', function ( ) {
 
     (typeof opts.date).should.equal('undefined')
   });
+
+  it('should keep non-ObjectId _id queries as strings', function ( ) {
```
bewest, line 39: Is this the correct requirement, @copilot?

PR #8469, opened 2026-04-14
fix: normalize non-IANA timezone strings in profile getTimezone()

### LINKED ISSUE #8461: Reports misaligned when using non-named timezones
### Summary

Nightscout reports can display data under the wrong calendar day when the site timezone is set to a non-named timezone (e.g. GMT+4) instead of a standard IANA timezone (e.g. America/Toronto).

This results in report data appearing offset by one day.

### Expected behaviour

Nightscout should correctly handle both:

Named IANA timezones (e.g. America/Toronto)
Non-named offset-based timezones (e.g. GMT+4, UTC+2)

Report generation should always align data to the correct calendar day regardless of timezone format.

### Actual behaviour

When a non-named timezone (e.g. GMT+4) is set:

Reports become misaligned
Data appears under the wrong day (typically shifted forward by one day)
This persists until the timezone is manually corrected to a named timezone
Real-world example

User located in Toronto (expected timezone: America/Toronto)

### Root cause:

External uploader (e.g. Loop) re-uploaded timezone as GMT+4
Nightscout accepted this value but did not correctly handle it in report generation
Steps to reproduce

Set Nightscout site timezone to a non-named value such as:

GMT+4

or

UTC+X
Ensure data exists across multiple days
Open any report view (e.g. Daily / AGP / stats)
Observe:
Data appears under incorrect calendar days
Typically shifted forward or backward depending on offset
Additional context
This can happen when external systems (e.g. Loop or other uploaders) send timezone values as offsets rather than named zones
Nightscout currently accepts these values but does not normalise or interpret them consistently in reporting logic
The issue is intermittent for users because the timezone may switch back and forth depending on uploader behaviour
Suggested fix / direction

### One or more of the following:

Normalise timezone input
Convert offset-based timezones (GMT+4) into a proper IANA timezone where possible
Add proper offset handling
Ensure report generation logic correctly interprets offset-based timezones
Validation / warning
Warn or reject non-named timezones in favour of IANA timezones
Fallback behaviour
If a non-named timezone is used, explicitly apply the offset during report calculations
Impact
Incorrect reports
Misleading data interpretation
Loss of trust in reporting accuracy
Workaround

Manually set timezone to a named IANA timezone (e.g. America/Toronto) in Site Settings.

### Notes

This issue is not specific to any single uploader (e.g. Loop), but occurs whenever a non-named timezone is used.

quarktwain:
## Problem

Loop and Trio uploaders sometimes set the profile timezone to non-IANA formats like `GMT+4` or `UTC+2` instead of proper IANA zone names (e.g. `America/Toronto`).

`moment-timezone` silently falls back to UTC for unrecognized strings, causing `startOf('day')` to use UTC midnight instead of local midnight. Reports then show data under the wrong calendar day.

## Root cause

`profile.getTimezone()` in `profilefunctions.js` returned the raw timezone string without normalization. Also had a pre-existing no-op bug: `rVal.replace('ETC','Etc')` wasn't assigning the result.

## Changes

- Normalize `GMT+N`/`UTC+N` to `Etc/GMT-N` (IANA sign inversion) in `getTimezone()`
- Fix the `ETC` → `Etc` replace to actually assign the result
- Add 11 tests covering normalization, sign inversion, edge cases (GMT+0, GMT-0, GMT+12, half-hour offsets, bare UTC/GMT, IANA passthrough)

| Input | Output | Status |
|-------|--------|--------|
| `GMT+4` | `Etc/GMT-4` | Fixed |
| `UTC+2` | `Etc/GMT-2` | Fixed |
| `GMT-5` | `Etc/GMT+5` | Fixed |
| `ETC/GMT+5` | `Etc/GMT+5` | Fixed |
| `GMT+5:30` | `GMT+5:30` | Unchanged (Etc/GMT doesn't support sub-hour offsets) |
| `America/Toronto` | `America/Toronto` | Unchanged |
| `UTC` | `UTC` | Unchanged |

## Test plan
- [x] All 35 profile tests pass (`npm test`)
- [x] Existing tests unaffected
- [x] Verified with moment-timezone that normalized strings produce correct UTC offsets

Closes #8461

### DIFF
--- a/lib/plugins/pump.js
+++ b/lib/plugins/pump.js
@@ -313,7 +313,7 @@ function init (ctx) {
 
   function prepareData (prop, prefs, sbx) {
     var pump = (prop && prop.pump) || { };
-    var time = (sbx.data.profile && sbx.data.profile.getTimezone()) ? moment(sbx.time).tz(sbx.data.profile.getTimezone()) : moment(sbx.time);
+    var time = sbx.data.profile ? sbx.data.profile.applyTimezone(moment(sbx.time)) : moment(sbx.time);
     var now = time.hours() + time.minutes() / 60.0 + time.seconds() / 3600.0;
     var batteryWarn = !(prefs.warnBattQuietNight && (now < prefs.dayStart || now > prefs.dayEnd));
     var result = {
--- a/lib/profilefunctions.js
+++ b/lib/profilefunctions.js
@@ -109,7 +109,7 @@ var moment = ctx.moment;
     // This WILL break on the server; added warnings elsewhere that this is missing
     // TODO: Better warnings to user for missing configuration
 
-    var t = profile.getTimezone(spec_profile) ? moment(minuteTime).tz(profile.getTimezone(spec_profile)) : moment(minuteTime);
+    var t = profile.applyTimezone(moment(minuteTime), spec_profile);
 
     // Convert to seconds from midnight
     var mmtMidnight = t.clone().startOf('day');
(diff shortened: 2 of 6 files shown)


zauberflote620:
@bewest @AndyLow91 - hey, first-time contributor here, CI hasn't auto-run yet (workflow-approval gate). Could one of you kick it off when you have a minute? PR closes #8461 (non-IANA timezone normalization in getTimezone()). Happy to address feedback. Thanks!

AndyLow91:
@zauberflote620 - thanks for picking this up! I've approved the CI test workflow now

AndyLow91:
Thanks for digging into this. The `ETC` fix and the whole-hour `GMT/UTC` normalization look good, but I think there’s still one correctness gap here: sub-hour offsets.

Right now the new test expects `GMT+5:30` to stay unchanged, but `moment-timezone` does not recognize `GMT+5:30` either, so those users still fall back to UTC and still get the wrong local day boundaries in reports. I verified that `moment.tz(..., 'GMT+5:30')` warns and behaves as UTC.

I think the safer fallback is to treat numeric offsets as fixed offsets instead of trying to force them into `Etc/GMT` names. In practice that means normalizing `GMT+5:30` / `UTC+5:30` to `+05:30` and then handling that separately from IANA zones at the `moment` call sites. `Etc/GMT` works for whole hours only, but fixed offsets would cover the half-hour / quarter-hour cases too.

As written, this PR fixes `GMT+4`-style inputs, but it still leaves `GMT+5:30`-style inputs broken while the new test suite now codifies that broken behavior.

quarktwain:
Pushed v2 in 79e0155d. getTimezone() now normalizes sub-hour offsets to fixed-offset strings (+05:30, +05:45, -03:30, etc.). Added two helpers, applyTimezone(mom) and parseInTimezone(str), that detect the fixed shape and apply via utcOffset / parseZone instead of moment.tz(). 8 call sites updated to use them. Tests updated, including the GMT+5:30 assertion that now expects +05:30 and a round-trip check that moment formatted back is +05:30.

Could you kick the CI workflow approval again when you have a sec? Thanks.

AndyLow91:
nice work @zauberflote620!

I’m happy to merge this as a solid step forward. The normalization in getTimezone(), the ETC -> Etc fix, and the new helper direction all look good, and CI is green.

I do think there’s still one follow-up gap: a few report paths still call moment.tz(..., zone) directly after getTimezone() can now return fixed offsets like +05:30. moment-timezone does not recognize those fixed-offset strings, so those remaining paths may still fall back to UTC for sub-hour offsets.

I’m going to merge this into the `wip/next-release` branch, and then handle the remaining report call sites in a follow-up patch so we can finish closing out #8461 cleanly and then get this into `dev`

Issue #8219, opened 2024-01-30
Date translation problem


stucka:
**Describe the bug**
Within reports/daily stats tonight:
At night I seem to be getting a day-of-the-week response based in UTC or where the server is hosted, but I'm getting a calendar-day response from my local time.

I see for example "Tuesday, 1/29/2024" -- Tuesday belonging to the latest date in the server location, and 1/29 belonging to my location. 1/29/2024 is a Monday.

**To Reproduce**
I suspect this problem will look different in the morning when the dates appear to be the same.

**Expected behavior**
There is a mismatch in day-of-week reporting and calendar-date reporting when the remote date is different from the local date.

**Screenshots**
Will attach from my phone in a comment.

**Your setup information**
Nightscout 15.0.2
Sharebridge

**Additional context**
I believe the problem is in `/lib/report_plugins/utils.js`, with some unnecessarily complex day-of-week hardcoded that will also break translations. The handling for the day-of-the-week is different than the handling for the localized date, but both can be handled in the same way.

However, I am getting readings for the actual Tuesday, 1/30/2024, here in my living room where it's Monday, 1/29/2024. Any fix would need some testing.

That said, however, if a localized date string is what makes sense here, then the answer might be to rip out the hard-coded day-of-week lookup array and let `toLocaleDateString` handle the full translation.

(for testing: var date = new Date()    )

`ret = date.toDate().toLocaleDateString({weekday: "long", month: "numeric", day: "numeric", year: "numeric"});`

Sample runs:
date.toLocaleDateString("en-US", {weekday: "long", month: "numeric", day: "numeric", year: "numeric"})
"Monday, 1/29/2024"
date.toLocaleDateString("de-DE", {weekday: "long", month: "numeric", day: "numeric", year: "numeric"})
"Montag, 29.1.2024" 

I had a little bit more information here: https://github.com/nightscout/cgm-remote-monitor/issues/8176#issuecomment-1915891160

stucka:
![Screenshot_20240129-212723](https://github.com/nightscout/cgm-remote-monitor/assets/533487/3b35c2ba-e71d-4f4a-860c-bb0e09fbf0a8)

stucka:
The next morning's is not better, on Chrome and Firefox for Android.
![Screenshot_20240130-044022](https://github.com/nightscout/cgm-remote-monitor/assets/533487/0cc91a6d-b55c-4ecb-8662-0642722258a4)

stucka:
[...]

Issue #8176, opened 2023-11-20
Time Change, Date and Treatment Discrepancy


marionbarker:
**Describe the bug**
I am noting discrepancies in date vs some treatments (and thus plots) before and during the recent time change. I believe it is basal and events that don't align as expected.

**Comment**
I'm happy to close this issue and add to an existing issue if that is preferred - just let me know which issue number. I believe this is related to a long-standing problem when the computer time is not aligned with the profile time. I believe this was partially fixed with release 15.0.

**Configuration**
* Nightscout 15.0.2
* Uploader: Loop, dev branch (commit 6b7bba6) on an iPhone 11 pro, iOS 17.1.1

**Details and Graphics**
The graphics are way zoomed out to show multiple days, so are poor resolution - just look at the overview. I describe the steps to generate each graph and then follow up with composite graphics to emphasize the discrepancy.

For all reports, I started fresh (close report, open report and configure) so that prior fetches from the database are cleared.

For the US, the time change was at 2 am on Nov 5, 2023 (11/5/2023 using US notation.)

First test, all days are before time change:
* Open report, select day to day from 11/1 to 11/4
    * first time hit of submit, plots labeled 10/31 to 11/4 are displayed
        * only first day (10/31) shows basal plot and "default" at midnight
        * subsequent days (11/1 through 11/4) show no basal plot and "default" at 23:00
    * second time hit submit, only the requested days are plotted (11/1 through 11/4), still with no basal plot and "default" at 23:00
    * extract just the plots for Nov 2, 3 and 4 and add to composite below (right side, top)

Second test, select days that span the time change:
* Open report, select day to day from 11/3 to 11/6
    * get same result every time you hit submit, plots labeled 11/2 to 11/6 are displayed
    * all show basal in plots, all show "default" at midnight
    * extract the full display and add to composite below (left side)

Third test, select days after the time change:
* Open report, select day to day from 11/6 to 11/8 (all after time change)
    * get same result every time you hit submit, plots labeled 11/6 to 11/8 are displayed
    * all show basal in plots, all show "default" at midnight
    * CGM Sensor Start of 11/7 shows up on the right date
    * extract just the plots for Nov 6 and 7 and add to composite below (right side, bottom)

First composite graphic - plots on left and right are for same date from different selections for date range for Day to Day plots.
* Please note the upper right plot does not show basal and has Default showing at 23:00

[...]

Issue #6853, opened 2021-02-08
Wrong dates in day to day report


baswieb:
If I generate a day to day report in Nightscout (with date range "last two weeks" and sort order "newest on top") the first day shown is "monday 8-2-2021". The second day shown is "sunday 6-2-2021". This is wrong because the date of last sunday is 7-2-2021.

Nightscout version 14.1.0
Data is uploaded with Loop.
Timezone is "Europe/Amsterdam"
Everytime I generate a report I have to (re)set the timezone manually to "Europe/Amsterdam" in the profile editor.
When loop has uploaded the profile the Timezone in the profile becomes empty.

baswieb:
![image](https://user-images.githubusercontent.com/8107422/107285825-02b8c400-6a60-11eb-9522-8a4fff0bf4b1.png)

sulkaharo:
Right, so this sounds like a Loop issue, where it's uploading an incorrect profile. To get that fixed, you'd need to get in touch with the Loop developers.

baswieb:
The timezone becoming empty can be a loop issue but the report showing "sunday 6-2-2021" when it should show "sunday 7-2-2021". Is this a bug in the report?

Nightfoxy:
@sulkaharo  Is this the same issue?
When I run a day to day report for my daughter, the dates are correct.
When I run the day to day on someone else's site who is in a different time zone, the dates are off by 1.
![image](https://user-images.githubusercontent.com/46848705/120916130-9ee48780-c65c-11eb-9a39-113b58953829.png)

All involved are using Loop.

AndyLow91:
Closing this as resolved.

The Day to Day report timezone/date handling has been fixed in `dev` through #8469 and #8485. Reports should now use the profile/report timezone consistently, which addresses the off-by-one-day/date-label problems described here.

If this is still reproducible on current `dev`, please comment with the profile timezone, selected report range, browser/server timezone if known, and a screenshot of the incorrect date labels.

Issue #5659, opened 2020-05-14
Day to Day report in Nightscout has erroneous time scale for first day of report


bcdiabetes:
The Day to Day report gives an erroneous timescale for the first day.  The other days are correct.

The screenshot below shows the situation.  It was taken at 5:10 pm - the tracing suggests that hit was almost 12:00 AM (midnight).

![image](https://user-images.githubusercontent.com/58049669/81984865-4b833f80-95ea-11ea-88b4-1703b253e1f5.png)

To reproduce the problem, in the main Nightscout window choose Reports.  In the new window choose Day to Day and click "show".

It should show a correct time scale such that the time of the last reading in the primary Nightscout window corresponds to the time shown on the scale in the Report view for the first day.

Setup Nightscount 13.0.1
Dexcom G6 feed.  Reproduced on Chrome & Firefox on my Linux machine running Ubuntu 20.04 and my brand-new Windows 10 machine running Chrome, as well as on my Android phone (V10) running Chrome.

My friend using an iPhone and Mac found no problem unless she chose "logarithmic" scale instead of "linear" scale.  Choosing Linear for me did not fix the bug.

ryceg:
This will be fixed by #8348 thanks to bniels' identifying it in 917f384

AndyLow91:
Closing this as resolved.

The Day to Day report timezone/date handling has been fixed in `dev` via #8469 and #8485. The report should now build day boundaries and labels using the intended profile/report timezone, which addresses the incorrect first-day time scale described here.

If this still happens on current `dev`, please comment with the profile timezone, selected report range, browser/server timezone if known, and a screenshot of the incorrect first-day scale.

PR #8493, opened 2026-05-02
Add tests for weektoweek report date displays.

bniels707:
With fixes for incorrect dates now landed in #8469 and #8485, add tests to verify correct date displays for the weektoweek report type.

These were written to verify the fixes from my [date fixes branch](https://github.com/bniels707/cgm-remote-monitor/tree/report_date_display_fixes) that I was siting on waiting for test suite fixes to land, now might as well merge the tests on their own to prove the related issues are fixed.

Longer term I'm hoping to add similar tests for all report types (and fix the issues with repeated API requests, and the "some html" test being slow).

### DIFF
--- a/tests/reportplugins.weektoweek.test.js
+++ b/tests/reportplugins.weektoweek.test.js
@@ -1,0 +1,592 @@
+'use strict';
+
+require('should');
+var _ = require('lodash');
+var benv = require('benv');
+var serverSettings = require('./fixtures/default-server-settings');
+
+var nowData = {
+  sgvs: [
+    { mgdl: 100, mills: Date.now(), direction: 'Flat', type: 'sgv' }
+  ]
+  , treatments: []
+};
+
+function buildProfile (timezone) {
+  return [
+    {
+    //General values
+    'dia':3,
+
+    // Simple style values, 'from' are in minutes from midnight
+    'carbratio': [
+      {
+        'time': '00:00',
+        'value': 30
+      }],
+    'carbs_hr':30,
+    'delay': 20,
+    'sens': [
+      {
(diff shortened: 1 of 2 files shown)


AndyLow91:
Thanks for adding this. I agree this is useful coverage now that the week-to-week date fixes from #8469 and #8485 have landed.

I pulled the branch locally and ran the focused test file:

`tests/reportplugins.weektoweek.test.js`

At the moment it fails in all 5 new cases because the report is rendering dates using the localized Moment `L` format, which gives zero-padded dates like:

`Tuesday 09/01/2015-Monday 09/07/2015`

but the test currently expects:

`Tuesday 9/1/2015-Monday 9/7/2015`

So I think the PR is conceptually good, but the expected strings need to be updated before this can merge cleanly. Once those assertions match the current report output, this should be a helpful regression test for the timezone/date handling.

bniels707:
Good catch on the zero padding. I pushed updates with zero padding expected.

Tests still fail though. The entries API request range is built in terms of the profile timezone, not the browser timezone. I _think_  #8176 and #6853 are fixed as timezones are no longer applied repeatedly. I'm not sure about #5659, I _think_ it is fixed as absolute size of the time range is correct since the rollover at beginning and ends of time ranges is appropriately handled by the single timezone application. #8219 might have more subtlety, it isn't just that the translation is wrong, it's that there was an expectation that the report would be generated in "local" time, not remote time.

If making the data request in profile time is the desired behavior, I can change the tests. If, instead, we think reports should be generated in browser / "local" time, changes  to the report client are needed.

Either way, I should explore adding some data alignment checks in the tests.

PR #8501, opened 2026-05-05
fix(alarms): externalize snooze state to mongo so multi-instance ack is honored (#8194)

### LINKED ISSUE #8194: CGM Alarms Can Not Be Snoozed
**Describe the bug**
Selecting a snooze interval will silence the alarm for [much] less than a minute, after which the alarm will sound again. CGM alarms cannot be snoozed.
 
**To Reproduce**
When BG level alarm is triggered, select a snooze interval. The alarm will be silenced but will sound again in less than a minute.

**Expected behavior**
The alarm should be silent until the snooze interval expires.

**Screenshots**

**Your setup information**
Nightscout 15.0.0.2 My experience has been with my site and five subscriber sites. This behavior was not observed or reported with versions <=14.2.6

My reproduction was with macOS 14.1.1 - Chrome Version 120.0.6099.109 (Official Build) (arm64)

I'll gather this data for subscriber reports of this going forward. I don't think CGM or Uploader information is required, but let me know if it is.

quarktwain:
Fixes #8194.

## Problem

`var alarms = {}` in `lib/notifications.js` is process-local. On multi-instance deploys (Cloud Run, k8s, Heroku scale-out, any reverse-proxy without sticky sessions), an ack on instance A is invisible to instance B, which re-emits the alarm well before the user-selected silence window.

NightMax70 documented this on the issue (2026-03-31): forcing Cloud Run to a single instance silences the bug. That workaround defeats horizontal scaling. This PR fixes the root cause by externalizing snooze state to MongoDB.

## Approach

New `alarmsnooze` MongoDB collection, keyed by `level-group`, holding `lastAckTime`, `silenceTime`, `expiresAt`. TTL index on `expiresAt` auto-cleans expired docs.

- `notifications.ack` -> `alarmStorage.setSnooze` (aggregation-pipeline upsert with conditional `$set` keeping the later `expiresAt`, so a shorter snooze on instance B cannot clobber a longer one from instance A).
- `getAlarm` -> async `getSnooze` on first creation per (level, group) per process. First emission for that pair is deferred until the storage call resolves or `ALARM_REFRESH_TIMEOUT_MS` (default 5000) elapses. INFO and lower are not snoozeable, so they are not deferred.
- `bootevent.js` binds `ctx.alarmStorage` early in `setupInternals` (before `ctx.notifications` is constructed), then kicks `ensureIndexes()` async during the existing index step (does not block boot).

## Pre-empting predictable questions

- **Aggregation pipeline vs find-then-update**: atomic single round-trip; prevents TOCTOU between racing instances. Requires Mongo 4.2+, already the project floor.
- **Why 5s timeout**: configurable via `ALARM_REFRESH_TIMEOUT_MS` (documented in README). Atlas free-tier RTT is typically 200-500ms; 5s covers cold-start. On timeout, emit proceeds from in-memory state (matches current behavior on Mongo-down).
- **Why a new collection**: alarms are bus events today, not persisted; new collection is the minimal surface.
- **Self-hosted without Mongo**: `getCollection()` returns `null`, all calls no-op, current in-memory behavior preserved. No regression for single-instance deploys.
- **Boot perf**: `ensureIndexes()` is async fire-and-forget (`.catch` only). Boot does not await.
- **URGENT cascade**: `notifications.ack(2, group, time)` recursively calls `notifications.ack(1, group, time)`, which writes both `1-group` and `2-group` rows to storage. Test asserts both rows are persisted.

## Operational notes

- New env var: `ALARM_REFRESH_TIMEOUT_MS` (default 5000) — see README.
- New collection `alarmsnooze` is created on first ack after deploy.
- TTL index is created best-effort; failure is logged but does not block boot.

## Known limitations / not in this PR

- **Rolling-deploy mixed-fleet**: during a blue-green window where some pods run old code and some new, ack honoring is asymmetric (old pods don't read storage). User-visible delta over the typical 2-5 minute window: at most one duplicate alarm if ack hits an old pod and the next BG cycle hits a new pod. This matches current production behavior; the PR does not regress it.
- **URGENT+WARN cascade is two non-atomic writes**: if a process crashes between the URGENT `setSnooze` and the WARN cascade, instance B sees URGENT snoozed but WARN not. Mid-cascade crash window is small in practice; flagging for visibility.
- **Companion bug in `lib/server/pushnotify.js`**: the same pattern applies to `receipts` and `recentlySent` NodeCache state — Pushover acks behind a load balancer can silently drop, and duplicate Pushover/Maker pages can ship from each instance. This PR does not address that subsystem (different storage shape needed). Will file as a separate companion issue if desired.
- **Tests use a JS reimplementation of the conditional pipeline**: a typo in the production aggregation operators (e.g., `$ifnull` lowercase) would not be caught by the unit suite. CI runs against real MongoDB which exercises the path via the `notifications-api` test boot. Adding an in-memory MongoDB round-trip test for the storage layer is doable in a follow-up if you'd like.

## Files

- `lib/storage/alarmStorage.js` (new) — Mongo CRUD with null-collection fallback
- `lib/notifications.js` — ack writes through to storage; getAlarm gates first emit on storage refresh
- `lib/server/bootevent.js` — early `ctx.alarmStorage` binding + async TTL index creation
- `tests/alarmStorage.test.js` (new) — unit coverage of getSnooze/setSnooze/ensureIndexes including conditional-upsert pipeline shape
- `tests/notifications.test.js` — extends existing suite with refresh-gate, timeout fallback, INFO-not-gated, undefined-group normalization
- `tests/notifications.multi-instance.test.js` (new) — integration test verifying ack on instance A is honored by instance B, conditional-upsert preserves longer snooze, URGENT cascade persists both rows
- `docs/data-schemas/alarmsnooze-schema.md` (new) — collection schema doc, indexed in `docs/INDEX.md`
- `README.md` — `ALARM_REFRESH_TIMEOUT_MS` env var documented


AndyLow91:
Thanks for tackling this. I think the overall direction is right: the root cause in #8194 is that snooze state is process-local, so externalizing it to MongoDB is the correct fix for multi-instance deployments.

I pulled the branch and ran the focused tests against the merge ref:

`alarmStorage.test.js`, `notifications.test.js`, `notifications.multi-instance.test.js`: 29 passing  
`notifications-api.test.js`: 1 passing

That said, I don’t think this is merge-ready yet. I see two behavioral gaps in the refresh gate:

1. On first WARN/URGENT creation, `emitNotification()` returns while `alarm.refreshPending` is true. When `getSnooze()` later resolves with no active snooze, the callback only clears `refreshPending`; it does not retry or resume the notification that was just suppressed. In production that means the first alarm on a fresh process is delayed until the next data/plugin processing cycle, and if the condition clears before then, it may never emit.

2. If the refresh timeout fires first, `emitNotification()` clears `refreshPending` and emits from memory. If Mongo then returns an active snooze from another instance, the callback ignores it because `!alarm.refreshPending`, so that process can keep using stale in-memory state for the lifetime of that alarm object.

I think the fix should make the refresh path either await before deciding to emit, or store/retry the pending notification once the storage read resolves. It should also decide how late-but-valid storage results update the in-memory alarm after timeout.

Small hygiene issue too: `git diff --check` fails on trailing whitespace in `docs/data-schemas/alarmsnooze-schema.md`.

So conceptually this addresses #8194, but I’d request changes before merging.

quarktwain:
@AndyLow91 apologies for the long silence; this was fixed the day after your review in 7f2182c3, I just dropped the ball on pinging you. Both refresh-gate gaps are addressed there, and I've now brought the branch up to date with current dev.

1. Deferred notify is now stored on `alarm.pendingNotify` and retried from the `getSnooze().then()` callback via a shared `resolveRefresh()` helper. The first alarm on a fresh process emits as soon as storage resolves, instead of waiting for the next cycle (or being lost if the condition clears first).

2. The `if (!alarm.refreshPending) return;` early-return is gone. Late-but-valid storage results now update `lastAckTime`/`silenceTime` whenever the doc is newer, even after the deadline fired and we emitted optimistically from in-memory state.

I also hardened the catch path (fail-open with stale state rather than silent drop, documented inline since this is a medical-safety path) and added tests for the retry and `.catch()` rejection paths. Trailing whitespace in `docs/data-schemas/alarmsnooze-schema.md` is fixed too.

The branch was well behind dev, so I merged current dev in (clean, no conflicts) and re-ran the focused alarm/notification suite locally against the merge. alarmStorage, notifications, notifications.multi-instance, and notifications-api all green (31 tests). CI is now green across the full matrix, too.

Ready for another look when you have a window.

IsraelAfangideh:
Thanks for this — I went deep on it because the multi-instance root cause is exactly right and worth landing. I pulled the branch, and the cold-instance path works as advertised. But I think there's a residual gap on the **warm-instance** path that's worth closing before merge, because it's arguably the scenario #8194 actually reports.

**The gap:** `getSnooze()` is read exactly once per `(level, group)` per process — only in `getAlarm()`'s `if (!alarm)` creation branch. The `alarms` map is never cleared in production (the only reset is `resetStateForTests`), so a long-lived instance that already holds an alarm object never re-reads storage. Concretely, with two warm instances both serving:

1. Instance B sees the alarm and creates its alarm object (storage empty) → emits.
2. User acks on instance A → `setSnooze` persists to Mongo, A's in-memory state updates.
3. Next CGM reading routes to **warm** B → B's cached alarm has `refreshPending === false`, so it skips the refresh and emits from stale `lastAckTime`/`silenceTime` → re-fires despite the ack.

That matches the "triggered again and again short after the silence period" reports, and it's why the single-instance workaround resolves it.

**Reproduction** — a warm-instance variant of your `notifications.multi-instance.test.js`. It's green for the cold case and red for the warm case on the current branch:

```js
// B creates its alarm object BEFORE the ack happens on A
feedReading(instanceB);            // warm B emits first alarm (no snooze yet)
await settle();
instanceA.ack(2, 'default', 60*60*1000);   // ack persisted to shared storage
await settle();
instanceB.__emits.length = 0;
feedReading(instanceB);            // new reading -> warm B
await settle();
emittedNotify(instanceB).should.equal(false);  // FAILS today: B re-emits
```

**Fix I've prototyped** (small, reuses your existing defer/`resolveRefresh` plumbing): extract the creation-time refresh block into a `maybeRefreshSnooze(alarm, level, group, force)` helper, and call it at the top of `emitNotification` with `force=false` so a warm alarm re-reads storage when its cached copy is older than a `REFRESH_STALE_MS` window (default 30s, env-tunable via `ALARM_REFRESH_STALE_MS`). That bounds storage load to ~1 read / 30s per active alarm while closing the warm-instance gap. The full notification + `alarmStorage` suites (incl. a Mongo-backed run) stay green with it, plus the warm test above goes green.

Happy to push this to your branch or open a PR against it — whichever you prefer. Either way, nice work getting the externalization in place.

quarktwain:
@IsraelAfangideh You're right, good catch! I traced it: `getSnooze()` only reads in the `if (!alarm)` creation branch, and the `alarms` map isn't cleared in production, so a warm instance that built its alarm object before a cross-instance ack never re-reads storage and keeps emitting from stale `lastAckTime`/`silenceTime`. My multi-instance test only exercises the cold path (B created after A's ack), so the warm case is genuinely uncovered, and it's closer to what #8194 reports than the case I covered.

The direction makes sense: reusing the existing refresh/`resolveRefresh` defer machinery for a staleness-bounded re-read in `emitNotification`. Two things I'll want to look at on the diff:
1. The staleness window is a real tradeoff on a medical alarm. For up to its duration a warm instance can still re-fire after an ack, so the re-read has to stay fail-open (like the existing `resolveRefresh` catch path) so a slow Mongo read never drops an alarm. Worth pinning the default and keeping it env-tunable, matching the existing `ALARM_REFRESH_TIMEOUT_MS` naming.
2. Please include the failing warm-path test in the PR so the fix can't silently regress later.

A PR targeting this branch (`wip/8194-alarm-snooze-externalize`) works. I'll review the diff and run it through the notification + `alarmStorage` suites (including the Mongo-backed path) plus your repro. Don't rebase on my account; I'll bring the branch current with dev myself. Appreciate you pulling the branch and writing the repro rather than just flagging it.

PR #8502, opened 2026-05-06
Remove Lodash dependency

AndyLow91:
## Summary

This carries forward the lodash removal work from #8360 by @ryceg, merged onto current `dev` via `wip/next-release`.

A new PR was needed because we ran into GitHub auth/permission issues when trying to push follow-up fixes directly to the original PR branch.

## Review Follow-Up

Reviewing #8360 revealed a few items that needed attention before opening this PR:

- The outstanding review note from sullkaharo was valid: the API3 search result callbacks had been converted into unclear `forEach` calls, making it look like date resolution/projection was not doing useful work.
- Resolving conflicts against current `dev` introduced more lodash-free paths that needed to be reconciled.
- The profile editor used browser `structuredClone`, but the repo still targets older Safari versions and Babel does not polyfill that API.
- The client plugin data refresh path had changed from lodash's shallow `_.clone(client.ddata)` behavior to a deep clone, which changed behavior and could add unnecessary cost on a hot path.

## Changes On Top Of #8360

- Brought `wip/next-release` up to current `dev` and merged in the lodash removal branch from #8360.
- Resolved merge conflicts while preserving current `dev` behavior and removing the remaining direct lodash usage.
- Clarified API3 search result processing so date resolution and field projection are explicitly applied.
- Replaced profile editor `structuredClone` usage with `client.utils.cloneDeep`.
- Restored shallow clone semantics for plugin sandbox data by using `Object.assign({}, client.ddata)`.
- Added regression coverage for the clone helpers and for the profile editor running without `structuredClone`.

## Validation

- `npx eslint tests/utils.test.js tests/profileeditor.test.js`
- Focused clone/profile tests: `8 passing`
- Changed `lib` files pass ESLint with warnings only
- `npm run bundle` passes with existing webpack size warnings
- `git diff --check origin/dev..origin/wip/next-release`
- `npm run test-ci` previously passed on the merge. After adding the regression tests, two reruns hit different existing-suite failures; both failing tests passed when isolated, and the new tests passed in the full runs.


AndyLow91:
Manual browser testing completed against this branch.

Tested:

- Profile Editor
  - Opened `/profile`
  - Added, cloned, renamed, and removed a profile
  - Added/removed basal, IC, ISF, and target BG rows
  - Saved, reloaded, and confirmed values persisted
  - Verified the profile editor works in browser after replacing `structuredClone`

- Main Dashboard
  - Loaded the main Nightscout page with realistic data
  - Confirmed chart rendering, plugin rendering, and pill updates
  - Toggled available views/plugins
  - No browser console errors observed

- Careportal / Treatments
  - Added a treatment
  - Edited the treatment
  - Deleted the treatment
  - Confirmed graph/API behavior updated as expected

- Admin Authorization UI
  - Created, edited, and deleted a role
  - Created, edited, and deleted a subject

- Reports
  - Generated day-to-day and week-to-week reports
  - Confirmed charts render and report/treatment data looks sane

PR #8506, opened 2026-05-07
Remove crypto browserify library

AndyLow91:
## Summary

Supersedes #8355. I had to open this replacement PR because GitHub would not allow me to push conflict-resolution commits back to the contributor fork, even though the original PR showed maintainer edits enabled.

The original change was previously reviewed and approved by @bewest on #8355, but it had since drifted from `dev` and needed conflict resolution. This branch resolves those conflicts while treating current `dev` as the priority.

## Changes

- Removes `crypto-browserify` from the browser bundle dependency path.
- Uses native browser `window.crypto.subtle.digest` for client-side API secret hashing when available.
- Preserves API-secret auth for `INSECURE_USE_HTTP` and other non-secure browser contexts with a dependency-free SHA-1 fallback.
- Keeps server-side crypto on Node's built-in crypto module via `node:crypto` where touched by the original PR.
- Preserves current `dev` changes from the conflict resolution and regenerates the current lockfile without `crypto-browserify`.
- Updates the auth test documentation to match the async `processSecret` flow and fallback coverage.

## Validation

- `npm ci` passed, including the production webpack build.
- `./node_modules/.bin/env-cmd -f ./tests/ci.test.env ./node_modules/.bin/mocha --timeout 5000 --require ./tests/hooks.js --exit ./tests/hashauth.test.js` passed.
- `./node_modules/.bin/eslint lib/client/hashauth.js tests/hashauth.test.js` passed.
- `npm run bundle` passed.
- `git diff --check origin/dev` passed.
- GitHub Actions test matrix passed on the updated PR branch.

Note: `npm run lint` still fails on existing unrelated lint errors in current `dev`.


AndyLow91:
Having reviewed this PR further, I agree with the goal of removing `crypto-browserify`, but I don’t think we can merge this as-is yet.

Nightscout still supports deployments that opt into HTTP via `INSECURE_USE_HTTP`, and some users may access their instance over plain HTTP on a LAN or self-hosted setup. This PR moves API-secret hashing in `lib/client/hashauth.js` to `window.crypto.subtle.digest`, but `Crypto.subtle` is only available in secure contexts in browsers. In those supported HTTP deployments, API-secret auth may fail before we can hash and verify the secret.

Could we preserve an HTTP-compatible fallback for the client-side SHA-1 hash path? That could be either a small SHA-1-only browser implementation/package, or a local fallback used only when `window.crypto?.subtle` is unavailable.

It would also be worth adding a regression test that disables/mocks out `window.crypto.subtle` and confirms `processSecret()` still produces the expected API-secret hash. That protects the `INSECURE_USE_HTTP` use case while still letting us remove the large `crypto-browserify` dependency chain.

AndyLow91:
Follow-up work added on top of the original conflict-resolution PR:

- `6c7c411` adds a dependency-free client-side SHA-1 fallback for API-secret hashing. The browser still uses `window.crypto.subtle.digest` when available, but Nightscout now continues to support `INSECURE_USE_HTTP` / non-secure contexts where `crypto.subtle` is unavailable. This also adds regression coverage for the fallback path.

- `01713bd` updates the authorization test documentation to match the now-async `processSecret()` flow and the new fallback coverage, and removes two unused imports from `tests/hashauth.test.js` so the focused lint check for the touched auth files is clean.

Validation after these commits:
- `hashauth.test.js` passes
- focused eslint for `lib/client/hashauth.js` and `tests/hashauth.test.js` passes
- `npm run bundle` passes
- GitHub Actions Node/Mongo matrix passes

PR #8515, opened 2026-05-10
Remove mongo44 support from README

PieterGit:
@bewest @AndyLow91 What's the ambition of MongoDB support in the upcoming Nightscout releases?
I think we should drop supporting MongoDB 4.4.x now @AndyLow91 added MongoDB 5.x and MongoDB 6.x support.

Since upgrading MongoDB should be done one major version at the time, I think it would be good to mimic that in the upcoming Nightscout major releases, e.g. 
image

I don't know what the status of the current Mongo Atlas support is. I don't use it.
Currently Mongo Atlas has only support for MongoDB 7.0 and higher.
MongoDB 6.0 in Atlas (released July 2022, EOL July 2025) 
See: https://www.mongodb.com/legal/support-policy/lifecycles#atlas

I can help fixing docs and doing testing, but unfortuately my programming skills are not adequate enough to regain full MongoDB and Node support in Nightscout.


PieterGit:
wrong branch.

PR #8516, opened 2026-05-10
remove MongoDB 4.4 support from docs

PieterGit:
@bewest @AndyLow91 What's the ambition of MongoDB support in the upcoming Nightscout releases?
I think we should drop supporting MongoDB 4.4.x now @AndyLow91 added MongoDB 5.x and MongoDB 6.x support.

Since upgrading MongoDB should be done one major version at the time, I think it would be good to mimic that in the upcoming Nightscout major releases, e.g. 

image

I don't know what the status of the current Mongo Atlas support is. I don't use it.
Currently Mongo Atlas has only support for MongoDB 7.0 and higher.
MongoDB 6.0 in Atlas (released July 2022, EOL July 2025) 
See: https://www.mongodb.com/legal/support-policy/lifecycles#atlas

I can help fixing docs and doing testing, but unfortuately my programming skills are not adequate enough to regain full MongoDB and Node support in Nightscout.


AndyLow91:
I agree with the direction here. I think Nightscout should only support MongoDB versions that are still supported upstream by MongoDB, but we should not cut users off cold without a migration window.

Current MongoDB upstream lifecycle dates are:

| MongoDB version | Upstream EOL |
|---|---:|
| 4.4 | February 29, 2024 |
| 5.0 | October 31, 2024 |
| 6.0 | July 31, 2025 |
| 7.0 | August 31, 2027 |
| 8.0 | October 31, 2029 |

So I agree that MongoDB 4.4 should be considered deprecated for Nightscout now. My preference would be that we start giving users clear notice that MongoDB 4.4 support is going away and that they should upgrade as soon as reasonably possible for their own environment.

I have already bumped the MongoDB version we write in the Nightscout `docker-compose.yml` example to `5.0.32` to guide new/self-hosted installs in the right direction. I think the README should do something similar, but we should be careful with the wording while CI still tests 4.4.

Suggested approach:

- document MongoDB 4.4 as deprecated, not recommended, and scheduled for removal
- recommend MongoDB `5.0.32+` or newer for now
- keep CI/docs aligned: once we remove MongoDB 4.4 from the test matrix, then README can say it is no longer supported
- give users upgrade-path guidance, since MongoDB major version upgrades should be done one major version at a time

So I support the intent of this PR, but I’d prefer wording that gives notice first rather than immediately saying 4.4 is unsupported while our CI still tests and passes it.

sulkaharo:
Looks to me like the next update should probably target 7.0 support. I was just looking at one Linux box and discovered it's difficult to maintain the older versions, so while it's fine to be compatible with older, it's reasonable to assume any new installation or existing installation that wants security patches will be on the 7 series.

PR #8517, opened 2026-05-10
Remediate runtime dependency CVEs

AndyLow91:
## Summary

This is the first staged dependency-security PR. It focuses on lower-risk runtime CVE remediation without taking the high-risk D3 major upgrade.

Changes included:

- Updates runtime dependencies for Express/body parsing, Socket.IO, and DOMPurify.
- Removes the `jsdom-purify` alias and uses the existing direct `jsdom` dependency in `lib/server/purifier.js`.
- Removes `parse-duration` and replaces it with an internal CommonJS careportal duration parser intended to preserve existing handling for blank, numeric, unit-suffixed, invalid, and cancel cases.
- Adds targeted `overrides` for vulnerable transitive runtime dependencies.
- Adds focused tests for careportal duration parsing and purifier behavior.

## Deferred

`d3` is intentionally not upgraded here. Clearing the `d3-color` finding requires a D3 major migration, and Nightscout still uses legacy APIs such as `d3.event` and `d3.mouse`, so that should be handled in a dedicated chart/report migration PR.

## Validation

Ran:

- `npm install --ignore-scripts`
- `npm ls` for the targeted runtime packages and overrides
- `npm audit --omit=dev`
- `npm run bundle`
- `npm run test:core`
- `npm run test:unit:ci`
- `npm run test:integration:ci`
- `npm run test:stress:ci`

Also smoke-tested the deprecated `bridge` / `mmconnect` initialization through the existing test coverage.


bewest:
Looks good.

bewest:
Embedding jsdom felt odd to me, and it was introduced "recently" as part of the admin notifications in order to sanitize.  There are other tools to sanitize the input, like https://www.npmjs.com/package/xss?activeTab=dependents, and we can correctly html encode the output as well.
Since jsdom represents an actual risky surface area, I'd like to eliminate using it in production code, in favor of using xss/input handling, rejecting invalid inputs and output encoding.


Recommendations

   1. Don't block on this for the current modernization track. PR #8517's change (dropping the
  alias) is a strict improvement and worth taking as-is.
   2. File a follow-up ("replace server-side dompurify+jsdom with xss or remove input-time
  sanitization in favor of output encoding") — this would let you drop jsdom from the runtime
  image entirely once test/jsdom is the only consumer (your modernize-jsdom track is already
  moving that direction).
   3. If you do swap, audit the eventType/notes/profile-name fields for downstream consumers
  that currently rely on DOMPurify's specific transforms (e.g. < → <). xss does the same
  escaping but the exact output bytes differ slightly.
   4. The iterate() walker has its own bugs worth fixing in the same PR: for…in enumerates
  inherited keys, isNaN(value) mis-classifies booleans/null/undefined, and it never returns the
   mutated object even though purifyObject does return iterate(obj).

  Net: the jsdom embedding is correct given the library choice; the library choice itself is
  heavier than this use case warrants and is a good candidate for a future cleanup PR — but not
  blocking #8517 or your track2 work.

  Phase A — Replace the sanitizer (no behavior surprise)

   - Swap lib/server/purifier.js to use xss (or sanitize-html with a strict empty-allowlist
  config that mimics DOMPurify's default "strip all tags").
   - Keep the same purifyObject(obj) signature so all four call sites are untouched.
   - Fix the iterate() bugs at the same time (Object.hasOwn, drop isNaN heuristic, handle
  arrays explicitly, return the mutated object, guard against cycles).
   - Add a golden-output test comparing the new sanitizer's results to DOMPurify's on a corpus
  of representative notes/eventType/profile-name strings (including the captured fixtures in
  tests/fixtures/captured/). Anything where bytes differ is a deliberate decision, not a
  regression.
   - Remove dompurify and jsdom from dependencies (jsdom may still appear as a devDependency
  for tests — that's fine; it never ships to the request path).
   - Validates: npm ls jsdom shows zero runtime edges.

  Phase B — Audit & harden output encoding

   - Grep for <%- (EJS raw), .html(, innerHTML, outerHTML, insertAdjacentHTML, document.write
  across views/, lib/client*, static/. Each hit is either "trusted constant" or "needs
  encoding".
   - For client-side, prefer .text()/textContent. For URL-context strings (e.g. href),
  encodeURIComponent.
   - For EJS server-rendered pages, ensure all interpolation of user/profile/treatment fields
  uses <%= %>.
   - Add a small lint rule (eslint-plugin-security or a custom selector via
  eslint-plugin-no-unsanitized) so future PRs can't reintroduce raw HTML sinks unreviewed.

bewest:
Updated to use library sanitize-html instead of jsdom.

PR #8533, opened 2026-05-27
Fix: mg vs mmol in displaying treatments

stevoh6:
# Fix: mg/dL vs mmol/L unit handling for displaying treatments

## Problem

Treatments with glucose values were displayed incorrectly when the stored unit didn't match
the display unit, or when no unit was stored at all.

**`treatmenttocurve.js`** — When a treatment had no `units` field, the code blindly assumed
the value was in whatever unit the *user's display preference* was set to. This caused mg/dL
values to be interpreted as mmol/L (or vice versa) for users whose stored treatments predated
unit tracking.

**`report_plugins/utils.js`** — Unit mismatch detection relied on comparing `treatment.units`
with `client.settings.units`, which only worked when both were set. When `treatment.units` was
missing, the raw value was passed through unchanged — meaning an mg/dL value like `120` could
be displayed as `120 mmol/L`.

---

## Decision logic — `treatmenttocurve.js`

```
Is treatment.glucose present?
│
├── NO  → synthesize BG from nearby SGVs via mgdlByTime()
│
└── YES
    ├── is NaN? → warn + skip (no BG set)
    │
    └── is valid number
        ├── are treatment.units set (loaded from  DB)
        │   └── YES → use it directly (always trusted, no guessing)
        │
        ├── is value >= 40 → use mg/dL
        │
        ├── is value <= 25 → use mmol/L
        │
        └── is value in 26–39 and  unit stored
            ├── is SGV available → guessUnitByProximity()
            │   └── picks whichever interpretation is numerically 
            │       closer to the nearest interpolated SGV (mg/dL)
            └── NO SGV → mgdlByTime() returns 180 mg/dL as default fallback
                └── biases toward mg/dL in this edge case
```

---

### `lib/units.js` — three new helpers

| Function | Purpose |
|---|---|
| `normalizeGlucoseUnit(unit)` | Normalizes `"mg/dl"`, `"mgdl"`, `"mmol/l"`, `"mmol"`, etc. to canonical `'mgdl'` or `'mmol'`; returns `null` if unrecognized |
| `guessUnitFromValueAndHint(value, unitHint)` | Returns recognized unit hint if present; otherwise infers from value range (≥ 40 → `mgdl`, ≤ 25 → `mmol`, 26–39 → `null`) |
| `guessUnitByProximity(glucoseValue, refMgdl)` | For ambiguous 26–39 range: picks the interpretation numerically closest to a known mg/dL reference |

### `lib/data/treatmenttocurve.js`

Replaced the old `settings.units` fallback with the new guessing logic above.
Ambiguous values in the 26–39 range are resolved by comparing proximity to the
nearest interpolated SGV via `mgdlByTime()`.

### `lib/report_plugins/utils.js`

Replaced the `treatment.units !== client.settings.units` mismatch-comparison path.
All treatment glucose values are now normalized to mg/dL first using the same
`guessUnitFromValueAndHint` / `guessUnitByProximity` logic, then scaled for display
via `client.utils.scaleMgdl()`.

---

### Bug fix — `lib/units.js`

`mmolToMgdl` was returning a string (`.toFixed(0)`) instead of a number, breaking
three pre-existing assertions in `units.test.js`. Fixed by removing `.toFixed(0)` so
the function returns `Math.round(...)` directly.

### `tests/units.test.js` — new describe blocks

| Suite | Cases |
|---|---|
| `normalizeGlucoseUnit` | mg/dl variants → `'mgdl'`; mmol/l variants → `'mmol'`; null / empty / unknown → `null` |
| `guessUnitFromValueAndHint` | recognized hint trusted; ≥ 40 → `mgdl`; ≤ 25 → `mmol`; 26–39 no hint → `null` |
| `guessUnitByProximity` | value closer as mg/dL → `'mgdl'`; value closer as mmol/L → `'mmol'` |

### `tests/data.treatmenttocurve.test.js` — new test case

`resolves treatment glucose unit from value when units field is missing` exercises four
treatments against SGVs around 90–100 mg/dL:

| Treatment | Input | Expected |
|---|---|---|
| `no_unit_high` | `glucose: 120`, no units | `mgdl = 120` (≥ 40 rule) |
| `no_unit_low` | `glucose: 6.5`, no units | `mmol = 6.5` (≤ 25 rule) |
| `no_unit_ambig` | `glucose: 35`, no units, SGV ~90 | `mgdl = 35` (proximity) |
| `invalid_glucose` | `glucose: 'abc'` | skipped — no `mgdl` or `mmol` set |

### DIFF
--- a/lib/data/treatmenttocurve.js
+++ b/lib/data/treatmenttocurve.js
@@ -2,6 +2,7 @@
 
 var _ = require('lodash');
 var consts = require('../constants');
+var { guessUnitFromValueAndHint, guessUnitByProximity } = require('../units')();
 
 const MAX_BG_MMOL = 22;
 const MAX_BG_MGDL = MAX_BG_MMOL * consts.MMOL_TO_MGDL;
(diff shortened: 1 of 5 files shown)


PieterGit:
Hi @stevoh6 . Can you explain where and why this goes wrong and with what setup. Nightscout should store data in mg/dl and only use mmol for display. I think fixing the root is more important than creating a workaround with ` guessUnitFromValueAndHint`.

PieterGit:
Hi @stevoh6. Can you give more insight in what causes the displaying problems for you. I think fixing that is more the way forward. Let me known if you more information. Otherwise I will close this PR.

stevoh6:
Hi @PieterGit  as you mention, NightScout should and is does store data in mg/dl.

Problem is when the data (for example in profile data or carbs treatment) are store without units. - in mine case they are comming from tandem source. 

I created fix which will make sure that what user wants to see, he will see.

 But in general I agree the root casue should be fixed first.  In this case is probably storing mg/dl without units. Mine understanding was that its working correctly - storing mg/dl, only displaying part was wrong - if units are missing Nigthscout display mg/dl instead of user mmol.

I hope that I explain it correctly. During whole time this fix is working on mine istance just fine - and also other plugins like bolus calc or profile editor forking okay now.

PR #8538, opened 2026-06-13
Render

chaunguyenhoabinh:


PieterGit:
Hi @chaunguyenhoabinh. I don't know Render. Can you explain if this is some kind of template, or that all environment variables will stored in Render and this vanilla `render.yaml`  is enough. New changes must be applied to `dev` branch and can you also supply more documentation in https://github.com/nightscout/cgm-remote-monitor/blob/dev/README.md 

Do you know if more people are interested in this service? If so, I would like other also to confirm it is working.

AndyLow91:
Closing as stale and we do not know if this is still required.

PR #8547, opened 2026-06-26
Use profile isAPNSProduction as source of truth for Loop APNs environment

bjorkert:
### Summary

When sending Loop remote-command push notifications, the APNs environment (production vs. sandbox) is currently determined solely by the `LOOP_PUSH_SERVER_ENVIRONMENT` environment variable. This means the environment has to be configured manually on the Nightscout site and can drift out of sync with how the Loop app itself is actually built.

Loop is about to start writing an `isAPNSProduction` boolean to the uploaded profile (see LoopKit/NightscoutKit#1). This is an optional, top-level profile field that is only serialized when set. This PR makes that field the source of truth when present, and keeps `LOOP_PUSH_SERVER_ENVIRONMENT` as the fallback.

### Behavior

- If the profile contains `isAPNSProduction`, it is used directly (`true` → production, `false` → sandbox).
- If the field is absent, behavior is unchanged: fall back to `LOOP_PUSH_SERVER_ENVIRONMENT === "production"`.

The `!== undefined` check (rather than a truthy fallback) is deliberate so that an explicit `isAPNSProduction: false` correctly forces the sandbox environment even when the env var says `production`.

### Compatibility

Fully backward compatible. Existing deployments where Loop does not yet write the field, or that don't run Loop, behave exactly as before.

### Related

- LoopKit/NightscoutKit#1 — adds the `isAPNSProduction` field to the uploaded profile.


AndyLow91:
Thanks for this contribution. This change was incorporated into dev through #8558 (commit 2dae077a6175ce6751b24f4ad462a12d7b9f0a6a) and is included in release 15.0.8. The implementation in dev matches this PR, and the three focused regression tests covering profile true/false overrides and the environment-variable fallback pass. This PR can be closed as completed; no additional merge is needed.

PR #8549, opened 2026-06-28
Don't snooze high alarm when there is no IOB (#6348)

### LINKED ISSUE #6348: Alarms Snoozed when IOB add-on enabled but not being used
**Describe the bug**
I enabled all add-ons in "config vars" just in case I want to use them in the future. 

I am not sure how but alarms were snoozed last night with this: "Snoozing high alarm since there is enough IOB". 

Nightscout does not know anything about IOB. We don't use IOB yet. We do not have it connected to the insulin pump. How does it know how much insulin he has on board?

Nov 02 02:51:03 nightscout-2020 app/web.1 2020-11-02T10:51:02.935Z	SNOOZING ALARM:
Nov 02 02:51:03 nightscout-2020 app/web.1   {"level":"Warning","title":"Warning HIGH","message":"BG Now: 273 +21 ⇈ mg/dl","group":"default","plugin":"simplealarms","debug":{"lastSGV":273,"thresholds":{"bgHigh":300,"bgTargetTop":260,"bgTargetBottom":80,"bgLow":55}}}
Nov 02 02:51:03 nightscout-2020 app/web.1   BECAUSE:
Nov 02 02:51:03 nightscout-2020 app/web.1     {"level":"Urgent","title":"Snoozing high alarm since there is enough IOB","message":"\n","group":"default"}

**To Reproduce**
1. Go to config vars
2. Add IOB
3. Wait for Alarm
4. See: "Snoozing high alarm since there is enough IOB"

**Expected behavior**
If not using IOB, should alarm.

**Screenshots**

**Your setup information**
- What version of Nightscout: Latest: https://github.com/nightscout/cgm-remote-monitor/releases/tag/14.0.6

- What type of CGM, and how do you get your data there? G6 with IOS, Share+Follow & Heroku 

**Additional context**

IsraelAfangideh:
Fixes #6348.

## Problem

The Bolus Wizard Preview plugin snoozes a high alarm when `prop.bolusEstimate < snoozeBWP`, treating a near-zero estimate as *"there is enough insulin on board to cover the high"*. But `bolusEstimate` starts at `0` and stays `0` in a case that has nothing to do with IOB coverage:

- the BG is above the alarm threshold (`bgTargetTop`) yet still **below the profile high target**, so neither the high nor the low branch of `calc()` runs.

With **no insulin on board**, `bolusEstimate` is then left at its initial `0`, and a genuine high alarm gets silenced. This matches the report in #6348: a complete profile, a default ISF, and no insulin treatments. (The exact `target_high` from the report is not in the issue, so this is the most likely mechanism consistent with it rather than a certainty.)

Because the snooze is requested at `URGENT` level and `notifications.snoozedBy()` matches by level across the group, this can also suppress *other* urgent alarms for the snooze duration.

> Note: the other path that leaves `bolusEstimate` at `0` — `calc()` returning early with `errors` (missing profile/IOB, stale data) — is already handled, since `checkNotifications()` returns early when `prop.errors` is non-empty before `highSnoozedByIOB` is ever called. So this change targets only the no-error, no-IOB case above.

## Fix

Guard the snooze on `prop.iob > 0`, so a high alarm is only snoozed when there is actually insulin on board to cover it:

```js
return high && prop.iob > 0 && prop.bolusEstimate < settings.snoozeBWP;
```

Legitimate *enough-IOB* snoozing is unchanged. No new dependencies.

## Tests

Added to `tests/boluswizardpreview.test.js`:
- does not snooze a high alarm with no IOB when the BG sits below the profile high target (the #6348 scenario — fails without the fix, passes with it)
- still snoozes a high alarm when real IOB covers it (unchanged behavior)

Full suite green locally (`1346 passing`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AndyLow91:
Thank you for the alarm-snoozing fix and regression tests. These changes were incorporated through merged PR #8558 and shipped in [Nightscout 15.0.8](https://github.com/nightscout/cgm-remote-monitor/releases/tag/v15.0.8).

I verified that the original patch matches release commit dee62a92, including the positive-IOB guard and both regression tests. The relevant tests pass, and issue #6348 is now closed as completed.

Closing this original PR as already incorporated through #8558. Thank you for the contribution.

PR #8580, opened 2026-08-11
Add configurable custom notification webhooks (#5742)

### LINKED ISSUE #5742: Custom WebHook Support
Currently NS supports IFTTT WebHooks, however due to notifications sent to IFTTT, it throttles the requests meaning they are not actioned immediately

The ability to take the same IFTTT WebHooks, but use a custom URL to send the request to
In the Configuration Variables, this could work like

`CUSTOM_WEBHOOK_1_URL:  "my-endpoint.com"`
`CUSTOM_WEBHOOK_1_EVENT: "ns-low"`

Alternatively, this could be a POST or GET request, sending the current event data in the body, however this make again cause issue with throttling

joshi-rushikesh:
Closes #5742

## Why

Issue #5742 describes delays caused by routing Nightscout notification events through IFTTT Maker. Today, the Maker notification path is tied to an IFTTT key and a hardcoded IFTTT destination, so operators cannot route those same notification events directly to their own endpoints.

This PR adds an independent custom notification-webhook path while preserving the existing Maker/IFTTT behavior.

## What changed

Operators can configure up to four URL/event pairs:

```text
CUSTOM_WEBHOOK_URL_1="https://my-endpoint.example.com/nightscout"
CUSTOM_WEBHOOK_EVENT_1="ns-urgent"

CUSTOM_WEBHOOK_URL_2="http://192.168.1.50:3000/nightscout"
CUSTOM_WEBHOOK_EVENT_2="ns-event"
```

Matching notifications are delivered directly to the configured endpoint as JSON `POST` requests.

The implementation:

- adds a dedicated `lib/server/customwebhook.js` sender
- dispatches custom webhooks independently from Maker in `pushnotify`
- uses Nightscout's existing numbered-settings convention for configuration
- supports both HTTP and HTTPS
- uses a 5-second timeout and treats any 2xx response as successful
- skips malformed or incomplete configuration without preventing startup
- deduplicates by destination so one notification is not accidentally delivered multiple times to the same endpoint
- protects configured webhook URLs through `secureSettings`
- avoids logging full webhook URLs or notification contents
- leaves `lib/plugins/maker.js` and `lib/plugins/webhook.js` unchanged

### Independent from Maker / IFTTT

`lib/plugins/maker.js` is intentionally not modified.

Maker's existing three-event fan-out (`ns-event`, `ns-<level>`, `ns-<level>-<name>`) is an IFTTT-specific compatibility behavior that dates back to the original implementation. Extending the Maker sender directly would also inherit the requirement for a `MAKER_KEY`.

Instead, this PR adds a sibling delivery path from the notification dispatcher. Custom webhooks therefore work even when `ctx.maker === null`, while existing Maker behavior remains unchanged.

### Configuration naming

The issue illustrates names such as `CUSTOM_WEBHOOK_1_URL`, but Nightscout's current numbered-setting convention maps naturally to:

```text
CUSTOM_WEBHOOK_URL_1
CUSTOM_WEBHOOK_EVENT_1
```

This matches existing settings such as `FRAME_URL_1` and avoids changing the global settings-name parser.

### Event matching

Custom webhook event matching follows Nightscout's existing Maker event vocabulary:

- `ns-event`
- `ns-<level>`
- `ns-<level>-<name>`
- `ns-allclear`

For low blood glucose alarms, the level-qualified forms are used, for example:

```text
ns-urgent-low
ns-warning-low
```

There is no bare `ns-low` event in the current event vocabulary.

### Delivery behavior

A single configured destination is sent at most one request for a notification, even if more than one generated event name would match that same destination.

The JSON payload includes the notification fields needed by the receiving endpoint, and the transport uses Node's built-in `http`, `https`, and `URL` APIs. No new dependencies were added.

## Security considerations

Webhook URLs may contain credentials or tokens, so the new URL settings are included in `secureSettings` and are not exposed through normal serialized settings output.

The implementation also:

- only accepts `http:` and `https:` schemes
- does not log full URLs, query strings, titles, or message bodies
- skips malformed URLs rather than crashing startup
- reports network failures through callbacks rather than throwing into the alarm path

Custom webhook URLs are operator-controlled configuration. They can point to private-network hosts, which is intentional for this use case and consistent with Nightscout's existing server webhook functionality. This is documented because it creates the expected SSRF surface of an operator-configurable outbound webhook feature.

## Backward compatibility

The default configuration contains no custom webhook destinations, so the new module initializes to `null` and produces no additional outbound traffic unless explicitly configured.

Existing behavior is preserved:

- `lib/plugins/maker.js` is unchanged
- `lib/plugins/webhook.js` is unchanged
- existing Maker key handling is unchanged
- existing Maker three-request fan-out is unchanged
- existing SGV webhook behavior is unchanged
- custom webhook delivery does not depend on `MAKER_KEY`
- custom delivery does not modify the existing `recentlySent` TTL or Pushover receipt handling

### Focused feature tests

```bash
npx env-cmd -f ./my.test.env npx mocha --timeout 5000 --require ./tests/hooks.js --exit ./tests/customwebhook.test.js
```

Result:

```text
26 passing
0 failing
```

### Related regression suites

```bash
npx env-cmd -f ./my.test.env npx mocha --timeout 5000 --require ./tests/hooks.js --exit ./tests/maker.test.js ./tests/webhook.test.js ./tests/settings.test.js ./tests/env.test.js
```

Relevant results:

```text
maker:        6 passing
webhook:     10 passing
settings:    16 passing
combined:    55 passing, 0 failing
```

The existing Maker and SGV-webhook test files were not modified.

### Broader unit suite

The broader local unit suite produced:

```text
293 passing
6 failing
```

The six failures are MongoDB-dependent local-environment failures in the existing security/authentication tests. To verify they were not introduced by this change, I temporarily removed the custom-webhook source changes and re-ran the failing suites; the same four `verifyauth` timeouts and two `security` failures remained.

### Lint

`npm run lint` reports the same 32 pre-existing problems with or without the custom-webhook source changes.

Running ESLint against the new/modified source and test files produces no new lint problems.

### Manual transport verification

Because automated tests stub the outbound request seam, I also exercised the real transport against a local HTTP listener:

```text
204 response -> success
500 response -> failure
dead port -> callback error, no throw
query string -> preserved
JSON body -> delivered
custom delivery with ctx.maker === null -> successful
```

## Acceptance criteria

- [x] Tests added for new/changed behavior
- [x] Relevant feature and regression suites pass
- [x] New and modified files introduce no new lint problems
- [x] Existing Maker behavior remains unchanged
- [x] Existing SGV-webhook behavior remains unchanged
- [x] No breaking changes intentionally introduced
- [x] Documentation updated
- [x] Pull request targets upstream `dev`
- [x] Original issue referenced with `Closes #5742`
- [ ] Full local unit suite completely green — 6 MongoDB-dependent failures remain, but they were reproduced without the feature changes and documented as pre-existing local-environment failures

## Before / after verification evidence

Before implementation, reproducing an urgent-low notification through the existing path generated three requests to the hardcoded IFTTT Maker endpoint and zero requests to the configured custom endpoint:

```text
maker.ifttt.com requests: 3
configured custom endpoint requests: 0
```

After implementation:

```text
customwebhook tests: 26 passing
related Maker/webhook/settings/env suites: 55 passing, 0 failing
custom delivery with ctx.maker === null: 1 destination delivered
```

The existing Maker and SGV-webhook implementations remain unchanged.

## Documentation

This PR updates:

- `README.md` with a new Custom Notification WebHooks section
- `docs/example-template.env` with commented configuration examples

The README documents the configuration variables, event matching behavior, JSON delivery behavior, low/high alarm event naming, and security considerations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md)
       - [Extended Settings](#extended-settings)
       - [Pushover](#pushover)
       - [IFTTT Maker](#ifttt-maker)
+      - [Custom Notification WebHooks](#custom-notification-webhooks)
     - [Treatment Profile](#treatment-profile)
   - [Setting environment variables](#setting-environment-variables)
     - [Vagrant install](#vagrant-install)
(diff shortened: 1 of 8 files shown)


joshi-rushikesh:
Hi @AndyLow91 — this is my first contribution to Nightscout. I implemented #5742 by adding configurable notification-event webhooks as an independent delivery path while preserving the existing Maker/IFTTT behavior. I also added focused tests and documentation. Since you reviewed the related server-webhook contribution #8427, I’d really appreciate a review when you have time. Thank you!

PR #8587, opened 2026-08-29
Use the COB reported by the uploading system

bjorkert:
The COB pill was showing Nightscout's own carb decay calculation instead of the value the uploading system reports. The plugin already preferred device status, but the treatment profile check ran first, so a site with no profile, or one without sens and carbratio, got no COB pill at all even though the reported value needs neither. AndroidAPS also sends suggested without a timestamp and enacted with neither a COB nor a timestamp, and moment(undefined) resolves to the current time, so those records were dated wrong and the pick between the two blocks depended on which millisecond each call landed in. The OpenAPS branch keyed off the openaps block being present rather than a usable COB, so it could never fall through to loop.cob.

The extraction moved to lib/client-core/devicestatus/cob.js next to the other pure device status modules. A block counts only when its COB is a finite number, and its timestamp dates it, falling back to the record mills. The freshness rules are unchanged, still 30 minutes to pick a candidate and 10 to accept it.

The tooltip now names the source and device and shows the treatment derived value when that is not zero. On the properties API cob.treatmentCOB is a number instead of an object, and cob.source is translated. The aaps and trio pill goldens gain a COB pill, and the loop one gains the source and device rows.


AndyLow91:
Added commit c7e6a589 to address the CI coverage gap found during review.

Why:
- The current test-ci command only matches tests/*.test.js, so the client-core suite, including the new device-status COB extractor tests and captured pill goldens, was not part of the green PR checks.
- The pill goldens include legacy Date.toLocaleString() output and changed with the runner timezone, so simply enabling the suite would have made it environment-dependent.

Changes:
- Pin the shared captured-pill runner to America/Los_Angeles, matching the existing committed goldens for both generation and verification.
- Run npm run test:core in the existing Node/Mongo CI matrix.

Validation:
- TZ=UTC npm run test:core: 283 passing
- TZ=UTC golden regeneration dry run: all four fixture sets unchanged
- tests/cob.test.js: 19 passing
- ESLint: no errors
- Workflow YAML parsed successfully

PR #8605, opened 2026-09-05
Modernize Nightscout runtime, dependencies, and memory usage

AndyLow91:
Complete the agreed M01–M30 Nightscout 15.0.9 modernization work on `chore/nightscout-modernization`, now at **`ee2a0b9ee83b26f811da8a096d6ebdb31e85780a`**. This remains the **draft integration PR to dev**, with proxy compatibility correction #8723 merged after green child CI; the refreshed parent checks are running. Nothing is automatically promoted to dev or claimed shipped.

The combined modernization child #8721 includes #8719's widget/Flot decision and #8720's connector shutdown pin, with database rehearsal #8718 already integrated. Security follow-up #8722 removes query-credential authentication from the measurement fixture and constructs canonical API3 Location headers from the registered route and encoded resource identifiers. New regressions cover the unsafe constructions and real API create/deduplication/readback behavior. No finding was suppressed or dismissed by this follow-up.

Proxy compatibility correction #8723 restores default forwarded HTTPS and legacy client-IP handling for existing ingress deployments, including changing pod addresses. Explicit IP/CIDR trust remains available, and `TRUST_PROXY=false` selects direct connections. The compatibility default relies on a sanitizing ingress and controlled backend access; it does not claim the spoofing protection of explicit trust. Legacy client-header precedence is now deterministic.

All 22 [child CI checks](https://github.com/nightscout/cgm-remote-monitor/actions/runs/34220986651) passed, including the revised native amd64/arm64 Docker default-proxy redirect smoke. CodeQL reports zero findings on the tested child merge. All 13 child browser/recovery artifact directories match virtual merge `eddac870152be976d332b8245cc9aae96da0b086`, including source hashes, bundle identity and owned database cleanup. The actual integration and independently computed dev merge tree match tested tree `54d3ab849491551f984451b79573c1d6d2d680ea`. Refreshed [parent CI](https://github.com/nightscout/cgm-remote-monitor/actions/runs/34221797010) is running on virtual merge `a1895d734d83ebc217a5671e1d90b2422d5cde82`, which has that same tree. [Parent CodeQL](https://github.com/nightscout/cgm-remote-monitor/actions/runs/34221797004) has completed successfully, and this merge ref has zero open code-scanning alerts.

Completed decisions include:
- Supported Node 22.23.2/24.20.0, maintained dependency/API consumers, build/runtime separation, narrower browser page entries and required real-browser tests.
[...]

PR #8613, opened 2026-09-05
Upgrade jsdom to 30 and preserve DOM test isolation

AndyLow91:
**ON HOLD — draft fallback.** The modernization plan now prioritizes [complete jsdom retirement](https://github.com/nightscout/cgm-remote-monitor/blob/chore/nightscout-modernization/docs/plans/jsdom-retirement.md). This upgrade is not scheduled for merge. Useful lifecycle fixes can be evaluated independently against the existing dependency.

Upgrade the test-only DOM runtime from jsdom 26.1.0 to the current **30.0.1** release, now that M07 provides the required Node floors. This is the jsdom part of M08 in #8605; Babel and UUID remain separate reviews.

jsdom 28 replaced ResourceLoader with a resource-interceptor API. Migrate the secure test fixture using the public requestInterceptor helper, retaining its fetch/XHR blocking and explicit resource overrides. New repeated-window tests exercise real stylesheet, iframe, script and WebSocket attempts against a loopback server and verify zero requests.

The full-suite review exposed an existing legacy-harness bug: teardown deleted Node's native Event constructor, causing undici WebSocket failures in jsdom 30. Restore original global property descriptors instead, including accessor globals and bundle-exported jQuery values. A fresh-process regression fails on the old shim and checks replacement windows, partial/full/repeated teardown and native EventTarget dispatch over two cycles.

Validation:
- Clean locked installs and production/development builds pass. Both production browser bundles are byte-identical to the parent (app 1,761,400 bytes; clock 152,131 bytes).
- Full main suites pass on Node 22.23.2 and 24.20.0: 1,981 before M06 integration and 1,986 after M06 respectively, with three existing pending. Client-core 283; dependency compatibility 318; DOM/native-global/Socket.IO focused tests 35 on both floors.
- Refreshed onto current integration parent `8b57b7b8` (including M04); the combined DOM/Socket.IO/provider/API focused suite passes 114 cases. All current matrix, npm 12, CodeQL and Docker checks must pass before merge.
- All 672 production-marked lock entries are unchanged. The development graph adds 17 paths, removes seven and updates 15 retained entries, all within jsdom's graph. No server-memory or production-package-size saving is claimed.

Major-version/API review and reproduction details are in `docs/test-specs/jsdom-ws-tests.md`, with upstream release/API links. No application UI, browser target, persisted data or configuration change. Rollback the dependency and resource-interceptor migration together and perform a clean install; the native-global restoration can be retained independently.

This child targets `chore/nightscout-modernization` and records M04's completion evidence. #8605 remains the single draft integration PR into dev.

The initial Node/MongoDB matrix, npm 12 and CodeQL checks passed, but both native Docker builds found a real install regression: Docker omitted `.npmrc`, causing npm to reinterpret the optional Undici 5 peer of http-cookie-agent against jsdom’s Undici 8 dependency. Commit `192994ad` copies the repository settings into the builder so installation uses the same resolution policy as the lockfile. A clean Node 22 install with the Docker flags, postinstall build and production prune now passes locally; the complete lockfile is unchanged and jsdom/Undici are absent after pruning. All hosted checks will run again before merge.


AndyLow91:
The initial Node/MongoDB matrix, npm 12 and CodeQL checks passed, but both native Docker builds found a real install regression: Docker omitted `.npmrc`, causing npm to reinterpret the optional Undici 5 peer of http-cookie-agent against jsdom’s Undici 8 dependency. Commit `192994ad` copies the repository settings into the builder so installation uses the same resolution policy as the lockfile. A clean Node 22 install with the Docker flags, postinstall build and production prune now passes locally; the complete lockfile is unchanged and jsdom/Undici are absent after pruning. All hosted checks will run again before merge.

AndyLow91:
User direction clarified: prioritize completing jsdom removal. The prior server-side removal is already complete (`npm ls jsdom --omit=dev` is empty). Remaining consumers are test-only. I have put #8613 back into draft and will not merge its major upgrade merely because CI passes.

Plan update `d87f77e6` adds the [retirement sequence](https://github.com/nightscout/cgm-remote-monitor/blob/chore/nightscout-modernization/docs/plans/jsdom-retirement.md): establish an isolated browser-test baseline, migrate actual bundle/Socket.IO coverage, move pure logic to Node tests and real DOM/security contracts to browser tests, then remove jsdom after assertion parity and all CI gates pass. Developer/CI installation and memory costs will be measured; no further production-memory saving is claimed. MongoDB 5/6 remain supported during migration.

PR #8615, opened 2026-09-05
Replace bundle and Socket.IO jsdom tests with real-browser coverage

AndyLow91:
Replace the jsdom bundle smoke test and four served Socket.IO browser cases with actual-browser coverage, preserving their namespace, transport, Unicode/binary payload, acknowledgement and repeated-reconnect assertions. Node parser/client coverage stays in Mocha. This is the first concrete jsdom retirement slice; remaining consumers still need migration before the package can be removed.

Use the existing Mocha runner with pinned development-only `playwright-core` 1.63.0. The unused Playwright wrapper/test-runner package is omitted, saving one path and 5.1 MB versus the comparison prototype. Browser downloads happen only in an explicit install step; all retained lock entries and production bundles remain unchanged. Required CI exercises Chromium, Firefox and WebKit on both exact supported Node floors, and Docker jobs depend on browser and backend tests.

A fresh context permits one exact local origin, blocks other HTTP/WebSocket destinations and service workers, rejects redirects, and closes after success or failure. Tests cover fetch/XHR/subresources/worker requests, redirect escape, storage isolation and repeated failure cleanup. Chromium's fixture-origin local-network permission allows native WebSockets while the destination policy stays enforced. The namespace regression fails with a deliberately broken reportclient export.

Validation:
- All six browser jobs passed with the old tests retained during comparison; then the mapped jsdom cases were removed.
- Refreshed local main suite: **2,004 passing, three existing pending**; core: **283**; dependency compatibility: **313**. Nine old Node cases moved into five grouped browser cases without dropping their assertions.
- Browser suite: **eight passing** in local Chromium/WebKit on supported Node floors; Linux CI passed Firefox too. Firefox temporary-profile launch fails on this Mac, documented as a local tooling limitation.
- Clean install, production/development builds, changed-file lint and actionlint passed. Both production browser bundles are byte-identical to the parent.
- Final CI passed against modernization parent `182e5694`: eight Node/MongoDB jobs, six browser jobs, npm 12, CodeQL and both native Docker startup checks. The final merge tree matches the verified proposal. Merged into modernization as `eb01da5b`.

Real browsers increase test resource use. Seven-process comparison medians were 0.90s / 173 MB summed process-tree RSS for the selected jsdom cases, 1.14s / 487 MB Chromium and 1.89s / 329 MB WebKit. This metric can double-count shared pages and is not server heap. The core package adds 13.45 MB of installed contents; explicitly installing all macOS engines adds about 1.18 GB of browser files. No production size/memory saving is claimed. The [specification](docs/test-specs/browser-tests.md) records measurements, coverage mapping, isolation limits, older-iOS scope and rollback.

M08 of #8605; targets `chore/nightscout-modernization`. #8613 remains on hold. Roll back this child as a whole to restore the old test/CI path; no application or data migration is involved.


AndyLow91:
All final checks passed for `d241ddfb`: eight Node/MongoDB jobs, six real-browser jobs (Chromium/Firefox/WebKit × both Node floors), npm 12, CodeQL and native amd64/arm64 Docker startup. The exact proposed merge tree matches current integration parent `182e5694`. The migrated jsdom cases are removed with their assertions preserved; remaining jsdom consumers are tracked separately. Production bundles and retained dependency entries are unchanged.

PR #8639, opened 2026-09-06
Replace async with bounded callback tasks

AndyLow91:
Replaces direct async 0.9 usage with a small bounded callback helper for dataloader, treatment pre-bolus writes, Maker, Alexa and Google Home. Preserves input ordering, callback timing, serial writes and the ten-handler concurrency limit while removing the package.

Maker now completes every key in a stage before advancing and stops later sends on error. Voice rollups return completed text once when a handler fails instead of throwing on missing results. Focused tests reproduce these four failures against the parent and verify two successive event/rollup cycles. A real MongoDB API test uploads the same pre-bolus batch twice without duplicate records.

Validation:
- 20 focused cases pass on Node 22.23.2 and 24.20.0; client-core 283 pass on Node 24.
- Real treatment API suite: 10 passing. Lint: no errors, two test-oracle dynamic-require warnings.
- Initial full backend: 1,627 passing, one pending, three socket hang-ups in API shape tests. An instrumented fresh-database full run then passed 1,631 cases with one pending, including the new replay test; it did not reproduce or explain the socket failures. Hosted CI remains open; this draft is not merge-ready.
- One installed lock path removed, no retained lock entries changed; 91,385 installed regular-file bytes removed. Six production JS bundles are byte-identical to the integration parent. No retained server-memory reduction is claimed.

M18 first slice only: bootevent replacement and boot lifecycle validation remain outstanding. Contracts, intentional behavior corrections and validation limitations are recorded in docs/test-specs/callback-tasks.md. Targets chore/nightscout-modernization; the combined #8605 stays draft into dev.

PR #8640, opened 2026-09-06
Replace bootevent with a local startup sequence

AndyLow91:
Replace bootevent with a local startup sequence, preserving the same fourteen stages, next-tick startup, asynchronous gating, shared context and context-only boot callback. This removes bootevent, chainsaw and its private traverse copy; no retained lock entries change.

Tests characterize the old and new queue together and exercise two real failed Nightscout boots with teardown, checking that configuration errors prevent storage/connector startup. Existing connector lifecycle and Axios settings-import tests now capture the stage array instead of mocking the removed package. Duplicate stage callbacks are explicitly rejected.

Validation:
- Seven old/new characterization cases pass; 40 focused cases pass on Node 22.23.2 and 24.20.0.
- Clean install/production build succeeds; six production JS bundles are byte-identical to the validated M16 build.
- Lint has no errors and one dynamic test-oracle warning.
- Three package paths and 102,370 installed regular-file bytes removed; no retained server-memory saving is claimed.
- Full backend run passes 1,621 cases with one pending against a fresh MongoDB 6 test database; hosted CI and final integration checks are required. This is a draft, not merge-ready.

M18 second slice; the separate async replacement is #8639. Targets chore/nightscout-modernization. #8605 stays draft into dev. Contracts and limitations are in docs/test-specs/boot-sequence.md.


AndyLow91:
The full local backend suite completed successfully at 26ddc260: 1,621 passing, one pending, against a fresh disposable MongoDB 6 database. The 40 focused tests also pass on Node 24. Hosted CI is still running, so this remains draft and unmerged. The checked-in test specification will receive final hosted/integration evidence when those gates complete.

AndyLow91:
Added a real successful-startup lifecycle test: boot, load data, verify a single data-received/data-loaded listener, and tear down; repeat with an independent context and bus. Both this test and the existing failed-startup lifecycle test pass against a fresh MongoDB 6 database. The initial fixture attempt exposed env() consuming API_SECRET; the fixture now supplies its owned secret separately for each boot and restores the inherited value.

Also refreshed against merged M17/#8637. Clean install and combined validation follow on the new tree; prior CI does not cover this new lifecycle test or merge. This remains draft.

AndyLow91:
The refreshed local full backend run at f266aa64 had one failure: native cachebuster compatibility / keeps development worker requests on the network reported socket hang up. The boot lifecycle cases pass, but this full run is not green. This resembles earlier local HTTP failures without establishing the cause. The PR remains draft pending investigation and exact-head hosted/integration checks.

AndyLow91:
The boot branch's full local backend suite passes with the proposed #8643 Supertest address-family adapter preloaded: 1,638 passing, one pending. Production code was unchanged for this diagnostic run. This supports integrating the fixture fix before final validation; it does not establish that every historical socket failure had the same cause. The run used a fresh disposable MongoDB 6 database.

AndyLow91:
Refreshed against integration a92d0882, including the merged callback helper and HTTP/browser fixture fixes. Resolved only an adjacent modernization-plan documentation conflict, preserving both work items. Head 723e42a3 is pushed; combined full backend validation and fresh hosted CI are running. M18 is not marked complete.

AndyLow91:
Combined full backend validation passes on 723e42a3 against integration a92d0882: 1,655 passing, one unrelated pending, using Node 22.23.2 and a disposable MongoDB 6 database. This includes both successful and failed repeated real boot/teardown tests and the integrated callback helper. Fresh hosted checks remain required before merge.

AndyLow91:
Previous head 723e42a3 completed its browser checks; refreshed to f1ab7034 to include the subsequently merged DST fix #8645. Merge was clean, with no boot implementation changes. Fresh checks on the current integration base are required before merge.

AndyLow91:
Merged as 1fab2a2426a25259de6767904a63ffc460757c97 after all required checks passed on f1ab7034. The actual merge tree 26e5c154 matches the reviewed preview. Together with #8639 this completes M18; #8605 remains draft into dev.

PR #8643, opened 2026-09-06
Match HTTP fixture requests to their address family

AndyLow91:
Make test requests use the loopback address family of their fixture server. Supertest currently starts an unspecified-address listener but builds an IPv4 URL. With IPv4 and IPv6 fixtures on the same numeric port, this can send a request to the wrong server; a deterministic local probe returned the other fixture's 405 response.

Install an idempotent adapter through the root test hooks for IPv6 wildcard listeners, and make both API fixture factories construct matching URLs. Production networking is unchanged. Two regression tests cover consecutive same-port IPv6 requests and explicit IPv4 handling; both pass on Node 22.23.2 and 24.20.0. Lint passes. Full backend validation is running, and hosted/integration checks remain required.

This addresses a proven wrong-fixture routing mechanism consistent with the local HTTP anomalies during modernization; it does not prove every previous socket failure had this cause and does not fix the separate Firefox interception stall. The adapter's dependency on Supertest.serverAddress is documented in docs/test-specs/http-fixture-addresses.md.

Targets chore/nightscout-modernization as a validation prerequisite for the pending cleanup children. #8605 remains draft into dev.

### DIFF
--- a/docs/test-specs/http-fixture-addresses.md
+++ b/docs/test-specs/http-fixture-addresses.md
@@ -1,0 +1,25 @@
+# HTTP fixture address families
+
+Supertest starts unbound fixture servers with listen(0), but its serverAddress
+implementation constructs an IPv4 loopback URL regardless of the bound family.
+An IPv6 wildcard listener and a separate IPv4 listener can own the same port.
+A local deterministic probe bound both and showed a request intended for the
+IPv6 fixture returning the IPv4 fixture's 405 response instead.
+
+The root test hooks now install an idempotent, test-only Supertest adapter:
+when the server is bound to IPv6 wildcard (::), use [::1] in the generated URL.
+Explicitly bound IPv4 fixtures retain their URLs. API fixture factories also
+choose their loopback address from the bound server family rather than relying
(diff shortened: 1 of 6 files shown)


AndyLow91:
The full local backend suite passes with the address-family adapter and API fixture changes: 1,634 passing, one pending, on Node 22.23.2 against a fresh disposable MongoDB 6 database. The deterministic same-port regression and IPv4 case also pass on Node 24. Hosted checks and exact merge verification remain required.

AndyLow91:
#8643 is merged as f96de163. All required backend, browser, npm 12, CodeQL and native Docker checks passed on 27da0f8a; the actual merge tree 74de1fa8 matches the verified proposed tree. The deterministic same-port regression passes on both Node floors, and the full local suite passed 1,634 cases with one pending.

This fixes the proven HTTP fixture address-family mismatch. Pending cleanup children will be refreshed against it before final validation. The separate intercepted browser fetch stall remains under investigation on #8639; the Connection: close candidate has passed the full local WebKit suite but still needs hosted validation. #8605 remains draft into dev.

PR #8651, opened 2026-09-06
fix: retain production import client without leaking configuration

AndyLow91:
The IMPORT_CONFIG boot stage uses Axios in production but declares it development-only, and its diagnostics expose full URLs and imported settings. Retain the existing client for proxy compatibility, move its declaration into dependencies, and keep only a fixed failure message plus numeric HTTP status in logs/boot errors.

M19 retain decision: an owned-proxy comparison on both Node 22.23.2 and 24.20.0 shows native fetch ignores the environment proxy without a process-startup opt-in. A global routing change or locally maintained proxy/CONNECT client is disproportionate to removing this one call. The plan allows retaining a dependency when replacement adds complexity or changes behavior. Connector versions/wrappers, request behavior, JSON merging and default timeout remain unchanged.

Validation: both new declaration/privacy regressions fail on the parent. All 29 Axios/boot contracts pass on both Node floors, including two authenticated proxy/NO_PROXY cycles and two successful/failed credential-bearing import cycles. Production build passes. The full Node 22/MongoDB 6 backend suite passes 1,670 tests with one existing pending. After npm prune --omit=dev (Mocha absent), the actual boot stage imports and merges settings twice on each supported Node version. All six browser bundles are byte-identical to the parent. Hosted combined CI remains required. Lockfile changes only direct dependency classification, with no installed entry/version/integrity changes and no claimed size/RAM saving.

Includes decision, observable diagnostic change and rollback documentation. Targets chore/nightscout-modernization; #8605 remains draft into dev. M19 completes only after verification and merge.


AndyLow91:
This candidate is now included in draft #8652, targeting chore/nightscout-modernization, alongside the other pending cleanup PRs. Exact source heads are recorded in docs/audits/cleanup-integration-heads.json. The combined branch preserves this PR's final implementation (the local-statistics successor intentionally includes the quantile batching work).

Full combined CI and exact merge-tree verification are required before integration. This source PR stays open and reviewable until that verified merge; its standalone green checks are not being used to skip combined validation. Any subsequent source-head change must be incorporated and revalidated in #8652. #8605 remains draft into dev.

PR #8652, opened 2026-09-06
chore: integrate and validate pending modernization cleanups

AndyLow91:
Combine the pending M08, M19, M23 and M26 cleanup PRs into one tested merge candidate. This avoids repeatedly moving the base beneath the other candidates while preserving each source PR and its review history.

Includes #8644 report quantile batching, #8646 scoped query leaves, #8647 local report statistics, #8649 native persisted UUID v5 identifiers, #8650 Babel 8/loader 10 and #8651 production import-client ownership/private diagnostics. Exact source heads are recorded in docs/audits/cleanup-integration-heads.json; every head is an ancestor. All 20 source/test/tool files match their final source PR, with #8647 intentionally succeeding #8644's report code.

Resolved manifest/plan conflicts retain production Axios and Babel 8, remove direct UUID/traverse/simple-statistics, and keep the Babel loader's new obug dependency while dropping traversal-only entries. A clean npm ci and production build pass. Existing private UUID dependencies remain under NYC/request; no claim that the whole UUID ecosystem is removed.

Combined local tests pass: 1,811 backend (one existing pending), 283 core, 273 dependency, and all 511 browser cases on each of Chromium/Node 22 and WebKit/Node 24. npm ls --all reports a valid tree. Development/HMR build passes with the same two existing warnings as the parent. All hosted Node/MongoDB, browser, npm 12, CodeQL and Docker checks are required before merge. This draft is not merge-ready; standalone source-PR checks do not substitute for combined validation. No task is marked complete merely by assembling this branch.

Targets chore/nightscout-modernization. #8605 remains the sole draft integration PR into dev. Source PRs will be reconciled after the verified combined merge; this does not merge dev or declare the full modernization plan finished.

PR #8653, opened 2026-09-06
perf: bound profile caching and share expiry timers

AndyLow91:
Replace the profile-only memory-cache dependency with a bounded reference cache and one expiry timer per instance. Preserve five-second get/put/clear semantics, caller reference mutations, exact-boundary reads, replacement expiry and repeated reset behavior. Notification caches remain unchanged because their cloning/acknowledgement contracts need a separate decision.

Nine characterization cases pass against the old and new cache; a tenth checks capacity/replacement/timer bounds. Backward/forward clock cases exposed an initial candidate regression: expiry scheduling now uses monotonic time while reads preserve legacy wall-time checks. Existing profile tests pass on Node 22/24. The cap is 524,288 entries, chosen to accommodate measured monthly minute-resolution lookups. Smaller trial caps caused repeated misses and were rejected. Above capacity, oldest-written computed values are evicted and recomputed; that bounded-memory trade-off is explicit.

Seven alternating paired processes over actual profile methods have identical outputs for 1/14/31-day workloads. Retained heap medians fall from 6.74 to 2.70 MB (day), 83.23 to 32.66 MB (14 days), and 195.11 to 70.17 MB (month). Warm timings remain close for daily/fortnightly workloads and improve for the measured month. These are isolated workload measurements, not total server-RAM guarantees. Raw samples and a reproducible tool are included.

Production build, 55 focused profile/cache tests on each supported Node floor, and the full Node 22/MongoDB 6 backend suite pass (1,821 tests, one existing pending). All 511 local WebKit tests pass on Node 24. Every required hosted backend/browser/npm 12/CodeQL/native Docker check passed. Results and measurements cover corrected head afa8bfad. Merged into chore/nightscout-modernization after exact merge-tree verification. M25 remains open for the notification-cache review and complete validation. Also records the verified completion of M08/M19/M23/M26 through #8652. Targets chore/nightscout-modernization; #8605 stays draft into dev.

Browser-size trade-off: app JavaScript grows by 2,029 raw bytes / 187 Python gzip bytes; the other five bundles are byte-identical. This candidate prioritizes the measured profile-heap reduction over a small bundle-size increase.


AndyLow91:
Full backend validation passed on the current candidate: 1,819 tests with one existing pending, Node 22.23.2/MongoDB 6. The 58 focused profile/cache tests also pass on both supported Node floors. Local Chromium and hosted validation remain running; no merge recommendation yet.

PR #8658, opened 2026-09-06
perf: retain only acknowledgement fields in notification receipts

AndyLow91:
Cache only level, group and eventName for Pushover receipt acknowledgements. The current cache clones the whole notification, including unused messages and plugin graphs; the actual acknowledgement and snooze paths need only these three fields. Sending, cloning, expiry and cancellation behavior stay unchanged.

Seven focused cache cases pass on Node 22.23.2 and 24.20.0, including real high/low/plugin snooze selection, post-send mutation, repeat acknowledgements/expiry, failure suppression and cancellation retries. Six pass against the parent; the new retained-field regression fails there. Production build passes; the full backend suite passes 1,823 tests with one existing pending case. Hosted validation remains in progress.

Draft M25 slice targeting chore/nightscout-modernization. This does not remove node-cache or claim measured whole-server memory savings. Capacity/eviction and cache-maintenance decisions remain open. #8605 stays draft into dev.

### DIFF
--- a/docs/test-specs/notification-receipt-payload.md
+++ b/docs/test-specs/notification-receipt-payload.md
@@ -1,0 +1,24 @@
+# Notification receipt payload
+
+Pushover receipt acknowledgements need only level, group and eventName.
+The actual settings snooze selector reads eventName/level; notifications.ack
+receives level/group. Cache those fields instead of cloning the complete
+notification, including message data and plugin objects/functions.
+
+The send payload is unchanged. node-cache still clones the receipt snapshot
+and preserves its one-hour TTL, periodic cleanup, get/delete and cancellation
+behavior. This slice does not introduce eviction of outstanding receipts,
+change notification deduplication, or remove node-cache. Broader bounded-cache
+and maintained-library decisions remain open under M25.
+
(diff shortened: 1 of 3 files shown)


AndyLow91:
Refreshed against integration `87340f59`; new head `3e35b71b`. All seven notification cache regressions pass on both Node floors, including repeated cancellation, acknowledgement, expiry and snooze behavior. Fresh hosted checks are running. The payload reduction remains structural evidence only, not a measured whole-server memory saving.

AndyLow91:
#8658 merged as `42fd9a14` after all 12 backend jobs, eight replica-set jobs, six browser jobs, npm 12, CodeQL and both Docker checks passed on `f51fe34c`. Actual integration tree `c7b1300db2b9ee61b8ec991b908ae9797f526cdf` matches the reviewed merge tree. Receipt acknowledgement storage now retains only level/group/eventName, with repeated cancellation, snooze and expiry regressions. This is a structural payload reduction, not a measured whole-server memory saving; remaining M25 cache bounds/maintenance decisions stay open.

#8667 also passed its full local backend suite (1,852 passing, one existing pending); hosted CI is pending. #8605 remains draft into dev.

PR #8659, opened 2026-09-06
chore: maintain shared AJV 6 consumer compatibility

AndyLow91:
Update the shared AJV 6 override from 6.12.6 to 6.15.0 for ESLint/eslintrc, webpack lint schema-utils and request HAR validation. Preserve a single hoisted copy; AJV 8 consumers remain unchanged. Clean npm ci and npm ls pass with no added package paths.

Four actual-consumer tests pass on Node 22.23.2 and 24.20.0. Three also pass against the parent; malformed dynamic patterns throw on the parent but return a validation failure with the update. Production build passes; the full backend suite passes 1,825 tests with one existing pending case. Hosted validation remains in progress.

Security scope: the release provides optional regex-engine mitigation and catches malformed dynamic patterns, but native RegExp remains the default. The production HAR consumer does not enable $data; the relevant build consumers use local schemas. This is not a claim that arbitrary runtime patterns are safe. Detailed source/consumer analysis is included.

Draft M09 slice targeting chore/nightscout-modernization. #8605 remains draft into dev; other audit findings and parent maintenance stay open.


AndyLow91:
Refreshed against integration `87340f59`, including the query JavaScript and count pipeline guards. New head `b071bb90` preserves the scoped AJV 6 change. All four actual consumer regressions pass on both Node 22.23.2 and 24.20.0. Fresh hosted checks are running; prior green checks are not being substituted for this head's results.

PR #8664, opened 2026-09-06
fix: reject custom aggregation pipelines in the count API

AndyLow91:
The public count endpoint previously forwarded caller-supplied aggregation pipelines to MongoDB. That allowed stages to read collections outside the endpoint's selected storage. An owned fixture with three entry records and two private records returned count 6 through a supplied join pipeline.

Reject top-level `pipeline` parameters with HTTP 400 before storage aggregation. Documented `find` filters, existing storage selection/fallback and ordinary document fields named `pipeline` remain supported. Internal application aggregation is unchanged. The README documents the API restriction; this does not redesign authorization for the existing storage choices.

Four HTTP/database regressions have two passes and two failures before the fix, and all pass afterward on Node 22/MongoDB 6 and Node 24/MongoDB 8. The full Node 22/MongoDB 6 backend run passes **1,835 tests, one pending** before refreshing onto #8663. After that refresh, 24 combined count/query-boundary cases pass on Node 24/MongoDB 8. Clean install/production build and development/HTTP asset checks pass; fresh hosted validation remains required.

Draft targeting `chore/nightscout-modernization`. Separate from #8663's JavaScript-operator guard: declarative pipeline stages also need a public API boundary. No package, schema or browser-code changes. #8605 remains draft into dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,9 @@ You can get many more results, by using the `count`, `date`, `dateString`, and `
 
 Once you've installed Nightscout, you can access API documentation by loading `/api-docs/` URL in your instance.
 
+The `/api/v1/count/:storage/where` endpoint accepts `find` filters for counting records.
+Custom aggregation `pipeline` parameters are rejected with HTTP 400; they are not part of the public count API.
+
 #### Example Queries
 
 (replace `http://localhost:1337` with your own URL)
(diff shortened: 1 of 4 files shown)


AndyLow91:
Merged as `87340f59` after all required checks passed on `b7e063db`: 12 backend jobs, six browser jobs, npm 12, CodeQL and both native Docker checks. Verified actual integration tree `9433d3db936193b6fbff55a43f75dad4bd9bcd55` matches the reviewed merge tree. This closes the public count pipeline boundary only; the separate storage authorization concern remains open in modernization review.

PR #8671, opened 2026-09-06
docs: review legacy bridge dependencies and TLS migration requirements

AndyLow91:
Records the production lockfile audit at integration 0181cc99 and traces the four remaining moderate affected-package records through the retained legacy bridges. request uses only the private UUID v4 API without a supplied buffer; inspected bridge request sites use POST with default redirect policy. Owned transport probes now confirm that Dexcom POST requests do not follow 301/302/303/307/308 redirects in either protocol direction on both Node floors; the standalone MiniMed uploader likewise rejects a 302. These are scoped findings, not blanket advisory dismissals.

The review also identifies an independent unresolved issue: share2nightscout-bridge disables TLS certificate validation at five request sites, including authentication/login/read calls. Normal BRIDGE settings migrate to CONNECT, but forced-legacy and mixed-source cases remain reachable. The plan sets concrete transport tests, configuration mapping, compatibility and rollback gates before fixing or replacing those paths. Required deployment cases have been requested without credentials; no feature retirement is inferred.

Docs/audit evidence only. Installed versions were matched to the lock, source hashes are recorded, and npm audit output is included. No dependency, runtime or support-policy changes; M09/M29 remain open. Targets chore/nightscout-modernization; #8605 stays draft into dev.

Follow-up `13794a10` adds a reproducible loopback HTTP/TLS diagnostic and raw observations. Two cycles on both Node floors confirm actual legacy authentication/login and glucose calls accept an untrusted self-signed certificate, while the strict native control rejects it and accepts the explicitly trusted CA. Ephemeral keys/listeners are cleaned up and only fake credentials are used. This records a known defect, not a correctness oracle that a future secure fix must preserve. The transport fix/replacement and full bridge data-cycle validation remain open.

Merged into `chore/nightscout-modernization` as `3f6a0761`. All required existing head checks passed. The owned transport diagnostic was rerun against the current merge result on both Node floors, with matching observations. The actual resulting integration tree matches the independently calculated current-base merge tree. Existing full head CI and focused current-tree validation are distinct evidence; no claim of a new full matrix on every audit-only merge. Follow-up modernization gates remain tracked by #8605, which stays draft into dev.

### DIFF
--- a/docs/audits/legacy-bridge-source-hashes.json
+++ b/docs/audits/legacy-bridge-source-hashes.json
@@ -1,0 +1,13 @@
+{
+  "node_modules/request/lib/redirect.js": "4056494e80f0354a6253c6fc08a526bf10920c1b2bb517efb1249945a1291672",
+  "node_modules/request/lib/auth.js": "033318b8870b4a942c02fe04f8a9dce0930d9160399671f4c21806083681d994",
+  "node_modules/request/lib/oauth.js": "53fdc5f23d96f57db1e2a2152fc156949b37fccb5d3d9a39e035e919233798cb",
+  "node_modules/request/lib/multipart.js": "af2c0b45347049efcd475a47c15bd448fc714482d34b2caf7e3c912bef3ecf43",
+  "node_modules/share2nightscout-bridge/index.js": "0dc7198e24693c72a9a5e7284561d989a05a4b751739cadb3154eccbce8edf21",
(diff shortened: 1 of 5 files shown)


AndyLow91:
Refreshed this branch onto integration cbbd4581 (MongoDB driver 7) and pushed 35104e4a. A fresh Node 22 production-only lockfile audit still reports the same four moderate affected-package records and zero high/critical records. Both audit snapshots are retained; inspected app/dependency source hashes and installed versions still match. The existing transport evidence therefore concerns unchanged bridge sources. Fresh CI is running; the known legacy Dexcom certificate-validation defect and deployment/migration gates remain open.

PR #8679, opened 2026-09-06
Release notification caches during teardown

AndyLow91:
Notification teardown previously left two cache housekeeping timers and cached values alive. Late provider callbacks could also parse responses or repopulate those caches after shutdown.

Close and clear both caches on teardown, detach the notification listener, and ignore sends, acknowledgements and provider callbacks on the closed service. Active-service TTL, duplicate suppression and receipt/acknowledgement behavior remain unchanged.

Validation: all 12 combined notification/cache cases pass on Node 22.23.2 and 24.20.0. Both new two-cycle teardown regressions fail against the original implementation (retained timers and late response parsing). Local focused tests used matching dependencies from the native-tooltip validation worktree; clean-install hosted CI and current-base verification are required before merge. No real provider calls were made.

Includes the test specification and updates M25 as partial. Notification-cache bounds/clone policy and the final library retain/replace decision remain open. No dependency removal or typical-instance RSS reduction is claimed.

Targets chore/nightscout-modernization. Parent #8605 stays draft into dev.


AndyLow91:
Merged into chore/nightscout-modernization as f045877c98cf54c753acf6c93d133383645be08d after all 22 applicable checks passed. The actual merged tree e699c6043f934e3ea394abfdcc58e165d8339fd4 matches the reviewed current-base merge tree (base f50ec747, head 5b08b7a3).

Local Node 22/24 notification/cache checks pass (12 each), including regressions that fail on the original code. M25 remains partial: cache bounds/clone policy and the library retain/replace decision are still open. Parent #8605 remains draft and unmerged into dev.

PR #8729, opened 2026-09-12
fix: guard chart.update() against a 0-height #chartContainer measurement

wyomarus:
## What

`chart.update()` measures `#chartContainer` with `getBoundingClientRect()` and derives `focusHeight`/`contextHeight` from it. On the very first call, before the container's CSS layout has settled (e.g. a cold page load), that measurement can come back as `height: 0`. Without a floor, `chartHeight` goes negative, and so do the derived focus/context heights, which then get applied as invalid negative-height `<rect>` attributes (on the brush/overview strip).

Real browsers silently drop the invalid attribute rather than throwing, so this is normally invisible — a later `update()` call (already relied on today; e.g. the `update(false)` that immediately follows the initial `update(true)` in `lib/client/index.js`) re-measures correctly and overwrites it. But if that correcting call is ever delayed, the chart can render completely blank in the meantime — on the one screen most Nightscout users leave open specifically to watch their glucose.

## Why I noticed this

I run my own instance for my son. Opening the dashboard, I occasionally caught it rendering with a totally blank chart for a moment before it self-corrected. The browser console showed:
```
Error: <rect> attribute height: A negative value is not valid. ("-9")
```
every single time — reproducible on every load, not intermittent. Traced it to `chartContainerRect.height` measuring `0` on the first `update()` pass: `0 - PADDING_BOTTOM(30) = -30`, `-30 * .3 = -9`, matching the console output exactly.

## Fix

One-line floor at the point `chartHeight` is derived:
```js
var chartHeight = Math.max(chartContainerRect.height - PADDING_BOTTOM, 0);
```
This doesn't change the existing self-healing behavior (a subsequent correctly-measured `update()` call still recomputes and overwrites everything normally) — it just stops the *first*, degenerate pass from ever computing or applying negative geometry.

## Testing

Added `tests/client-core/chart-container-height-race.test.js`:
- Confirmed it **fails** against the unpatched code, reproducing the exact `-9` from the live repro.
- Passes with the fix.
- Added a third case confirming the existing self-healing path (0-height pass, then a correctly-measured pass) still ends up with the right final `focusHeight`/`contextHeight`.

Also ran the full suite to check for regressions:
- `npm run test:core` — 286 passing
- `tests/dependency-d3.test.js` — 24 passing

## Checklist

- [x] Targets `dev` per CONTRIBUTING.md
- [x] Root-cause fix (not a downstream null-check) per CONTRIBUTING.md's bug-fixing guidance
- [x] Regression test added
- [x] No new dependencies
- [x] Smoke-tested against a live instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/client/chart.js
+++ b/lib/client/chart.js
@@ -318,7 +318,14 @@ function init (client, d3, $) {
     var dataRange = client.dataExtent();
     var chartContainerRect = chartContainer[0].getBoundingClientRect();
     var chartWidth = chartContainerRect.width;
-    var chartHeight = chartContainerRect.height - PADDING_BOTTOM;
+    // getBoundingClientRect() can report 0 on the very first update() call,
+    // before the container's CSS layout has settled (e.g. a cold page load).
+    // Without the floor, that produces a negative chartHeight and therefore
+    // negative focus/context heights below, which get applied as invalid
+    // (and silently dropped) negative-height <rect> attributes -- harmless
+    // once a later update() call re-measures correctly, but if that next
+    // call is delayed the chart can render completely blank in the meantime.
+    var chartHeight = Math.max(chartContainerRect.height - PADDING_BOTTOM, 0);
 
     // get the height of each chart based on its container size ratio
     var focusHeight = chart.focusHeight = chartHeight * .7;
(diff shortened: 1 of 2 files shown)


AndyLow91:
a great find! thanks for your contribution!

PR #8732, opened 2026-09-13
fix(profile): pill and profile editor behaviour on sites with incomplete data

### LINKED ISSUE #7324: IAGE Pill changes color to yellow but not to red
**Describe the bug**
When reaching the time in minutes set in "iage_urgent" the pill on the main screen of Nightscout for IAGE is supposed to turn red. For other parameters like BAGE or CAGE it is working but not for IAGE. It remains in my case yellow. I have also set the "iage_warn" field an this works correctly and turns the pill yellow. The behavior is the same on all my devices (Windows Laptop, Android Phone) and different browsers.
I don't know if the persistent warning sound is affected as well as I don't use Nightscout alarms.

**To Reproduce**
Set the fields iage_warn and iage_urgent and monitor when the color changes.
For better comparison do the same with the fields "cage_warn" and "cage_urgent" or other parameters...

**Expected behavior**
I expect the pill for IAGE to turn red when the time in minutes set in "iage_urgent" is reached.

**Screenshots**
Minutes set for CAGE:
grafik

Minutes set for IAGE:
grafik

Wrong color for IAGE, correct color for CAGE on the Nightscout-Site with the above settings
grafik

**Your setup information**
- Nightscout 14.2.5 on ns.10be.de

**Additional context**
I contacted the Administrator of ns.10be.de and he was able to reproduce this behavior on his own setting.

alias-rahil:
Fixes #7324.

Most of what follows only appears on a site with incomplete data, which is the state a new install is in and where a lot of long-running installs stay.

The insulin age pill never turned red. `insulinage.js` compares the age against `insulinInfo.urgent`, which is assigned nowhere, so the comparison was always false and the pill stopped at warn while `batteryage`, `cannulaage` and `sensorage` compared against `prefs.urgent` correctly. Sensor age has a separate problem of the same family: it renders a bare `SAGE` label with no value on any site that has never recorded a sensor change. 604f0cba created that plugin carrying `shownAge = 'n/a '`, and 564ab830 dropped the line while moving cage/sage/iage onto the properties API, a change its own message describes as a clean up. `cannulaage` and `batteryage` still carry the identical fallback.

A site with no active profile could not be used at all. `addBasals` raises the profile dialog and assigns `window.location` from inside the draw path, so the alert reappeared on every render and could not be dismissed. It now fires once per page load. The dialog, its wording and the redirect are unchanged.

In the profile editor, an empty record set was replaced by a fabricated profile before the branch that decides which status to report, so a site with no profile at all reached "Values loaded." and the "Default values used." warning the editor already carries was unreachable. Letting the empty set through restores that warning, and the default values themselves are untouched. The fetch-error path is treated differently, because there a real profile may exist and simply have failed to load, and offering defaults invites overwriting it with values nobody chose. That path now leaves the clinical fields empty and refuses a save while any of them is unset.

Two smaller changes on the main screen. Pills that build no tooltip of their own, the trend arrow and the delta among them, carried no hover text at all, and now take a title from the label their plugin already declares, resolved the way `browser-settings.js` resolves the same field. Pills that already build a tooltip are left alone so that none ends up with two. Separately, `main.css` carried no `:hover` rule, so the pills, the range selector and the toolbar icons gave a pointer nothing to work with; they now use the hover background already defined for the drawer navigation.

Each commit is a single concern and can be dropped without disturbing the others. Every commit but the CSS one carries a test, and no existing test was modified. `npm run lint` reports the same set of errors before and after this branch; the `convertToRanges` unused-variable error in `profileeditor.js` reproduces on a clean `dev` checkout of that file and predates these changes.


AndyLow91:
The branch now incorporates current dev and the merge conflicts are resolved. The production build, 32 focused tests and 286 client-core tests passed locally. Two review findings remain:

1. The profile-editor AJAX stub invokes `.done()` after an error, unlike real jQuery. Consequently, the tests initialize fields and save handlers that would not be initialized after an actual failed fetch. Replacing the stub with a real jQuery Deferred reproduces two failures. The intended recovery should be explicit: keep editing and saving disabled after a failed load, retain the reload message, and only enable editing after a successful fetch. A successful empty result should still offer defaults, and incomplete values should still prevent saving.
2. A pill that initially has a native title retains that title when a later update supplies a custom tooltip. This can display both tooltips. The custom-tooltip path should remove the native title, with regression coverage for switching in both directions.

I will push a targeted follow-up implementing those behaviors and correcting the tests to use real Deferred success/failure semantics.

PR #8739, opened 2026-09-16
An urgent insulin-age alarm that could never fire, and two delivery defects (BF-28, BF-29, BF-31)

bewest:
# `bf/alarms` — an urgent insulin-age alarm that could never fire, and two delivery defects

Three commits on `origin/dev` `a8888f0d`, tip `5dcf783f`. 8 files, +400/−22, of which four are
test files. No `CHANGELOG.md` edit. Merges clean against `dev` and against every other open Phase 0
branch.

> **This one needs an explicit yes, not just a review.** It makes an alarm start firing that has
> never fired in any deployment, **and it removes a capability with no replacement** (per-request
> locale on the Alexa and Google Home endpoints). Both are the intended end state, but they are
> decisions rather than typo fixes, and an operator cannot undo either through configuration.

## What changes for you

**If you use the Insulin Age plugin with alerts turned on, you will start getting an URGENT
alarm you have never received before.**

Nightscout lets you set an age — by default 72 hours — after which your insulin reservoir is
"urgently" overdue for a change. Because of a one-word mistake in the code, that urgent alarm
has **never** been sent, in any release, to anyone. You may have set that threshold years ago
and never seen it honoured. After this change it works as documented.

Two things follow from that, and both are worth knowing before you upgrade:

- **Urgent severity and notification delivery are different.** The urgent indicator applies
  at and beyond the configured threshold. Notifications retain the existing window: with
  alerts enabled, they are requested only during the threshold hour through minute 20.
  At the default threshold, 72h can notify; 72h21m or 80h will show urgent severity without
  a notification. Starting or upgrading after that window does not issue a catch-up alarm.
- **The severity of what you already saw was also wrong.** Past the threshold the plugin was
  reporting the lower "warning" level for as long as the reservoir stayed overdue — not only
  during the hour the urgent notification would have fired. So this is not just a missing
  alarm; it is a level that was understated the whole time.

This alarm only reaches you if you have already turned Insulin Age alerts on
(`IAGE_ENABLE_ALERTS`). With alerts disabled, no notification is requested; the overdue
insulin-age indicator still changes from warning to urgent.

**Two smaller changes in the same branch:**

- **A plugin name in `ENABLE` that Nightscout does not recognise is no longer silently
  ignored.** If you wrote `insulinage` where Nightscout expects `iage`, the plugin simply did
  not load and nothing told you. You now get a suggestion in the server log naming the plugin
  it thinks you meant. **Your existing settings are not changed and nothing is renamed for
  you** — this only adds a message.
- **Voice assistant replies (Alexa, Google Home) use Nightscout's configured language,
  including relative-time phrases.** Request locale fields are ignored. The shared glucose
  and full-status handlers localize individual Moment instances, using the configured
  translation locale (including aliases such as `dk` → `da_DK` and `br` → `pt_BR`).
  They do not change process-global locale.

> **Read this if you use Alexa or Google Home with Nightscout.** Previously, a request's
> locale changed shared language metadata and Moment's global locale. It did not reload
> the translation catalogue, so this was not complete per-request translation support.
> It could change relative-time phrases for subsequent replies and other Moment consumers.
> Replies now consistently use the server's configured translation language for text and
> voice relative-time phrases. If you relied on request-specific date/time localization,
> that behavior is removed; request-scoped locale selection is not implemented here.

Nightscout is not a medical device and none of this is medical advice. If a change in when you
are alerted about insulin or site age affects how you manage your therapy, talk it through with
your care team.

---

## The commits

| commit | what |
|---|---|
| `8714093b` | `insulinage`'s urgent branch tested a field that is never set (**BF-28**) |
| `99e46a52` | an `ENABLE` entry naming the plugin's file instead of the plugin is silently ignored (**BF-29**) |
| `5dcf783f` | one assistant request re-languages the whole process (**BF-31**) |

Each lands alone; there is no ordering constraint inside the branch.

### BF-28 — the urgent branch is unreachable

```
lib/plugins/insulinage.js:92    if (insulinInfo.age >= insulinInfo.urgent) {
lib/plugins/insulinage.js:93      sendNotification = insulinInfo.age === prefs.urgent;
```

`urgent` is assigned on `prefs` (`:19`, `sbx.extendedSettings.urgent || 72`) and never on
`insulinInfo`. Line 92 evaluates `age >= undefined`, which is always `false`. Line 93, one line
below, reads `prefs.urgent` correctly. All three sibling age plugins — `cannulaage`,
`sensorage`, `batteryage` — use `prefs.urgent` on both lines. The fix is that one identifier.

Measured through the real sandbox at four ages: at 72 h the plugin requested nothing before, and
the URGENT overdue notification after. At 80 h, and at every age past the threshold, it reported
level **WARN** rather than URGENT — so the severity was wrong for the whole time the reservoir
stayed overdue, not only at the boundary.

**Notification window.** The fix changes the severity comparison to `prefs.urgent`; it does
not change the existing `age === prefs.urgent` or `minFractions <= 20` notification gates.
`IAGE_ENABLE_ALERTS` must still be enabled. The urgent indicator persists after the threshold,
but there is no catch-up notification outside the threshold-hour window. Regression tests cover
the minute-20 boundary, minute 21, 80h overdue, and alerts disabled at the threshold.

### BF-29 — an unrecognised `ENABLE` entry is silently ignored

Matching is against `plugin.name` (`bwp`, `cage`, `iage`, `sage`, `bage`, `basal` — six aliases),
not against the file name. `ENABLE=insulinage` loads nothing and says nothing. The branch adds a
suggestion based on edit distance.

**Known limitation, measured.** The suggester is silent on a stock settings string — `food`,
`bridge`, `delta`, `devicestatus` and `cors` produce no suggestion — while all six file-name
aliases and near-misses such as `pushove` do. It helps the case it was built for and does not
chatter on correct configuration.

### BF-31 — a voice request re-languages the whole process

`ctx.language.set(locale)` and `moment.locale()` are process-global, and `POST /api/v1/alexa` and
`POST /api/v1/googlehome` called them per request.

This is a **shared-state** defect. `language.set()` changes metadata but does not reload the
translation catalogue, so request locale did not translate the catalogue or alarm text.
`moment.locale()` did change formatting for subsequent Moment instances. The follow-up fix
uses instance-local locale in the shared voice glucose and full-status handlers, preserving
configured-language time phrases without changing shared state.

## Verifying it

```
TEST=insulinage       npm run test-single    #  5 passing   (no database)
TEST=plugins          npm run test-single    # 13 passing   (no database)
TEST=api.alexa        npm run test-single    #  4 passing   (needs MongoDB)
TEST=api.googlehome   npm run test-single    #  2 passing   (needs MongoDB)
```

Re-measured 2026-09-16 on `5dcf783f`, all four green.

Each is ablated, restoring the file it covers to `dev` and re-running:

| restore to `dev` | result |
|---|---|
| `lib/plugins/insulinage.js` | `TEST=insulinage` → 3 passing / **2 failing** |
| `lib/plugins/index.js` | `TEST=plugins` → 5 passing / **8 failing** |
| `lib/api/alexa/index.js` | `TEST=api.alexa` → 3 passing / **1 failing** |
| `lib/api/googlehome/index.js` | `TEST=api.googlehome` → 1 passing / **1 failing** |

The first two are database-free and give the same counts against a dead mongo port.

**`tests/api.alexa.test.js` and `tests/api.googlehome.test.js` are in neither `npm run test:unit`
nor `npm run test:integration`** — they are among the files only `npm run test-ci` reaches, which
is what `main.yml` runs. So a green `test:unit` is not evidence for BF-31, which is the commit that
makes this branch a major.

## Semver: major

The driver is not the alarm. It is the **removal of per-request locale handling from two HTTP
endpoints**: request locale previously affected shared language metadata and Moment formatting;
voice time phrases now use the server's configured language. The translation catalogue already
followed the server configuration. Request-scoped locale selection is not provided. The removal
is correct — the mechanism was process-global — but it is a capability removal on a declared HTTP
surface, and that is what grades it.

**If Phase 0 should land as a minor, this is one of exactly three changes that would have to be
split out.** Dropping `5dcf783f` leaves BF-28 and BF-29, which are a clean minor.

The "What changes for you" text above is the release-note source; this branch adds no
`CHANGELOG.md` entry. **Three paragraphs must not be dropped from it:** the
notification-window explanation, the note that the severity you already saw was understated, and the Alexa/Google Home
block — that last is the only part of this branch that takes something away from an operator who
was relying on it.

## Follow-ups deliberately not in this PR

- **The alexa `switch` has no `default`.** An unrecognised `request.type` calls neither
  `res.json()` nor `next()`, so the request hangs until the client times out. Low reachability,
  but it sits beside the `ctx.language.set(locale)` line this branch already changes, so **it
  would be cheaper to take it here than on its own** — flagged so the reviewer can ask for it now.
- **`plugins.isPluginEnabled` always returns `true`** — `find` returns `undefined`, compared
  against `!== null`. No caller today, so nothing observable.
- **`lib/authorization/storage.js:84` has an unguarded `console.log` on a request path**, printing
  request-derived values. Not introduced by this branch and not in a file it touches. It is
  repaired on the `bf/auth` branch, which is not yet open as a PR, so it is still live on `dev`.

## Follow-up verification

The targeted follow-up adds `tests/api.voice-locale.test.js`, which exercises both HTTP routes
and their real metric/status handlers without MongoDB. Twenty cases cover French, Danish,
Brazilian Portuguese, Traditional Chinese and English while requesting German, and assert
that global Moment locale, configured language, speech code and translations remain unchanged.
All twenty cases fail with the pre-fix voice implementation and pass with the fix.

Focused local validation: 58 passing across voice-locale, insulinage, plugins and language; changed
production-file lint and whitespace checks pass. The full CI matrix must rerun on the follow-up.


AndyLow91:
The insulin-age threshold fix is useful, and the current CI is green, but review found two issues to resolve before merging:

1. Removing per-request locale changes exposes a mixed-language response. With Nightscout configured in French, the real Google Home glucose-status handler returns `100 et stable à partir de 5 minutes ago.` instead of `100 et stable à partir de il y a 5 minutes.` The server does not initialize Moment's locale from its configured language. The fix should localize the Moment instances used for voice replies without modifying process-global locale. Regression tests should exercise actual metric/status responses in a non-English server language through both assistant routes and confirm that a request asking for another language cannot change shared state.

2. The description's claim that upgrading with an already-overdue reservoir triggers an urgent alarm is inaccurate. The existing notification window remains: at the default threshold, 72h triggers an urgent notification; 72h21m and 80h have urgent severity but no notification. This PR does not add catch-up notifications. The release-note source should distinguish urgent severity from notification delivery.

I will push a targeted follow-up commit covering localized voice time phrases and these regression cases, and correct the description. Per-request locale will remain ignored, as intended by this PR.

PR #8741, opened 2026-09-17
Fix Dexcom phone number accountNames and other numeric credential failures

### LINKED ISSUE #8056: Dexcom bridge does not work for Dexcom account names including '+'
I am not sure this is the correct place to file this report. The problem is with the share2nightscout-bridge. 

If the Dexcom user accounbt includes a '+' sign, the share2nightscout-bridge is not able to authenticate the user with the dexcom server. I get this in my Heroku log:

```
2023-07-14T21:17:26.611383+00:00 app[web.1]: Cannot authorize account:  null 500 {
2023-07-14T21:17:26.611407+00:00 app[web.1]:   Code: 'AccountPasswordInvalid',
2023-07-14T21:17:26.611407+00:00 app[web.1]:   Message: 'Publisher account password failed',
2023-07-14T21:17:26.611411+00:00 app[web.1]:   SubCode: '<OnlineException DateThrownLocal="2023-07-14T21:17:26.574" DateThrown="2023-07-14T21:17:26.574Z" ErrorCode="AccountPasswordInvalid" Type="13" Category="4" Severity="1" TypeString="InvalidPassword" CategoryString="Validation" SeverityString="Minor" HostName="" HostIP="" Id="{633224d6-8307-4c8d-9f81-020ac835ac08}" Message="Publisher account password failed" FullText="com.dexcom.udp.common.data.exception.OnlineException: Publisher account password failed" />',
```

I have tried multile Dexcom accounts that has '+' in them and that I know work (I can successfully login to Dexcom with them), all get the same error message. When I change to dexcom account that does not include a '+' it works. 

Again, sorry if this is the wrong place. I've tried to file the isse in the  share2nightscout-bridge repo but it seems no one is there to pick up.

### LINKED ISSUE #8181: Special character in the variable BRIDGE_USER_NAME
Hello, I have an email as the value in the variable special character in the variable BRIDGE_USER_NAME. If I put an ampersand in BRIDGE_USER_NAME, bridge does not work, that is, it does not read the dexcom account data. If I put a BRIDGE_USER_NAME without the ampersand it works perfect

vidia:
Nightscout doesn't support phone number based Dexcom accounts without the dependent account [workaround](https://nightscout.github.io/troubleshoot/dexcom_bridge/#username-is-a-phone-number). This is because phone number accounts get entered into configs as "+15551234567" which gets interpreted as a number, stripping the leading "+" (and potential leading zeroes), which breaks login.

## What's wrong

`findExtendedSettings()` in `lib/server/env.js` converts any `<PLUGIN>_<SETTING>`
value that looks numeric into a Number:

```js
if (!isNaN(value)) { value = Number(value); }
```

Which means Nightscout sends `{"accountName":15551234567}`, and Dexcom Share answers
`AccountPasswordInvalid`. The leading `+` is gone and the JSON type changed
from string to number.

This also applies to other env values like the account password (explaining the issues people have run into with [numeric passwords](https://nightscout.github.io/troubleshoot/dexcom_bridge/#username-and-password)), especially if variables have leading zeros since the type conversion will strip them.

`CONNECT_SHARE_PASSWORD=007700` becomes `7700`.

#8056 mentions this is just related to the old bridge approach, but this repros on the CONNECT_SHARE_* options too. 

## Related Conversations

**Credit for the diagnosis:** @BrianWieder identified this in a [comment](https://github.com/nightscout/cgm-remote-monitor/issues/8181#issuecomment-1849338353) on #8181.

## The change

I'm proposing an allowlist to let certain env variables skip this type conversion:

```diff
-            if (!isNaN(value)) { value = Number(value); }
+            if (!isNaN(value) && !isStringSetting(enable, ext)) { value = Number(value); }
```

`stringSettings` names the settings whose values are credentials or identifiers. Everything else is untouched — every other setting still converts to a number, and `on`/`true`/`off`/`false` still convert to booleans.

## Other approaches

In #8056 there wasa suggestion to resolve this in nightscout-connect, but this type conversion strips the `+` or any leding `0` before reaching that point, so nightscout-connect won't have enough information to reconstruct the values. Please let me know if I'm missing something here that makes it possible to handle in connect.

## Alternatives considered

- **Use the Dexcom Share Account ID instead of a username**: this was proposed in #8056, but would require a new config option and step for users to take to set up their account (even if its a simple one), and it looked like the existing account_id handling might need larger changes to handle it cleanly. Also it doesn't address the (admittedly less important, but still) issue with passwords and some other config options which should be treated as strings.
- **Detect damage generically** with a round-trip guard such as
  `String(Number(v)) === v`. It would break on values like `48.0` in fields which are meant to be numbers.

## Impact

- Dexcom accounts whose username is a phone number work without creating a
  second account. (And can remove the workarounds from documentation.)
- Passwords with leading zeros are sent as typed (can probably remove the numeric password warning.)
- If this is accepted, there should be a followup to clean up docs in `nightscout.github.io` (`docs/troubleshoot/dexcom_bridge.md`):
  *"Username is a phone number … This is not supported by Nightscout"*, and the
  advice to change an entirely numeric password.

## Related issues

- Fixes #8056
- Fixes #8181

This also happened with the old bridge connection too, but repros in the new Share version too: 
- Related: [share2nightscout-bridge#60](https://github.com/nightscout/share2nightscout-bridge/issues/60)
- Related: [share2nightscout-bridge#37](https://github.com/nightscout/share2nightscout-bridge/issues/37)

## Manual Validation

I tested with my Dexcom account (phone number based) using Docker:

| Configuration | Result |
|---|---|
| `dev` branch + phone-number account | Repros with `AccountPasswordInvalid`, 0 entries received from Dexcom |
| This branch + `CONNECT_SHARE_ACCOUNT_NAME` | Authenticated, live CGM entries stored |
| This branch + `BRIDGE_USER_NAME` (legacy path) | Authenticated, live CGM entries stored |

**I did not test the other configuration options I added to the allowlist since I don't have those accounts. I am happy to trim the list if that's better, but I included them since they are also credentials which would have similar issues (depending if numeric passwords or usernames are allowed on their services).**

## Tests

Unit tests added to `tests/env.test.js`:

- a phone-number Dexcom account name and a leading-zero password are kept as
  typed
- `BRIDGE_INTERVAL` still converts to a number, so the exception is per-setting
  rather than per-plugin
- `PROFILE_HISTORY=on` and `PUSHOVER_ALARM_KEY=off` still convert to booleans (an earlier version of this change skipped the boolean massaging too, but I opted to keep the test in for safety since it caught something for me while writing)

Suite: `tests/env.test.js` passes, as do `bridge`, `bridge-connect-compat`,
`pushover`, `maker`, `loop`, `mmconnect` and `debug-logging`. `eslint
lib/server/env.js` reports no new warnings.


AndyLow91:
An excellent solution to a long-running problem. thanks for the contribution and I'm happy to merge

